PR #269, opened 2013-08-15
Implemented logging in Tronstore

Codeacious:
This is an extension to Tronstore that implements logging. Now, there is a new [logging_tronstore] logger in the default logging and test logging configuration files, which is registered to a new handler that writes to a completely different file.

The actual logger is implemented by simply passing the tronstore logging object in the arguments to tronstore.main from trond. Python is actually really smart about recreating this object, so when Tronstore starts, it has a nice, configured logging object all ready to go.

I've made notes in the documentation about ensuring that the Tronstore logger writes to a different file, as Python cannot ensure logging concurrency between processes (threads are fine, however). This means that if the same file is used for Tronstore and trond, then it's very likely that log lines will be written at the same time.

In testing, since the critical tronstore logging (warns, errors) will only happen when trond is blocked on something, using the same file ended up working fine so long as the tronstore logging configuration was set to WARN or higher- however, turning on INFO level logging will make everything look pretty nasty.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -343,11 +343,17 @@ As of v0.3.3 Logging is no longer configured in the tron configuration file.
 
 Tron uses Python's standard logging and by default uses a rotating log file
 handler that rotates files each day. The default log directory is
-``/var/log/tron/tron.log``.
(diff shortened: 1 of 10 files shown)


solarkennedy:
I'm going to close this as won't fix for now. We aren't supporting backpors to older releases.

Issue #138, opened 2012-04-25
allow viewing by host in tronview


Roguelazer:
It would be great to be able to give tronview some flag to look at things by host instead of by job/service. Being able to see all of the services currently running on a host, as well as all of the jobs scheduled to run on it, would make administration a lot easier sometimes...

solarkennedy:
I'm going to close this as wont-fix, as we are moving tron away from host-specific schedules.

PR #271, opened 2013-08-16
Implemented a --hostname argument for tronview

Codeacious:
Pretty similar to what #251 did, but now with a hostname argument instead! As requested in #138, there is now a `--hostname` argument for tronview that shows all jobs and services that have that hostname in their `node_pool`.


solarkennedy:
I'm going to close this as wont-fix, as we are moving away from supporting the running of jobs on dedicated hostnames like this.

PR #294, opened 2014-08-20
Fix case where a machine crashes and won't restart.

bobtfish:
sudo /etc/init.d/tron start
- Starting tron  tron                                                                                                                                                                             Daemon was running as 20990. Remove PID file.

makes no sense - the computer knows the PID file is unused,
it should be able to remove it


solarkennedy:
I'm going to close this as won't fix, as we will (probably) be getting rid of pid-based locking and instead farming that out to systemd/upstart.

PR #301, opened 2015-01-15
Service monitor task should be requeued before notify

yyejun:
Current service monitor task run into a loop when node permanently down.
In this case notify code will try to cancel the monitor task, but the queue happens after the observer being notified.


dnephin:
Cool, lgtm

PR #304, opened 2015-04-02
Vagrant 'development playground' environment

mikepea:
This adds a multi-node Vagrant environment consisting of a tron 'master' node
and three 'batch-*' nodes for jobs/services.

The master node is fully configured, and ready for addition of the nodes and node_pool
entries via tronfig, which is documented in docs/developing.rst.

The tron master node is currenly Ubuntu 10.04 LTS, due to limitations in the debian package
build process. This also mirrors the most common deployment pattern for tron at the time of
writing, hopefully with a view to making this easier to migrate from.

### DIFF
--- a/docs/developing.rst
+++ b/docs/developing.rst
@@ -47,6 +47,57 @@ of Tron create a working directory and start
 
 A sample testing config file is available at ``tests/data/test_config.yaml``
 
+Running Tron under Vagrant
+--------------------------
+
+A Vagrantfile is present that will fire up a working multi-node Tron playground
+environment.
+
+Fire this up with::
+
+    $ vagrant up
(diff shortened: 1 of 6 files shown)


docs/developing.rst
```diff
@@ -47,6 +47,57 @@ of Tron create a working directory and start
...
+environment.
+
+Fire this up with::
+
+    $ vagrant up
+    $ vagrant ssh master.tron-v
+
+:command:`trond` will be running, with no nodes configured.
+
+A basic environment can then be loaded with :command:`tronfig`::
```
baris: It would be a good idea to put this config on the master to get everything ready.

baris:
Looks good to me, I like it!

PR #312, opened 2015-10-17
Introduce find_forgotten_procs.py

pkoch:
Sometimes Tron loses state and spins up new processes, even though the previous processes are still there and running.

I've been using this script to find those forgotten processes.

### DIFF
--- a/contrib/find_forgotten_procs.py
+++ b/contrib/find_forgotten_procs.py
@@ -200,13 +200,14 @@ def main(command, tron_base, *target_service_names, **options):
     )
 
     pprint(results)
-    print "Total instances", sum(
+    forgotten_count = sum(
         len(pids)
         for service_pid_pairs in results.itervalues()
         for _, pids in service_pid_pairs
     )
+    print "Total instances", forgotten_count
 
-    if command == 'kill':
+    if forgotten_count and command == 'kill':
         execute(
             parallel(kill_forgotten),
             results,
--- a/contrib/find_forgotten_procs.py
+++ b/contrib/find_forgotten_procs.py
@@ -1,0 +1,260 @@
+"""
+Find processes missing from Tron.
+"""
+import json
+import optparse
+import sys
+import urllib2
+from pprint import pprint
+
+try:
+    from fabric.version import VERSION, get_version
+    if VERSION < (1, 3):
+        raise Exception(
+            "This script needs fabric>=1.3 installed. You have {0}".format(
+                get_version()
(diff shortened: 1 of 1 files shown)


contrib/find_forgotten_procs.py
```diff
@@ -0,0 +1,279 @@
...
+                if inst['node']['hostname'] == hostname
+            ],
+        )
+        for service in services
+    ]
+
+
+def ssh_get_instances(host, service_to_target):
+    remote_output = _check_output(
+        ['ssh', host, 'bash'],
```
yyejun: Add an option to ssh as user@host

contrib/find_forgotten_procs.py
```diff
@@ -0,0 +1,279 @@
...
+    signal=DEFAULT_SIGNAL,
+    kill_prefix=DEFAULT_KILL_PREFIX,
+):
+    my_targets = sum((pids for _, pids in results[host]), [])
+
+    if not my_targets:
+        return
+
+    return _check_output(
+        ['ssh', '-A', host, 'bash'],
```
yyejun: Same as above need to ssh as user@host, and -A should be optional

contrib/find_forgotten_procs.py
```diff
@@ -0,0 +1,279 @@
...
+    if 'stdout' in kwargs:
+        raise ValueError('stdout argument not allowed, it will be overridden.')
+
+    if 'stdin' in kwargs:
+        curr = kwargs['stdin']
+        if isinstance(curr, basestring):
+            f = TemporaryFile()
+            f.write(curr)
+            f.seek(0)
+            kwargs['stdin'] = f
```
yyejun: Can use StringIO instead?

contrib/find_forgotten_procs.py
```diff
@@ -0,0 +1,279 @@
...
+        for node in service['node_pool']['nodes']
+    ])
+    if target_service_names:
+        target_services = [
+            service
+            for service in services
+            if any(fnmatch(service['name'], t) for t in target_service_names)
+        ]
+    else:
+        target_services = services
```
yyejun: It shouldn't default to all services. Add an option --all instead.

contrib/find_forgotten_procs.py
pkoch: Unfortunately, no. It requires an fd.

pkoch:
Added the three requested options.

PR #316, opened 2016-01-05
fix event memory leak

yyejun:
The cleanup method uses a different name during cleanup. So all jobrun events are not properly cleaned up.


patricklucas:
+1 :ship:

PR #317, opened 2016-01-06
extra cleanup after twisted ssh channel close

yyejun:
In twisted connection.py, deferreds[channel.id] is set to [] after channel close. Latest twisted use del self.deferreds[channel.id][:], but both will leak 1 int and 1 empty list per command execution. Memory will be release after connection close but in a real world usage ssh connection usually persist the entire duration of tron uptime.

All the code use self.deferreds expect non existent value, so it is safe to not leave an empty list behind.


patricklucas:
shipit :ship:

PR #318, opened 2016-08-03
Add 'hour' to the command context

solarkennedy:
Primary: @yyejun 

This adds an "hour" variable to the command context to make it easier for people to have hourly batches who don't want the granularity of seconds with unixtime.


yyejun:
lgtm

PR #319, opened 2016-08-11
only keep last buffer from ssh connection

yyejun:
Attempt reduce tron memory usage. Only the last ssh network data will be saved to disk.


mikepea:
LGTM 🚢

fede1024:
+1

PR #320, opened 2016-09-27
Use tox for running tests

solarkennedy:
Primary: @asottile ?

I removed testify globally, and this is one of the few straggling things that doesn't use virtualenv.

This is my attempt to make it.. work?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,8 +32,9 @@ setup(
         'pyasn1>=0.0.13',
         'pycrypto>=2.4',
         'pytz>=2011n',
-        'python-daemon>=1.5.2',
+        'python-daemon',
         'lockfile>=0.7',
+        'SQLAlchemy>=1.0.15',
     ],
     packages=find_packages(exclude=['tests.*','tests'])+['tronweb'],
     scripts=[
--- a/tests/commands/client_test.py
+++ b/tests/commands/client_test.py
@@ -120,7 +120,7 @@ class ClientTestCase(TestCase):
     def test_action_runs(self):
         self.client.action_runs('/api/jobs/name/0/act', num_lines=40)
         self.client.request.assert_called_with(
-            '/api/jobs/name/0/act?num_lines=40&include_stdout=1&include_stderr=1')
+            '/api/jobs/name/0/act?include_stderr=1&include_stdout=1&num_lines=40')
 
     def test_job_runs(self):
         self.client.job_runs('/api/jobs/name/0')
@@ -135,7 +135,7 @@ class ClientTestCase(TestCase):
     def test_jobs(self):
         self.client.jobs()
         self.client.request.assert_called_with(
-            '/api/jobs?include_job_runs=0&include_action_runs=0')
+            '/api/jobs?include_action_runs=0&include_job_runs=0')
 
 
 class GetUrlTestCase(TestCase):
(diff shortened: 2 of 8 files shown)


.travis.yaml
```diff
@@ -0,0 +1,4 @@
+---
```
asottile: traditionally this file is called `.travis.yml`

.travis.yaml
```diff
@@ -0,0 +1,4 @@
...
+languages: python
```
asottile: I believe the key is actually `language: python`

setup.py
```diff
@@ -32,7 +32,7 @@
         'pyasn1>=0.0.13',
```
asottile: I apparently can't comment on the rest of the file, but you can remove the distutils try: except: from the top.  New pip won't manage distutils installs and setuptools is pretty universal

tox.ini
```diff
@@ -0,0 +1,10 @@
+[tox]
+envlist = py27
+
+[testenv]
+deps =
+  Testify
+passenv = USER
+commands =
+    python setup.py install
```
asottile: shouldn't need this line, tox will automatically install your package for you

tox.ini
```diff
@@ -0,0 +1,10 @@
+[tox]
+envlist = py27
+
+[testenv]
+deps =
+  Testify
```
asottile: despite our github calling it Testify, the actual package name is [testify](https://github.com/Yelp/Testify/blob/ab5eba82cacd91a2774bec955852eda5b3128984/setup.py#L4), p4

asottile:
lol, too quick I guess.  You had already stared iterating :)

solarkennedy:
Ok tests pass and ready for review.

setup.py
solarkennedy: I.. would rather not, because this is still installed system-wide in the old way for prod.

tests/commands/client_test.py
```diff
@@ -120,12 +120,12 @@ def test_http_get(self):
     def test_action_runs(self):
         self.client.action_runs('/api/jobs/name/0/act', num_lines=40)
         self.client.request.assert_called_with(
-            '/api/jobs/name/0/act?include_stdout=1&num_lines=40&include_stderr=1')
+            '/api/jobs/name/0/act?num_lines=40&include_stdout=1&include_stderr=1')
```
asottile: Instead of pinning hash seed and depending on "random"ness, I'd rather these either:
- produce deterministic results
- compare hash-agnostically

asottile: Here's a diff that I believe accomplishes this (I also fixed the 5 space indent, wat)

```diff
diff --git a/tron/commands/client.py b/tron/commands/client.py
index 483bc2c..3bc2397 100644
--- a/tron/commands/client.py
+++ b/tron/commands/client.py
@@ -72,7 +72,7 @@ def request(uri, data=None):
 
 
 def build_get_url(url, data=None):
-     return '%s?%s' % (url, urllib.urlencode(data)) if data else url
+    return '%s?%s' % (url, urllib.urlencode(sorted(data.items()))) if data else url
 
 
 class Client(object):
```

solarkennedy: I knew you would say that :(

Its just tron and testify are just not worth it to me to spend more time fixing this.
I think the 30m I've spend adding travis and tox integration are "good enough" to get us past the testify purge.

solarkennedy: I'll try it out.

solarkennedy:
Sorry. I put the SEED back in because there are yet more non-deterministic tests:
https://travis-ci.org/Yelp/Tron/builds/163224474

I just don't believe it is worth either of our times to improve these.

asottile:
fine, fuckit :shipit:

PR #321, opened 2017-08-10
first pass at adding pre-commit

Rob-Johnson:
leaving the flake ignores in and I'll tackle them later.


Rob-Johnson:
@keymone raised a valid concern that this is spitting out python2.7 code which wont run on 2.6 (which Tron does). We should probably get Tron running on 2.7 before we do this

Rob-Johnson:
@solarkennedy @keymone what do you think of me merging now master is on python 2.7?

asottile:
looks like tests are failing, needs more `yaml.safe_dump(...)` somewhere

.pre-commit-config.yaml
```diff
@@ -0,0 +1,48 @@
...
+    -   id: end-of-file-fixer
+        language_version: python2.7
+    -   id: autopep8-wrapper
+        language_version: python2.7
+    -   id: check-docstring-first
+        language_version: python2.7
+    -   id: check-json
+        language_version: python2.7
+    -   id: check-yaml
+        files: ^(?!.*template).*\.(yaml|yml|eyaml)$
```
asottile: probably can remove this since this repository isn't paasta

.pre-commit-config.yaml
```diff
@@ -0,0 +1,48 @@
...
+    sha: v0.6.2
+    hooks:
+    -   id: add-trailing-comma
+-   repo: local
+    hooks:
+    -   id: no-catchall-except
+        name: Prevent catchall except
+        entry: except:$
+        language: pcre
+        files: \.py$
```
asottile: `types: [python]` is probably better

.pre-commit-config.yaml
```diff
@@ -0,0 +1,48 @@
...
+    -   id: flake8
+        exclude: ^docs/source/conf.py$
+        language_version: python2.7
+    -   id: check-ast
+        language_version: python2.7
+-   repo: https://github.com/asottile/reorder_python_imports.git
+    sha: v0.3.5
+    hooks:
+    -   id: reorder-python-imports
+        language_version: python2.7
```
asottile: you _should_ be able to drop the `language_version` stuff everywhere as long as you're using `pre-commit>0.15.1`: https://github.com/pre-commit/pre-commit/blob/master/CHANGELOG.md#0151

.pre-commit-config.yaml
```diff
@@ -0,0 +1,48 @@
...
+        args:
+        - --add-import
+        - from __future__ import unicode_literals
+        - --add-import
+        - from __future__ import absolute_import
```
asottile: can format this more readably:

```yaml
args: [
    '--add-import', 'from __future__ import absolute_import',
    '--add-import', 'from __future__ import unicode_literals',
]
```

as of pre-commit 0.16.3, these can be formatted nicely without fear of `pre-commit autoupdate` reformatting them: https://github.com/pre-commit/pre-commit/blob/master/CHANGELOG.md#0163

.pre-commit-config.yaml
keymone: `pre-commit --version` is 0.18.1, version guessed is 3.6 which fails flake8 and check-ast. how can i make it guess 2.7?

asottile: 0.18.1 is pretty old too.  It guesses based on the currently running interpreter, otherwise yeah  you should manually specify this.

If you're running `pre-commit` through `tox`, you can control the interpreter by settings `basepython = ...`

bin/action_runner.py
```diff
@@ -21,18 +25,19 @@
...
 class StatusFile(object):
     """Manage a status file."""
+
     def __init__(self, filename):
         self.filename = filename
 
     def write(self, command, proc):
         with opener(self.filename, 'w') as fh:
-            yaml.dump(self.get_content(command, proc), fh)
+            yaml.safe_dump(self.get_content(command, proc), fh)
```
solarkennedy: Yay safe!

asottile: fwiw this is going to use the slow pure-python dumper (very very slow)

The pattern I tend to use (or some variation thereof):

```python
import yaml

# could also *force* the c dumper by accessing it directly, though yaml can install without it if `libyaml-dev` isn't present
SafeDumper = getattr(yaml, 'CSafeDumper', yaml.SafeDumper)

yaml.dump(..., Dumper=SafeDumper)
```

keymone:
@asottile cool trick with ordered loader, might just use your code after this is shipped :)

docs/conf.py
```diff
@@ -10,12 +10,16 @@
...
+from __future__ import absolute_import
+from __future__ import unicode_literals
 
-import sys, os
+import os
+import sys
+
+import tron
 
 sys.path.insert(0, os.path.abspath('..'))
```
solarkennedy: This may have been above the "import tron" for a reason, for doc generation on pythonhosted.org?
Double check to make sure docs generation is not broken after this... I don't know how to do that off the top of my head

PR #324, opened 2017-08-16
[wip] use dh-virtualenv to package tron

Rob-Johnson:
Note: this also makes the python version in the virtualenv 2.7. Happy to do that in a separate PR if we want to?

I'm copy+pasting a lot of this from paasta, and am certainly no expert at debian packaging, so I'm open to feedback.

still to do: 
- [x] update the Make target for building the package
- [x] create symlinks in /usr/bin
- [x] make sure the init script for trond works

EDIT: seems the init script is broken anyway. My change hasn't broken it any more?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,13 @@ tronweb_tests/lib/
 vagrant/insecure_tron_key
 vagrant/insecure_tron_key.pub
 .tox
+
+# Generated debian artifacts
+debian/tron
+debian/files
+debian/tron.substvars
+debian/tron.debhelper.log
+debian/tron.postinst.debhelper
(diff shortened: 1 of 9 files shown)


Rob-Johnson:
ping @keymone - could you take a look?

solarkennedy:
Are you planning on installing dh-virtualenv in the Dockerfile so it can be built in a self-contained way? Otherwise I don't see where the dh-virtualenv package itself comes from.

The init script, I have no idea :(

Rob-Johnson:
fixed the test - we were asserting on the exception message that came back from parsing invalid json, which I guess changed in 2.7?

keymone:
🚢

PR #327, opened 2017-10-03
Issues 306 & 326 

Rob-Johnson:
/cc @asottile

### DIFF
--- a/tron/serialize/runstate/yamlstore.py
+++ b/tron/serialize/runstate/yamlstore.py
@@ -9,8 +9,9 @@ import itertools
 import operator
 import os
 from tron.serialize import runstate
+import yaml
 
-yaml = None # For pyflakes
+
 
 YamlKey = namedtuple('YamlKey', ['type', 'iden'])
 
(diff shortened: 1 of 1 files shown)


tron/serialize/runstate/yamlstore.py
```diff
@@ -9,8 +9,13 @@
 import operator
 import os
 from tron.serialize import runstate
+import yaml
+
+try:
+    from yaml.cyaml import CLoader as Loader
+except ImportError:  # pragma: no cover (no libyaml-dev / pypy)
+    Loader = yaml.Loader
```
asottile: not a regression, but perhaps a followup should change this to the `SafeLoader` (`CSafeLoader`) (for [more sanity](https://www.youtube.com/watch?v=ATY_R6qEa3s))

asottile: Here's [an example of that](https://github.com/pre-commit/pre-commit-hooks/blob/dec98f7bc860eb34889eab98bcc8eebd2a379b62/pre_commit_hooks/check_yaml.py#L8-L11)

asottile:
:boat:

PR #330, opened 2017-12-12
multi platform reactor; fix tests

### LINKED ISSUE #305: tron on OS X without Poll Reactor
Hi,

I'm getting an error 

```
Traceback (most recent call last):
  File "/usr/local/bin/trond", line 11, in <module>
    from tron import trondaemon
  File "/Library/Python/2.7/site-packages/tron/trondaemon.py", line 10, in <module>
    from twisted.internet import pollreactor, defer
  File "/Library/Python/2.7/site-packages/twisted/internet/pollreactor.py", line 18, in <module>
    from select import error as SelectError, poll
ImportError: cannot import name poll
```

From my understanding, poll is not supported on OS X (rather than using pollreactor I could be using cfreactor). Am I correct/is this the case?

If so, I imagine I can easily modify trondaemon.py to use cfreactor? 

Thanks,
Alex

Rob-Johnson:
fixes  #305 

switch to using epoll reactor rather than poll on linux; use select
reactor everwhere else (should work on darwin)

fix broken tests in shelvestore_test

### DIFF
--- a/tron/trondaemon.py
+++ b/tron/trondaemon.py
@@ -6,15 +6,22 @@ import logging.config
 import os
 import pkg_resources
 import daemon
+import platform
 
-from twisted.internet import pollreactor, defer
-from twisted.internet.main import installReactor
 
 import lockfile
 from tron.utils import flockfile
 import signal
 from twisted.python import log as twisted_log
 import tron
+
+if platform.system() == 'Linux':
+    from twisted.internet.epollreactor import EPollReactor as Reactor
+else:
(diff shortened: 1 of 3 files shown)


tron/trondaemon.py
```diff
@@ -173,8 +179,7 @@ def run(self):
 
 
     def setup_reactor(self):
-        self.reactor = pollreactor.PollReactor()
-        installReactor(self.reactor)
+        reactor.install()
```
solarkennedy: Not self.reactor?

PR #331, opened 2018-01-08
move from sysv-init to upstart

Rob-Johnson:
note that I'm forcing --nodaemon here, because I can't think of a reason for tron to fork and manage a pid file?


debian/tron.upstart
```diff
@@ -0,0 +1,14 @@
...
+stop on shutdown
+
+respawn
+
+script
+  set -a
+  if [ -f /etc/default/tron ] ; then
+      . /etc/default/tron
+  fi
+  exec start-stop-daemon --start -c $DAEMONUSER --exec /usr/bin/trond -- --nodaemon $DAEMON_OPTS
```
solarkennedy: We use modern versions of upstart, can you use 
http://upstart.ubuntu.com/cookbook/#setuid ?

maybe not due to the variables

solarkennedy:
The transition is tricky. You must init.d stop => replace init stuff => upstart start for it to work correctly.

Probably easiest to do manually with so few puppet servers we have, and not try to encode this in puppet.

debian/tron.upstart
Rob-Johnson: yeah, I couldn't figure out a way to squeeze DAEMONUSER into setuid.

PR #333, opened 2018-01-16
Handle configuration check requests

keymone:
Also add a flag to bin/tronfig to do a configuration check.


Rob-Johnson:
looks legit. shipit

bin/tronfig
```diff
@@ -32,18 +32,26 @@ def parse_options():
                       help="Print config to stdout, rather than uploading")
     parser.add_option("-n", "--no-header", action="store_true", default=False,
                       help="Print named config without a header")
+    parser.add_option("-C", "--check", action="store_true", dest="check",
```
solarkennedy: when I first read this, I was about to suggest that it never enter the upload_config function at all.

But in reality, this does require doing very down the path, entering the config controller, etc.

I can't think of a better name for this, "maybe" "dry-run", but it did surprise me that this command requires actually interacting with the API to function at all.

Especially considering that all the validation really does happen "statically" here:
https://github.com/Yelp/Tron/blob/b1d8e1a074c5ed489e8570fd967225b54f74cad5/tron/config/config_parse.py

I'm ok with shipping this as a first pass, and then as a second pass making the validation happen without going through the API.

PR #334, opened 2018-01-23
add a json schema for describing tronfig

Rob-Johnson:


tron/config/schema/tronfig_schema.json
```diff
@@ -0,0 +1,218 @@
+{
+    "$schema": "http://json-schema.org/draft-06/schema#",
+    "description": "",
+    "minProperties": 1,
+    "type": "object",
+    "additionalProperties": false,
+    "minProperties": 1,
+    "properties":{
+                "ssh_options": {
```
keymone: not a blocker, but what's the deal with indentation here?

tron/config/schema/tronfig_schema.json
```diff
@@ -0,0 +1,218 @@
+{
+    "$schema": "http://json-schema.org/draft-06/schema#",
+    "description": "",
+    "minProperties": 1,
+    "type": "object",
+    "additionalProperties": false,
+    "minProperties": 1,
```
keymone: `min_properties: 2` because we always need `jobs` and `services`?

tron/config/schema/tronfig_schema.json
```diff
@@ -0,0 +1,218 @@
+{
+    "$schema": "http://json-schema.org/draft-06/schema#",
+    "description": "",
+    "minProperties": 1,
```
keymone: duplicate `min_properties`

tron/config/schema/tronfig_schema.json
```diff
@@ -0,0 +1,218 @@
...
+                      "node": {
+                        "type": "string"
+                      },
+                      "schedule": {
+                        "oneOf" :  [
+                          {
+                            "type": "string"
+                          },
+                          {
+                            "type": "object"
```
keymone: attributes of schedule object missing?

tron/config/schema/tronfig_schema.json
```diff
@@ -0,0 +1,218 @@
...
+                            },
+                            "node": {
+                              "type": "string"
+                            }
+                          }
+                        }
+                      },
+                      "queuing": {
+                        "type": "boolean",
+                        "default": "true"
```
keymone: here and in other places: use json correct primitives for default values (actual boolean `false`, not string `"false"`)

tron/config/schema/tronfig_schema.json
Rob-Johnson: I think we always need more than that actually. let me figure out which ones are actually required and how to write the json schema for that.

tron/config/schema/tronfig_schema.json
Rob-Johnson: yeah, because they're variable and I don't know that you can really express it? I'll have another look.

tron/config/schema/tronfig_schema.json
Rob-Johnson: im not sure you can in jsonschema, but I'll double check

keymone: if that's the case, then does this line needs fixing: https://github.com/Yelp/Tron/pull/334/files#diff-a7ea5d2cfec07683e11d09dec94d44edR14 and the other places where there are number literals?

PR #335, opened 2018-01-26
Added prototype check_tron_jobs command

solarkennedy:
This command loops over jobs and looks at their last run status.

Eventually it will send alerts to teams?
Till then I'll dogfood it.

### DIFF
--- a/bin/check_tron_jobs
+++ b/bin/check_tron_jobs
@@ -1,0 +1,114 @@
+#!/usr/bin/env python
+import sys
+
+from tron.commands import cmd_utils
+from tron.commands import display
+from tron.commands.client import Client, get_object_type_from_identifier
+
+
+def parse_options():
+    usage  = ""
+    parser = cmd_utils.build_option_parser(usage)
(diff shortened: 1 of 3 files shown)


Rob-Johnson:
a fine start - the hard part is going to be figuring out owners without an explicit ``monitoring`` key being available in tronfig. Are you planning on blocking on that before releasing this?

Rob-Johnson:
its also nice to see that this is a bit simpler without having to try and merge history between 'temporary' rerun chronos and real jobs.

solarkennedy:
I'm not planning on blocking on `monitoring` config. I can do that in parallel and then put the pieces together.

Certainly the logic about alerting when "the last run failed" and "when a job is stuck" and all the edge cases about that are going to be issues to face as we implement this.

Also striking the balance between continuous checking (sensu) and "send 1 email when it fails" might be something we need to overcome. People are currently used to just getting 1 "notification" on failure, where this is a constant thing. We'll have to tune the realert_every setting just like in chronos.

PR #340, opened 2018-01-31
Local validation options, cleanup of unused functionality

keymone:
- add `-V` option to validate a single namespace file
- add `-D` option to do a full validation of a config directory
- comment out some validations for options that we intend to move out of master namespace into separate config file
- drop code related to editing namespaces directly

### DIFF
--- a/bin/tronfig
+++ b/bin/tronfig
@@ -105,17 +105,18 @@ def validate_dir(path):
 def get_config_input(namespace, source=None):
     if source is None:
         source = namespace
+        namespace = None
 
     if source == '-':
         source_io = sys.stdin
-        namespace = schema.MASTER_NAMESPACE
+        namespace = namespace or schema.MASTER_NAMESPACE
     else:
         source_io = open(source, 'r')
-        namespace = os.path.basename(source)[0:-5]
+        namespace = namespace or os.path.basename(source)[0:-5]
 
     content = source_io.read()
 
-    return namespace, content, source
+    return namespace, content
 
 
 if __name__ == '__main__':
@@ -125,7 +126,7 @@ if __name__ == '__main__':
 
     if options.validate or options.validate_dir:
         if options.validate:
-            name, content, source = get_config_input(*args)
+            name, content = get_config_input(*args)
             result = validate(name, content)
         elif options.validate_dir is not None:
             result = validate_dir(options.validate_dir)
--- a/bin/tronfig
+++ b/bin/tronfig
@@ -2,7 +2,9 @@
 from __future__ import with_statement
 import logging
 import os
+import shutil
 import tempfile
+import traceback
 import sys
 
 from tron.commands import cmd_utils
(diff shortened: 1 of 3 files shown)


bin/tronfig
```diff
@@ -58,33 +52,12 @@ def upload_config(client, config_name, contents, config_hash, check=False):
...
-            "Editor returned an error. Continue editing? (y/n): ")
-        if response[:1].lower() == 'n':
-            return
-
-    with open(fi.name) as upload_file:
-        return upload_file.read()
-
-
-def validate(config_name, config_content):
+def validate(config_name, config_content, source=None):
```
qui: Do you still need the source argument?

qui: Do you still need the source argument?

bin/tronfig
```diff
@@ -103,51 +76,79 @@ def delete_config(client, config_name):
...
+            print result
+            sys.exit(1)
+
+    client = Client(options.server)
     if options.print_config:
-        print_config(client, config_name, options.no_header)
+        print client.config(config_name)['config']
     elif options.delete:
         delete_config(client, config_name)
     elif options.from_stdin:
```
qui: This option doesn't exist anymore, remove?

bin/tronfig
```diff
@@ -103,51 +76,79 @@ def delete_config(client, config_name):
...
+            print "OK"
+            sys.exit(0)
+        else:
+            print result
+            sys.exit(1)
+
+    client = Client(options.server)
     if options.print_config:
-        print_config(client, config_name, options.no_header)
+        print client.config(config_name)['config']
```
qui: config_name is now undefined for both printing and deleting.

bin/tronfig
```diff
@@ -14,32 +16,24 @@ from tron.config import schema, manager, ConfigError
...
+    usage  = "usage: %prog [options] [<name>] [-|path]"
     parser = cmd_utils.build_option_parser(usage)
 
     parser.add_option("-p", "--print", action="store_true", dest="print_config",
                       help="Print config to stdout, rather than uploading")
-    parser.add_option("-n", "--no-header", action="store_true", default=False,
-                      help="Print named config without a header")
     parser.add_option("-C", "--check", action="store_true", dest="check",
-                      help="Only check configuration, don't apply")
+                      help="Upload and check configuration, don't apply")
```
qui: I think it would be helpful to say when you might want to upload, vs not.
Looks like uploading is good for checking one namespace against the current trond configuration, whereas -v just checks that namespace fragment?

bin/tronfig
```diff
@@ -103,51 +82,81 @@ def delete_config(client, config_name):
...
-        if validate(config_name, config_content):
-            if upload_config(client, config_name, config_content, config_hash):
-                return
+    content = source_io.read()
 
-        response = raw_input(
-            "There are errors in your configuration. Continue editing? (y/n): ")
-        if response[:1].lower() == 'n':
-            return
+    return namespace, content, source
```
qui: get_config_input still returns namespace, content, source.
There's one place below (validation) where all 3 are unpacked, but a couple others where it's assumed get_config_input just returns namespace, content.

Not sure which is correct, but should match up.

bin/tronfig
```diff
@@ -103,51 +82,82 @@ def delete_config(client, config_name):
...
-            return
-    raise SystemExit("tronfig failed")
+def validate_dir(path):
+    try:
+        manifest_dir = tempfile.mkdtemp()
+        manifest = manager.ManifestFile(manifest_dir)
+        manifest.create()
+        for fname in os.listdir(path):
+            if fname.endswith(".yaml"):
+                namespace = fname[0:-5]
```
Rob-Johnson: can you use 'os.path.splitext' here instead

bin/tronfig
```diff
@@ -103,51 +82,82 @@ def delete_config(client, config_name):
...
-        config_content = edit_config(config_content)
-        if not config_content:
-            log.warn("Cancelling edit.")
-            return
+    if source == '-':
+        source_io = sys.stdin
+        namespace = namespace or schema.MASTER_NAMESPACE
+    else:
+        source_io = open(source, 'r')
+        namespace = namespace or os.path.basename(source)[0:-5]
```
Rob-Johnson: splitext again

PR #343, opened 2018-02-05
Example cluster improvements

keymone:
(this is based off pre-commit branch)

rearrange some things, add console logger, integration tests stub


tron/serialize/runstate/shelvestore.py
```diff
@@ -42,7 +42,7 @@ def build_key(self, type, iden):
 
     def save(self, key_value_pairs):
         for key, state_data in key_value_pairs:
-            self.shelve[key.key] = state_data
+            self.shelve[str(key.key)] = state_data
```
solarkennedy: This could have ramifications on our production clusters, nod?
Are any of our keys non-strings?

keymone: i've checked - top level keys are always byte strings

i think we will get rid of shelve to store tron state anyway, but for now this should be safe

PR #344, opened 2018-02-06
Added dry-run and filtering on check_tron_jobs

solarkennedy:
Also starting to get this into shape with a test.
Also learning testify :(

### DIFF
--- a/bin/check_tron_jobs
+++ b/bin/check_tron_jobs
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 from __future__ import absolute_import
 from __future__ import unicode_literals
-
+import logging
 import sys
 
 from pysensu_yelp import send_event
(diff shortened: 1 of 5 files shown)


bin/check_tron_jobs
```diff
@@ -115,30 +124,38 @@ def compute_check_result_for_job(client, job):
...
     cmd_utils.load_config(options)
     client = Client(options.server)
+    log.debug("DEBUG")
 
     jobs = client.jobs(include_job_runs=True)
     for job in jobs:
-        check_job(job=job, client=client)
+        if options.job is None:
+               check_job(job=job, client=client, dry_run=options.dry_run)
+        elif job['name'] == options.job:
```
qui: Seems like it should be possible to check one job without looping through all of them...

From a quick look at the client code, maybe you can use client.get_url(job_name) and then get the job with client.job or client.job_runs?

bin/check_tron_jobs
```diff
@@ -115,30 +124,38 @@ def compute_check_result_for_job(client, job):
...
+            log.debug(result)
+            send_event(**result)
 
 
 def main():
     options, args = parse_options()
     cmd_utils.setup_logging(options)
     cmd_utils.load_config(options)
     client = Client(options.server)
+    log.debug("DEBUG")
```
qui: did you mean to leave this in?

debian/tron.links
```diff
@@ -1,4 +1,4 @@
-opt/venvs/tron/bin/check_tron_jobs usr/bin/check_tron_jobs
+opt/venvs/tron/bin/check_tron_jobs.py usr/bin/check_tron_jobs.py
```
Rob-Johnson: do we want this to have the .py extension in /usr/bin?

PR #345, opened 2018-02-08
Don't start tron on boot

keymone:
also tweak makefile and add preinst from dh_installinit


debian/tron.preinst.debhelper
```diff
@@ -0,0 +1,9 @@
+# Automatically added by dh_installinit
+if [ "$1" = install ] || [ "$1" = upgrade ]; then
+	if [ -e "/etc/init.d/tron" ] && [ -L "/etc/init.d/tron" ] \
+	   && [ $(readlink -f "/etc/init.d/tron") = /lib/init/upstart-job ]
+	then
+		rm -f "/etc/init.d/tron"
```
Rob-Johnson: we don't install /etc/init.d/tron anymore, only /etc/init/tron.

Makefile
```diff
@@ -1,62 +1,66 @@
...
-package_%_deb: clean  build_%_docker tronweb/js/cs
-	$(DOCKER_RUN) /bin/bash -c "dpkg-buildpackage -d && mv ../*.deb dist/"
+package_trusty_deb: clean build_trusty_docker coffee
+	$(DOCKER_RUN) /bin/bash -c "                   \
+		dpkg-buildpackage -d &&                      \
+		mv ../*.deb dist/$*/ &&                      \
+		chown -R $(UID):$(GID) dist/$*               \
+	"
 
 publish:
```
Rob-Johnson: want to bin this one too?

debian/control
```diff
@@ -8,7 +8,7 @@ Standards-Version: 3.8.3
 Package: tron
 Architecture: all
 Homepage: http://github.com/yelp/Tron
-Depends: python2.7, ${shlibs:Depends}, ${misc:Depends}
+Depends: python2.7, libyaml-dev, ${shlibs:Depends}, ${misc:Depends}
```
solarkennedy: why does it need libyaml-dev? This should only be a build dep

keymone: good point, i got confused what is installed when

PR #346, opened 2018-02-09
remove unicode_literal package since it causes unicode issue

chlgit:
The error message is as follows: https://fluffy.yelpcorp.com/i/kKCSLBKnW4gWT1K7fFvXXJgPN6hh7fW1.html

Or anyone has better idea about how to address this

### DIFF
--- a/tron/api/resource.py
+++ b/tron/api/resource.py
@@ -3,7 +3,6 @@ Web Services Interface used by command-line clients and web frontend to
 view current state, event history and send commands to trond.
 """
 from __future__ import absolute_import
-from __future__ import unicode_literals
 
 import datetime
 import logging


solarkennedy:
Obviously this is not a sustainable fix. I'm really worried we might have more cases of using unicode with twisted (we only exercised /api/status).

We can either sprinkle in `b'` where we need to, or give up on unicode_literals. Long term I think we need to add `'b`s. But how can we know where we need to add all of them?

qui:
@solarkennedy or this could be a concrete reason to move away from twisted?

qui:
Also found https://twistedmatrix.com/documents/current/core/howto/python3.html, which might help a bit with figuring out where bytes might be required?

solarkennedy:
@qui I don't think it is a super great reason to move to twisted. The errors are from our own introduction of unicode literals. 

And yes, I think that doc perfectly explains/enumerates where we need to be careful about when to use bytes and when to use strings, but in this case I think it is more subtle. 
https://github.com/Yelp/Tron/blob/1d3f19ae312dfc8c5fa29c786bb197ce8c040231/tron/api/resource.py#L438

It isn't obvious from that doc, but I think in this case when we say `'web'` with unicode literals, that becomes `u'web'`, and the object that eventually gets constructed can't handle that as part of a url. I don't think this is something we could determine statically or by following that guide by "grepping".

qui:
Yeah, those were just some initial thoughts I had before investigating the error itself. Thanks for adding context on the cause. The doc does mention that the twisted.web.resource.IResource class (which the linked code inherits from) needs bytes for URLs, but it's true that it wouldn't be easy to grep for, and there might be even more subtle cases elsewhere.

Scratch my first comment - my first impression was twisted couldn't deal with unicode at all, which doesn't actually make sense.

Anyway, we can drop unicode_literals for now, but we'll need to add the appropriate `b'`s for python3, since string literals will become unicode by default. Seems like you guys have this under control, but happy to brainstorm if you want any more ideas.

keymone:
this seems to be fixing it: https://github.com/Yelp/Tron/pull/347/files#diff-95e48de1392aabfd219bcb51c331d33aR46

solarkennedy:
Thanks @keymone, your solution looks much better.

PR #353, opened 2018-02-13
Python3 port 1/NaN

keymone:
merge #352, skip 3.6 travis/tox to keep master green


tron/commands/display.py
```diff
@@ -381,6 +385,9 @@ class DisplayJobs(TableDisplay):
     def format_value(self, field_idx, value):
         if self.fields[field_idx] == 'scheduler':
             value = display_scheduler(value)
+        elif self.fields[field_idx] == 'owner':
+            if isinstance(value, list):
```
qui: Is it possible for owner to be a list? 

https://github.com/Yelp/Tron/blob/07c0255e8a2c1a5653b5a58683c3849c3d69089a/tron/config/config_parse.py#L333 makes it seem like it should always be a string.

keymone: that was the test case, i didn't want to make this green by removing tests. will cleanup after merging.

PR #355, opened 2018-02-15
add stuck status for jobs

chlgit:
If a job is scheduled to run now, but there are still jobs running, couple of scenarios might happen:
    1. if allow_overlap is True, it runs anyway
    2. if queueing is True, it's marked as queued
    3. if queueing is False, it's marked as cancelled
Currently, we treat  (2) and (3) as job is stuck for some reason and send event to sensu. Let me know if you have other thought about the definition of stuck.

Make test works. And I also tested it through kwatest at devc and playground.

### DIFF
--- a/bin/check_tron_jobs.py
+++ b/bin/check_tron_jobs.py
@@ -55,7 +55,6 @@ def compute_check_result_for_job_runs(client, job, job_content):
         url_index, relevant_job_run['id'],
     )
     action_runs = client.job(job_run_id.url, include_action_runs=True)
-
     # A job action is like MASTER.foo.1.step1
     relevant_action = get_relevant_action(action_runs["runs"])
     action_run_id = get_object_type_from_identifier(
@@ -63,12 +62,12 @@ def compute_check_result_for_job_runs(client, job, job_content):
     )
     action_run_details = client.action_runs(action_run_id.url, num_lines=10)
 
-    if last_state == "succeeded":
-        prefix = "OK"
-        status = 0
-    if last_state == "queued" or last_state == "cancelled":
+    if is_job_stuck(job_content):
         prefix = "STUCK"
         status = 1
+    elif last_state == "succeeded":
+        prefix = "OK"
+        status = 0
     elif last_state == "failed":
         prefix = "CRIT"
         status = 2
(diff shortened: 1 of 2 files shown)


bin/check_tron_jobs.py
```diff
@@ -66,6 +66,9 @@ def compute_check_result_for_job_runs(client, job, job_content):
     if last_state == "succeeded":
         prefix = "OK"
         status = 0
+    if last_state == "queued" or last_state == "cancelled":
```
solarkennedy: This is not enough. last_state here refers to the last complete job. relevant_job_run considered the most recent active run to be the most relevant

bin/check_tron_jobs.py
```diff
@@ -111,6 +113,16 @@ def get_relevant_run(job_runs):
     return None
 
 
+def is_job_stuck(job_runs):
+    for run in job_runs['runs']:
+        run_state = run.get('state', 'unknown')
+        if run_state in ["running"]:
+            return False
+        if run_state in ["queued", "cancelled"]:
+            return True
```
solarkennedy: This isn't really a robust way to detect if a job is stuck.

What if a user simply runs 'tronctl stop' on a job and it becomes canceled? I would not want to email them telling them that it is stuck just because of that.

solarkennedy: Also `queued` is not enough to detect if a job is stuck or not if a user has allowed `allow_overlapping`

chlgit: Good point, I didn't take stopping job into account. 
But if a user has allowed allow_overlapping, user would expect one job run might longer than the scheduled interval, right? In this case, do we still concern the job being stuck?

solarkennedy: Yes, we still need to be concerned about a job being stuck.

"stuck" really is the wrong word. What users really want is: "Has it been too long since my job has completed successfully? (I don't care what the "state" is in the tron api)"

bin/check_tron_jobs.py
```diff
@@ -122,7 +137,7 @@ def compute_check_result_for_job(client, job):
     kwargs = {
         "name": "check_tron_job.{}".format(job['name']),
         "team": 'noop',
-        "notification_email": "kwa+checktron@yelp.com",
+        "notification_email": job['monitoring'].get('notification_email', 'unspecified'),
```
solarkennedy: whoops!

Somewhere, we need to be merging in the `monitoring` second in here directly and updating the kwargs dictionary. This way we get all the options (like team, runbook, alert_after, etc) instead of hand picking keys.

We can't do it this way, because we'll be sending a lot of emails to `unspecified@yelp.com`!

chlgit: I think it will only send email to unspecified :p
But yeah, hand picking keys is not a good idea.

PR #358, opened 2018-02-16
Document owner/notes/summary. Fixes #99

### LINKED ISSUE #99: Add 'owner' field to jobs
Sometimes the people who noticed job failures are not the people who maintain the jobs. One solution is to add an `owner` field to each job. Like so:

```
owner: require # one of {require, warn, ignore}
job:
  - name: my_sweet_job
    owner: sjohnson@yelp.com
```

```
tronctl owner my_sweet_job
> mailto:sjohnson@yelp.com # this can be command-clicked in iTerm2, which is nice
```

solarkennedy:
@mikepea added this feature in 2015, but we don't use it.

Do you think we actually do need it? I would rather we get rid of the code if we don't need it. (especially now that we have the "monitoring" section)

### DIFF
--- a/docs/jobs.rst
+++ b/docs/jobs.rst
@@ -78,6 +78,15 @@ Optional Fields
     Note: This requires an :ref:`action_runners` to be configured. If
     `action_runner` is none max_runtime does nothing.
 
+**owner** (default **""**)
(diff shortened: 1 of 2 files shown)


Rob-Johnson:
I say ditch the code. no one uses it.

PR #359, opened 2018-02-16
Ignore unknown ActionRuns

qui:
To fix the issue where Tron can't restore the state for some old job runs because the action dependencies have changed. Described in more detail in https://jira.yelpcorp.com/browse/TRON-210.

action_runs_for_actions is only used in _is_run_blocked, to check the state of required actions. If a "required" action isn't in the action map, that action didn't exist when that job ran, so it's not really required and we can ignore it.

The real problem is that Tron assumes the current action graph applies to all job runs, but that seems much harder to fix.

Plus one tiny change to the example cluster so it will clean up after you exit - realized I've just been leaving these containers around.

### DIFF
--- a/tests/core/actionrun_test.py
+++ b/tests/core/actionrun_test.py
@@ -687,6 +687,10 @@ class ActionRunCollectionIsRunBlockedTestCase(TestCase):
         self.run_map['action_name'].machine.state = ActionRun.STATE_FAILED
         assert self.collection._is_run_blocked(self.run_map['second_name'])
 
+    def test_is_run_blocked_required_actions_missing(self):
+        del self.run_map['action_name']
+        assert not self.collection._is_run_blocked(self.run_map['second_name'])
+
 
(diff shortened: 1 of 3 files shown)


tests/core/actionrun_test.py
```diff
@@ -687,6 +687,10 @@ def test_is_run_blocked_required_actions_failed(self):
         self.run_map['action_name'].machine.state = ActionRun.STATE_FAILED
         assert self.collection._is_run_blocked(self.run_map['second_name'])
 
+    def test_is_run_blocked_required_actions_missing(self):
+        del self.run_map['action_name']
```
Rob-Johnson: does this mutate self.run_map for the rest of the test class? might be worth cloning the run_map before mutating,

qui: No, that's the magic of testify! :) The @setup method that creates the run_map runs before every test case.

PR #360, opened 2018-02-17
Allow jobs to override the default timezone

solarkennedy:
US DST is coming up quickly on March 10.

It is embarrassing to me how many DST bugs we have every year because of tron. I would like to help eliminate some of them by allowing users to opt out on a per-job basis.

Internal ticket is TRON-117

### DIFF
--- a/docs/jobs.rst
+++ b/docs/jobs.rst
@@ -70,13 +70,18 @@ Optional Fields
     is only relevant when a Job is first added to the configuration, after
     which this value will be ignored.
 
-**max_runtime** (defailt **None**)
+**max_runtime** (default **None**)
     A time interval (ex: "2 hours") that limits the duration of each job run.
     If the job run is still running after this duration, all of it's actions
     are sent SIGTERM.
 
     Note: This requires an :ref:`action_runners` to be configured. If
     `action_runner` is none max_runtime does nothing.
+
+**time_zone** (default **None**)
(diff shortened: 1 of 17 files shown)


solarkennedy:
This still need some work.

solarkennedy:
This was way more work than I thought.
Also I think this has uncovered other situations where we are *not* using the `time_zone` configuration setting.

Also there is a high risk here for bugs related to jobs in the state that don't have a TZ, but the configs do. There is all sorts of crazy bugs that can pop up here when dealing with tz-aware and tz-naive datetimes.

solarkennedy:
I tested this manually using the UTC-based example playground with a US/Pacific `time_zone` setting and a US/Eastern job-specific tz setting and verified it was scheduled at the "right time". Still this change significantly increases our exposure to yet more tz-related bugs.

solarkennedy:
@qui re :177, I don't think that is the case. In this if block there is *no* tz specified, so it should run on the system time.

Tronview looks "reasonable" for some definition of reasonable. tronview shows actual localized scheduled time, tronweb shows local time per whatever your browser is set to (for better or worse)

qui:
@kwa I think there could be a timezone, there's just no start_time, which means that the job hasn't been scheduled before (there's no last_run_time).

solarkennedy:
Ah you are right.

PR #362, opened 2018-02-19
Python3 port

keymone:
Re-creating previous PR #352 on a cleaner branch.

This is no longer work in progress, all tests are green.


tests/core/serviceinstance_test.py
```diff
@@ -7,11 +7,9 @@
 from testify import setup
 from testify import setup_teardown
 from testify import TestCase
-from testify.assertions import assert_in
 from testify.assertions import assert_not_equal
 
 from tests.assertions import assert_length
-from tests.testingutils import autospec_method
```
qui: Since you're not getting rid of autospec_method, maybe we can bring it back in this file?

tron/serialize/runstate/yamlstore.py
```diff
@@ -41,8 +42,8 @@ def restore(self, keys):
             self.buffer = yaml.load(fh)
 
         items = (self.buffer.get(key.type, {}).get(key.iden) for key in keys)
-        key_item_pairs = itertools.izip(keys, items)
-        return dict(itertools.ifilter(operator.itemgetter(1), key_item_pairs))
+        key_item_pairs = zip(keys, items)
```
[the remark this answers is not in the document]
keymone: i thought so at first but then realize that these functions (zip, filter, map, etc) are implemented in python 2.7, which is what we're running now

tests/core/serviceinstance_test.py
keymone: we will gradually replace it with create_autospec. i'm genuinely afraid of that thing considering that googling for it returns this repository in the first search result: https://www.google.com/search?q=python+autospec_method&nfpr=1

contrib/find_forgotten_procs.py
```diff
@@ -210,7 +212,7 @@ def _get_services_from_tron(tron_base):
 
     url = tron_base + 'api/services'
 
-    return json.loads(urllib2.urlopen(url).read())['services']
+    return json.loads(request.urlopen(url).read())['services']
```
solarkennedy: I don't see this being imported, where did it come from?

setup.py
```diff
@@ -29,7 +30,8 @@
         "Development Status :: 4 - Beta",
     ],
     install_requires=[
-        'Twisted >=10.0.0, <=12.3',
+        'cryptography>=2.1.4',
```
solarkennedy: We were clamping twisted to <=12.3, now we are not clamping the upper bound, so we'll be jumping 6 years of twisted release. Any idea on how we will be able to tell if this works without just running it?

tron/command_context.py
```diff
@@ -199,3 +199,6 @@ def __mod__(self, _):
 
     def __nonzero__(self):
         return False
+
+    def __bool__(self):
```
solarkennedy: How did we get by without this before?

tron/utils/crontab.py
```diff
@@ -51,7 +51,15 @@ def parse(self, source):
             return None
 
         groups = [self.get_values(group) for group in self.get_groups(source)]
-        return sorted(set(itertools.chain.from_iterable(groups)))
+        groups = set(itertools.chain.from_iterable(groups))
+        has_last = False
+        if 'LAST' in groups:
```
solarkennedy: Why is this new logic here? Changing the python version shouldn't requires change like this, no?

tron/serialize/runstate/yamlstore.py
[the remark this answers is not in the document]
qui: Ah, I guess the only difference is that the python2.7 version returns a list, while the python3 version returns an iterable. But for both of the cases I commented on, that doesn't matter. Cool.

tron/utils/crontab.py
keymone: this is because python 3 no longer allows comparing numbers to strings, so i have to remove `'LAST'` from the array before passing it to `sorted()`

tron/command_context.py
keymone: i think python2 was using __nonzero__ on objects passed as `if` condition, python3 just does __bool__

setup.py
keymone: well we're able to pass integration tests on py2.7, which builds a deb, installs it, runs trond and sends some api requests - all of it works. for sure we will need to do a more thorough testing on playground / dev / stage.

contrib/find_forgotten_procs.py
keymone: imported on line 21 from six.moves.urllib

PR #363, opened 2018-02-20
store the actionrun id in the status output by action_runner

Rob-Johnson:
- changes action_runner to use argparse
- add the run_id as an arg to go in the status file
- just use pyyaml rather than the tron version, so that you don't need the tron package installed to use it. I'm open to reverting this, it mostly makes it easier to test (and the performance impact will be nbd for the action_runner)

here's the lines from the state file that the action_runner left behind

```
root@26a95ee742b5:/tmp/tron/MASTER.test.9085.first# cat status
{command: echo hi >> out && sleep 20, pid: 8540, return_code: null, run_id: MASTER.test.9085.first}
{command: echo hi >> out && sleep 20, pid: 8540, return_code: 0, run_id: MASTER.test.9085.first}
```

### DIFF
--- a/bin/action_runner.py
+++ b/bin/action_runner.py
@@ -5,22 +5,19 @@ Write pid and stdout/stderr to a standard location before execing a command.
 from __future__ import absolute_import
 from __future__ import unicode_literals
 
+import argparse
 import contextlib
 import logging
 import os
 import subprocess
 import sys
 
-from tron import yaml
-
+import yaml
 
 log = logging.getLogger("tron.action_runner")
 
 
 STATUS_FILE = 'status'
-
-
-opener = open
 
 
 class StatusFile(object):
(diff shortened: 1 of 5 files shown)


bin/action_runner.py
```diff
@@ -5,79 +5,96 @@
...
+    parser = argparse.ArgumentParser(description='Action Runner for Tron')
+    parser.add_argument(
+        'output_dir',
+        help='an integer for the accumulator',
+    )
+    parser.add_argument(
+        'command',
+        help='the command to run',
+    )
+    parser.add_argument(
```
solarkennedy: Is this immune to "quoting" issues? Before the command was the "last" argument, it it OK to have the ID as the last argument now, and is it safe to assume the "command" is properly contained as the "second argument".

The commands passed to tron are very complex and have very complex quoting. Here of all places we need to be super-careful that the optparse-argparse transition is safe.

bin/action_runner.py
```diff
@@ -5,79 +5,96 @@
...
     @contextlib.contextmanager
-    def wrap(self, command, proc):
-        self.write(command, proc)
-        try:
-            yield
-        finally:
-            self.write(command, proc)
-
-
-class NoFile(object):
```
solarkennedy: Did we not need this?

bin/action_runner.py
```diff
@@ -5,79 +5,96 @@
...
+import argparse
 import contextlib
 import logging
 import os
 import subprocess
 import sys
 
-from tron import yaml
-
+import yaml
```
solarkennedy: Are you sure you want to do this? Before we had guarantees that we were using the C extensions and stuff

solarkennedy:
Can you test this with a complex command stolen from prod to make sure quotes are preserved properly?

qui:
Just curious, what's the motivation for adding the id to the status file?
Also, a possible alternative - you could use the basepath of the status_path to get the job id? I guess it's not guaranteed to be the job id, but it is now.

bin/action_runner.py
Rob-Johnson: this was the 'silent' failure mode that was used when the os.makedirs call failed. I opted to get rid of it and fail hard, rather than degraded.

Rob-Johnson:
you're right @qui, I could use the directory. I think it's probably a less stable interface though? There isn't really a contract in place that says output_path *must* contain the job id.

bin/action_runner.py
Rob-Johnson: I don't think its the command parser that causes the problem, just that we don't quote correctly in actioncommand.py where we generate the args to action_runner (see TRON-213 for an example of it being broken right now).

I've fixed this by quoting everything in actioncommand.py and tested that the broken example in TRON-213 is fixed (as well as adding a test that splits the command generated, as argparse would).

bin/action_runner.py
Rob-Johnson: eh, this was kind of a convenience thing so that we don't have to have the whole tron package on the batch machine. I can put it back.

bin/action_runner.py
```diff
@@ -5,79 +5,96 @@
...
-opener = open
-
-
 class StatusFile(object):
     """Manage a status file."""
 
     def __init__(self, filename):
         self.filename = filename
 
-    def write(self, command, proc):
```
qui: Keeping this would make `wrap` a little cleaner, up to you.

tests/bin/action_runner_test.py
```diff
@@ -51,40 +40,56 @@ class RegisterTestCase(TestCase):
     def patch_sys(self):
         with contextlib.nested(
             mock.patch('action_runner.os.path.isdir', autospec=True),
-            mock.patch('action_runner.os.makedirs', autospec=True),
+            mock.patch('action_runner.os.access', autospec=True),
             mock.patch('action_runner.StatusFile', autospec=True),
         ) as (
             self.mock_isdir,
-            self.mock_makedirs,
```
qui: Might want to keep the mock of makedirs in this setup_teardown. That way, you don't have to remember to mock makedirs in every test that calls it, to avoid touching the filesystem.

tron/actioncommand.py
```diff
@@ -3,6 +3,7 @@
 
 import logging
 import os
+from pipes import quote
```
qui: didn't know about this! While looking this up, I saw that it's deprecated after python2.7.

You might want to import this with six.moves.shlex_quote to help maksym out :)

PR #364, opened 2018-02-22
Failure improvements

qui:
The example cluster changes will make it a lot less error-prone to test configuration from production.

The exception change will put the full error traceback in the upstart log, instead of the unhelpful "Failed to configure MCP" message. More details in internal ticket TRON-215.


Rob-Johnson:
gosh, its a real pain that we're not getting all the info we need from the syslog handler. maybe we should switch the logging (again) to log straight to /var/log/tron.log then use stdintoscribe to forward to scribe.

PR #369, opened 2018-02-27
restart package on upgrade

Rob-Johnson:
when we added the --noscripts option, we stopped dh_installinit autogenerating some of the install scripts - particularly we need the prerm script which stops tron before installing the new package.

these are already in .gitignore, so shouldn't bother us (other than being on the fs after the build of a deb).

also added a fix to the upstart script - we need the init-functions to call the log_failure_msg function


solarkennedy:
Are you sure this is a good idea? Would you rather have puppet restart the service when the install changes?

keymone:
unfortunately i don't recall anymore why i needed to override dh_installinit, but i would also rather have puppet stop/restart the service instead of doing it during upgrade. it is often useful to have new package installed while the old process is still running (config check before stopping the service, more control during manual upgrades, etc).

Rob-Johnson:
I agree that its useful to puppet to do upgrades it, but I think its surprising for this to not behave like a normal package. There are other oddities that come up like trond not stopping when you remove the package without these files. I'm happy to leave this if you both feel strongly, but I don't think we're doing this in the 'normal' way without this.

keymone:
well, if this is causing actual pain - shipit, i'm not going to block :)

solarkennedy:
Hmm. At least with puppet we have the *option* to not restart tron automatically on package upgrade. Ummmm..

solarkennedy:
We also want to install tron on places like devboxes to get tronfig tools and stuff (hooks). 

Will this change make tron auto-start on devboxes if we "just" install the package?

Rob-Johnson:
@solarkennedy yeah it will cause tron to try and start on devboxes, but we have RUN=no in /etc/default/tron templated in puppet, so Tron will refuse to start and upstart will give up after 5 attempts. This is another argument for packaging tron-client separately to trond.

I am going to remove the tron.default file from the debian folder though, so we don't have a race between puppet and the package to install the first defaults file.

Rob-Johnson:
I can't really think of a time that we'd want to install a new version of the package without doing the upgrade (and wanting to control via puppet)?

solarkennedy:
I think there are many situations where you want to install a package but you *don't* want it to automatically start or restart. redhat-based distros already don't do this, and I think you'll see debian-based distros start to do this more as the industry moves more towards containers.

I vote we continue to not take post install actions on package installs/upgrades and leave that up to the operator (puppet).

Rob-Johnson:
okay, I'll close.

PR #370, opened 2018-02-28
Python3 deb

keymone:
Just enough code to make the deb with py3 build and pass "integration tests".


.gitignore
```diff
@@ -17,6 +17,7 @@ tronweb_tests/lib/
 tron_state
 
 # Generated debian artifacts
+debian/debhelper-build-stamp
```
qui: This is already in here (towards the bottom).

tron/api/resource.py
```diff
@@ -406,25 +406,25 @@ def __init__(self, mcp):
 
         # Setup children
         self.putChild(
-            'jobs',
+            b'jobs',
```
qui: Do we really need this? We already have unicode_literals everywhere, so I would have expected that the current version wouldn't work if we needed bytes here.

keymone: yeah, i was surprised, but apparently everything URL-related is still handled as bytes in twisted

keymone:
@qui you didn't miss anything, i intended to merge this without forcing trusty package builds to be python 3 immediately, but i guess we can always revert if we have to ship more of py2 to prod.

keymone:
no longer WIP, though more testing is required on non-prod environments

tron/api/resource.py
```diff
@@ -36,23 +37,33 @@ def default(self, o):
...
+    request.setHeader(b'content-type', b'text/json; charset=utf-8')
     for key, val in six.iteritems((headers or {})):
         request.setHeader(str(key), str(val))
     return str(
         json.dumps(response_dict, cls=JSONEncoder) if response_dict else "",
-    )
+    ).encode('utf-8')
+
+
+def get_string_arg(request, name):
```
qui: Maybe the encoding/decoding logic should be added to requestargs.get_string? Seems like name.encode() could apply to all of the functions there, too.

tron/api/resource.py
```diff
@@ -218,7 +229,7 @@ def __init__(self, job_collection):
     def getChild(self, name, request):
         if not name:
             return self
-        return resource_from_collection(self.job_collection, name, JobResource)
+        return resource_from_collection(self.job_collection, name.decode(), JobResource)
```
qui: We should probably be decoding the name in all getChild calls before we use it, not just here. I'm going by the "use bytes at the boundaries, unicode everywhere else" rule. 

And since b'job_name' != 'job_name', I anticipate issues elsewhere. For example, there are some comparisons to '_event, and I think the dictionary keys from state are going to be unicode, not bytes. The unit tests don't really cover this.

.travis.yml
```diff
@@ -3,6 +3,7 @@ language: python
 env:
   - TENV=py27
   - TENV=py36
+  - TENV=itest
 python:
   - 2.7
```
solarkennedy: Can't we get rid of this part of the matrix?

If we are running 3.6 in the dh-virtualenv, I don't see any reason to test against 2.7 at all.

yelp_package/trusty/Dockerfile
```diff
@@ -1,20 +1,33 @@
...
-        wget \
-        gdebi-core \
+        dpkg-dev \
         gcc \
-        python-dev \
-        coffeescript \
+        gdebi-core \
+        git \
+        help2man \
+        libffi-dev \
```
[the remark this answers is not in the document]
keymone: fixed

example-cluster/images/xenial
```diff
@@ -0,0 +1,22 @@
+FROM ubuntu:xenial-20180123
```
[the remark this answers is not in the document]
keymone: iirc for a while ubuntu:xenial was a pretty old build. this also helps build reproducibility. we can get rid of it later if we want.

tron/api/resource.py
keymone: yeah, you're right, i'll update other resources

keymone:
i think this is ready for final review cc/ @Rob-Johnson @solarkennedy @qui

tron/core/actionrun.py
```diff
@@ -190,7 +190,7 @@ def __init__(
         self.start_time = start_time
         self.end_time = end_time
         self.exit_status = exit_status
-        self.bare_command = bare_command
+        self.bare_command = maybe_decode(bare_command)
         self.rendered_command = rendered_command
```
solarkennedy: why do we need it on bare_command but not on rendered_command?

keymone: iirc it's because we use bare_command as template for `%` operator and that can't be bytes

PR #372, opened 2018-03-01
Upgrade to argparse

jglukasik:
This needs to be done before we can add tab completion to Tron.

Sorry for the large review. Because of how each script used optparse on its own, in addition to sourcing "cmd_utils.py" which used it, all these had to be upgraded at the same time instead of one by one 😕

`tox` runs cleanly, as does the example cluster. I did some manual testing of the argument parsing by attaching to the trond master container in the example_cluster, and running the tron scripts in there. I also added a section to the example_cluster README explaining  how to do this.

### DIFF
--- a/bin/tronctl
+++ b/bin/tronctl
@@ -8,9 +8,9 @@ from __future__ import absolute_import
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import argparse
 import datetime
 import logging
-import optparse
 import sys
 import urlparse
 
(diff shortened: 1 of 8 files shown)


bin/tronfig
```diff
@@ -124,37 +124,31 @@ def validate_dir(path):
...
-    if source is None:
-        source = namespace
-        namespace = None
-
+def get_config_input(source):
     if source == '-':
         source_io = sys.stdin
-        if not namespace:
-            namespace = schema.MASTER_NAMESPACE
+        namespace = schema.MASTER_NAMESPACE
```
solarkennedy: It looks like we lost the ability to read a specific namespace (non MASTER from stdin)

We still need this functionality, please make sure there is a namespace argument (-p) sent in so we can still use our existing tools

bin/tronview
```diff
@@ -14,39 +14,43 @@ from tron.commands.client import TronObjectType
...
-    parser.add_option(
+    parser.add_argument(
         "--events", action="store_true", dest="show_events",
         help="Show events for the specified entity",
         default=False,
     )
+    parser.add_argument(
+        'name',
+        nargs='?',
+        help='job name',
```
solarkennedy: "job name" needs to be `[ job|run...` as before

bin/tronview
```diff
@@ -14,39 +14,43 @@ from tron.commands.client import TronObjectType
...
         help="Solely displays stderr", default=0,
     )
-    parser.add_option(
+    parser.add_argument(
         "--events", action="store_true", dest="show_events",
         help="Show events for the specified entity",
         default=False,
     )
+    parser.add_argument(
+        'name',
```
solarkennedy: Is the default still OK here and still work? I'm not sure what the default was (None?) but here it is it `[]`? whatever works as long as just running `tronview` still works with nothing extra

bin/tronview
```diff
@@ -113,35 +117,35 @@ obj_type_to_view_map = {
...
+    cmd_utils.load_config(args)
 
-    display.Color.toggle(options.display_color)
-    client = Client(options.server)
+    display.Color.toggle(args.display_color)
+    client = Client(args.server)
 
-    if not args:
-        output = view_all(options, client)
+    if not args.name:
```
solarkennedy: Oh, looks like it will work

example-cluster/README.md
```diff
@@ -15,3 +15,12 @@ $ tox -e example-cluster
...
 $ ./example-cluster/start.sh
 ```
+
+# To test tronview, tronctl, etc
+First, start Tron in a container using the above steps. Then in a different terminal,
+you will attach to that running container. There you will be able to run `tronctl`,
+`tronview` and others against the Trond master in the example cluster.
+
+```
+$ sudo docker exec -it `docker ps | grep examplecluster_master | cut -d' ' -f1` /bin/bash
```
solarkennedy: sudo .tox/example-cluster/bin/docker-compose -f example-cluster/docker-compose.yml run master

bin/tronview
qui: Yeah, this can either be a job name, job run id, or action run id.

tron/commands/cmd_utils.py
```diff
@@ -40,18 +40,26 @@ class ExitCode(object):
 opener = open
 
 
-def build_option_parser(usage, parser_class=optparse.OptionParser):
-    parser = parser_class(usage, version="%%prog %s" % tron.__version__)
+def build_option_parser(usage=None, parser_class=argparse.ArgumentParser, epilog=None):
```
Rob-Johnson: I can't see any reason to keep the parser_class argument around - just bin it and make everything use argparse.ArgumentParser.

example-cluster/README.md
jglukasik: Thanks for commenting on this, this line felt really hacky to me and I was hoping someone would give a better way.

But, this line doesn't seem to do what i want. This spins up another examplecluster_master container, instead of attaching to the existing one

solarkennedy: `sudo .tox/example-cluster/bin/docker-compose -f example-cluster/docker-compose.yml attach master` ?

I'm sure there is some docker-compose incantation that does what you want.

Not a blocker.

bin/tronctl
```diff
@@ -38,49 +38,54 @@ COMMAND_HELP = (
...
-        usage, parser_class=TronCtlOptionParser,
+        parser_class=argparse.ArgumentParser,
+        epilog=command_help_epilog(),
     )
 
-    parser.add_option(
-        "--run-date", action="callback", callback=parse_date,
-        type="string", dest="run_date",
+    parser.add_argument(
+        "--run-date",
```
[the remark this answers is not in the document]
jglukasik: Yup, looks like this works:
```
$ tronctl start MASTER.test --run-date 2018-03-28
Created JobRun:MASTER.test.61244
```

bin/tronctl
```diff
@@ -38,49 +37,53 @@ COMMAND_HELP = (
...
+        type=parse_date, dest="run_date",
         help="For job starts, what should run date be set to",
     )
+    parser.add_argument(
+        'command',
+        help='Tronctl command to run',
+        choices=[row[0] for row in COMMAND_HELP],
+    )
+    parser.add_argument(
+        'ids',
```
qui: Should this be 'id' now?

jglukasik:
I brought back the alternate namespace functionality. @solarkennedy I couldn't use `-p`, because thats already used to print, so i went with `-n`. I couldn't find any internal tools doing `tronfig $namespace $source` (that is, calling tronfig with two positional args). The only place i found this happening was in `itest.sh` in this repo, which i updated to use `-n $namespace`

I think my help string is a little vague: "Alternate namespace to use", but that might be my lack of experience with Tron.  Suggestions welcome

solarkennedy:
@jglukasik  can you rebase, and then do you feel comfortable changing puppet to sync with these changes when you push them?

PR #373, opened 2018-03-02
Only localize datetimes when they lack tzinfo

solarkennedy:
This fixes this stacktrace:
```
  File "/opt/venvs/tron/local/lib/python2.7/site-packages/twisted/internet/base.py", line 805, in runUntilCurrent
    call.func(*call.args, **call.kw)
  File "/opt/venvs/tron/local/lib/python2.7/site-packages/tron/core/job.py", line 338, in run_job
    self.schedule()
  File "/opt/venvs/tron/local/lib/python2.7/site-packages/tron/core/job.py", line 293, in schedule
    self.create_and_schedule_runs()
  File "/opt/venvs/tron/local/lib/python2.7/site-packages/tron/core/job.py", line 260, in create_and_schedule_runs
    for job_run in self.get_runs_to_schedule(ignore_last_run_time):
  File "/opt/venvs/tron/local/lib/python2.7/site-packages/tron/core/job.py", line 387, in get_runs_to_schedule
    next_run_time = self.job.scheduler.next_run_time(last_run_time)
  File "/opt/venvs/tron/local/lib/python2.7/site-packages/tron/scheduler.py", line 186, in next_run_time
    start_time = self.time_zone.localize(start_time, is_dst=None)
  File "/opt/venvs/tron/local/lib/python2.7/site-packages/pytz/__init__.py", line 226, in localize
    raise ValueError('Not naive datetime (tzinfo is already set)')
exceptions.ValueError: Not naive datetime (tzinfo is already set)
```

And I tried to cover other situations of adding and changing the timezone of jobs that have existing localized times. 

I'm worried though, because I think as written all of the special DST-specific logic is bypassed when the next run time has a tz set. Maybe that is a good thing?

### DIFF
--- a/tests/scheduler_test.py
+++ b/tests/scheduler_test.py
@@ -171,6 +171,7 @@ class GeneralSchedulerLongJobRunTest(GeneralSchedulerTimeTestBase):
 class GeneralSchedulerDSTTest(testingutils.MockTimeTestCase):
 
     now = datetime.datetime(2011, 11, 6, 1, 10, 0)
+    now_utc = timeutils.current_time(tz=pytz.timezone('UTC'))
 
     def hours_until_time(self, run_time, sch):
         tz = sch.time_zone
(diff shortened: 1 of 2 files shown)


solarkennedy:
Whoa, I didn't see the spec_match tz duplication logic. 

I thought about this overnight and I think this isn't "right" in the situation where you have the "start_time" (which is really "what time did it last run) which is tz-aware (which it will be if you specify timze_zone explicitly in a job and it runs at least once), then this DST logic won't fire, which means it will run twice on the fall-back scenario.

But if this is logic duplicate in the timespec code, then it might be "safe", but it also could easily mean we could get a similar stack trace later on.

We can't really afford tron to crash due to this new feature, so I'm thinking about adding try/except blocks here and logging an error if *anything* goes wrong. Prod tron crashes are catastrophic, and I don't want to have an incident because I couldn't predict all the interactions with this timezone math :(

PR #374, opened 2018-03-05
Chl/tron 224 detect jobs not scheduled

chlgit:
Sensu would alert if these no job run is scheduled or queued, and it won't alert if the only run is scheduled but not run yet (might be scheduled to run sometime in the future.)
Also, check if any kwargs of Sensu is missing before calling send_event.

### DIFF
--- a/bin/check_tron_jobs.py
+++ b/bin/check_tron_jobs.py
@@ -62,7 +62,7 @@ def compute_check_result_for_job_runs(client, job, job_content):
     )
     action_run_details = client.action_runs(action_run_id.url, num_lines=10)
 
-    if last_state == "succeeded" or last_state == "waiting_run_done":
+    if last_state == "succeeded" or last_state == "waiting_for_first_run":
         prefix = "OK"
         annotation = ""
         status = 0
@@ -127,7 +127,7 @@ def get_relevant_run_and_state(job_runs):
     for run in job_runs['runs']:
         if run.get('state', 'unknown') in ["failed", "succeeded"]:
             return run, run.get('state', 'unknown')
-    return job_runs['runs'][0], "waiting_run_done"
+    return job_runs['runs'][0], "waiting_for_first_run"
 
 
 def is_job_scheduled(job_runs):
--- a/bin/check_tron_jobs.py
+++ b/bin/check_tron_jobs.py
@@ -15,9 +15,6 @@ from tron.commands.client import get_object_type_from_identifier
 
 
 log = logging.getLogger('check_tron_jobs')
-SENSU_REQUIRED_KEYS = frozenset(
-    ['name', 'runbook', 'status', 'output', 'team'],
-)
 
 
 def parse_options():
@@ -77,7 +74,7 @@ def compute_check_result_for_job_runs(client, job, job_content):
         prefix = "CRIT"
         annotation = ""
         status = 2
-    elif last_state == "no_scheduled":
+    elif last_state == "not_scheduled":
         prefix = "CRIT"
         annotation = "Job is not scheduled at all"
         status = 2
@@ -123,7 +120,7 @@ def get_relevant_run_and_state(job_runs):
         return None, "no_run_yet"
     run = is_job_scheduled(job_runs)
     if run is None:
-        return job_runs['run'][0], "no_scheduled"
+        return job_runs['run'][0], "not_scheduled"
     run = is_job_stuck(job_runs)
     if run is not None:
         return run, "stuck"
@@ -206,9 +203,13 @@ def check_job_result(job, client, dry_run):
         log.info("Would have sent this event to sensu: ")
         log.info(result)
     else:
-        for key in SENSU_REQUIRED_KEYS:
-            if key not in result:
-                result[key] = 'unspecified'
+        log.debug("Sending event: {}".format(result))
+        if 'runbook' not in result:
+            log.info("No runbook specified")
+            result['runbook'] = 'foo'
+        if 'team' not in result:
+            log.info("No team specified")
+            result['team'] = 'noop'
         send_event(**result)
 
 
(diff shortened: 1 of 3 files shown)


bin/check_tron_jobs.py
```diff
@@ -112,12 +118,24 @@ def pretty_print_actions(action_run):
     return display.format_action_run_details(action_run)
 
 
-def get_relevant_run(job_runs):
+def get_relevant_run_and_state(job_runs):
+    if len(job_runs['runs']) == 0:
+        return None, "no_run_yet"
+    run = is_job_scheduled(job_runs)
+    if run is None:
+        return job_runs['run'][0], "no_scheduled"
```
solarkennedy: Can you call this "not_scheduled" ?

bin/check_tron_jobs.py
```diff
@@ -188,7 +206,9 @@ def check_job_result(job, client, dry_run):
         log.info("Would have sent this event to sensu: ")
         log.info(result)
     else:
-        log.debug("Sending event: {}".format(result))
+        for key in SENSU_REQUIRED_KEYS:
```
solarkennedy: Leave the debug in here. That is how we found it in the first place.

There is no need to iterate over these keys. The only keys we need to worry about being unset are runbook and team. Please replace those in the dictionary as needed.

For runbook please say something like "No runbook specified. Please add one with 'runbook: foo'"

bin/check_tron_jobs.py
```diff
@@ -63,18 +65,22 @@ def compute_check_result_for_job_runs(client, job, job_content):
     )
     action_run_details = client.action_runs(action_run_id.url, num_lines=10)
 
-    if last_state == "succeeded":
+    if last_state == "succeeded" or last_state == "waiting_run_done":
```
qui: Are you sure that you want to return 0 for waiting_run_done? That means if a job fails, the alert will resolve as soon as another run starts. Open to discussion, I'm not sure which is better.

tests/bin/check_tron_jobs_test.py
```diff
@@ -65,3 +65,29 @@ def test_is_job_stuck_when_runtime_not_sorted(self):
         }
         run = check_tron_jobs.is_job_stuck(job_runs)
         assert_equal(run['id'], 'MASTER.test.2')
+
+    def test_no_job_scheduled_or_queuing(self):
```
qui: Doesn't have to block this review, but I think long-term we should have tests for compute_check_result_for_job_runs with various lists of job_runs and what we expect the check result to be, instead of testing the helpers.

bin/check_tron_jobs.py
qui: On second thought, I think I misunderstood this state. What case is this state supposed to cover?

chlgit: I think it would only get "waiting_run_done" if (1) it's the first job run and it's scheduled but not run yet (runtime is set to future,) or (2) the first job run is running and the second job jun is scheduled. As long as the job has one or more run which state is "failed" or "succeeded", it shouldn't get "waiting_run_done" anymore.

tests/bin/check_tron_jobs_test.py
chlgit: Good idea, I will add more tests later.

bin/check_tron_jobs.py
chlgit: Sure.

bin/check_tron_jobs.py
qui: Okay, maybe we can call it "waiting_for_first_run" or something?

chlgit: Sure, that does sound better.

bin/check_tron_jobs.py
```diff
@@ -112,12 +115,24 @@ def pretty_print_actions(action_run):
...
+def get_relevant_run_and_state(job_runs):
+    if len(job_runs['runs']) == 0:
+        return None, "no_run_yet"
+    run = is_job_scheduled(job_runs)
+    if run is None:
+        return job_runs['run'][0], "not_scheduled"
     run = is_job_stuck(job_runs)
     if run is not None:
-        return run
+        return run, "stuck"
```
solarkennedy: It looks like it will be very easy to break the "setting" of these magic strings and the interpretation of these strings in the other function. Silly typos can easily break this code.

Can you use an ENUM to eliminate this type of bug?
https://docs.python.org/3/library/enum.html

bin/check_tron_jobs.py
```diff
@@ -189,6 +204,12 @@ def check_job_result(job, client, dry_run):
         log.info(result)
     else:
         log.debug("Sending event: {}".format(result))
+        if 'runbook' not in result:
+            log.info("No runbook specified")
+            result['runbook'] = 'foo'
```
solarkennedy: Say `No runbook specified. Please specify a runbook in the monitoring section of the job definition.`

No need to log

bin/check_tron_jobs.py
```diff
@@ -189,6 +204,12 @@ def check_job_result(job, client, dry_run):
...
+        if 'team' not in result:
```
solarkennedy: This needs to be short-circuited early. If we don't have a team we can quickly return.

bin/check_tron_jobs.py
```diff
@@ -6,17 +6,27 @@
...
 from tron.commands import display
 from tron.commands.client import Client
 from tron.commands.client import get_object_type_from_identifier
 
-
 log = logging.getLogger('check_tron_jobs')
 
 
+class State(Enum):
+    SUCCESSED = "succeeded"
```
solarkennedy: just do succeeded. `SUCCESSED` is not a word (but kinda looks like it should be)

bin/check_tron_jobs.py
```diff
@@ -176,6 +201,9 @@ def check_job(job, client):
             job['name'],
         ))
         return
+    if not job.get('monitoring').get('team', None):
```
solarkennedy: say "is None"

PR #379, opened 2018-03-08
replace enum by string until next release

chlgit:
If ship it, I will deploy check_tron_job.py to all tron master, and see if it can resolve page by itself.


solarkennedy:
Please use the enum34 method instead.

PR #380, opened 2018-03-08
First pass at tab complete

jglukasik:
This is a first iteration of tab complete. It parses the output of the /api/jobs endpoint of the default server.

We have {{args.server}} that will hold the tron server to use. It can optionally be specified with {{--server}}, or it will take the default of localhost. But, i cant use this variable, because at the time argcomplete is called, the {{args}} var has not been populated.

So instead, i added a helper function to give the default server as its hardcoded in tron/command/cmd_utils.py .

But is this default what we're using going forward? My understanding is eventually we want devs to be able to {{tronview}} from their devbox. Having the default as localhost wouldn't allow this.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,6 +29,7 @@ setup(
         "Development Status :: 4 - Beta",
     ],
     install_requires=[
+        'argcomplete >= 0.8.1',
         'humanize >= 0.5.0',
         'Twisted >=10.0.0, <=12.3',
         'PyYAML>=3.0',
@@ -38,6 +39,7 @@ setup(
         'pysensu-yelp',
         'python-daemon',
         'lockfile>=0.7',
+        'requests == 2.18.4',
         'six>=1.11.0',
         'SQLAlchemy>=1.0.15',
         'yelp-clog',
--- a/setup.py
+++ b/setup.py
@@ -39,7 +39,6 @@ setup(
         'pysensu-yelp',
         'python-daemon',
         'lockfile>=0.7',
-        'requests == 2.18.4',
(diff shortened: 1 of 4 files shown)


solarkennedy:
I think it is ok to start off with defaulting to whatever default server is available.

setup.py
```diff
@@ -38,6 +39,7 @@
         'pysensu-yelp',
         'python-daemon',
         'lockfile>=0.7',
+        'requests == 2.18.4',
```
qui: Don't need this anymore, right?

tron/commands/cmd_utils.py
```diff
@@ -40,6 +41,15 @@ class ExitCode(object):
 opener = open
 
 
+def get_default_server():
+    return DEFAULT_CONFIG['server']
+
+
+def tron_jobs_completer(prefix, parsed_args, **kwargs):
+    default_client = Client(get_default_server())
```
solarkennedy: How safe is this when it doesn't work? Do users get spammed with stacktraces as tab-complete suggestions?

solarkennedy:
@jglukasik on the next pass, can you break out a small script to pre-compute the tab completion options, and we'll cron that into a file. Then make the tab completer prefer the cached version and fall back to querying the api.

tron/commands/cmd_utils.py
jglukasik: Nope, you only see those stack traces if you have that "_ARC_DEBUG" env var set. Otherwise, nothing happens (you just get no suggestions on pressing tab)

PR #382, opened 2018-03-12
Allow tab complete to read from cached file

jglukasik:
If tab complete cache file does not exist, it will fallback to hitting api

Also adds the debian links file to automatically enable tab completion

I'm cargo-culting the _tabcomplete.sh file from what we have in paasta. I think this might enable zsh autocompletion too? But i haven't tested it.

This is paired with the internal review r:285778 , which has the puppet code that populates this cache file.


solarkennedy:
It won't enable zsh completion.

There will be a different cr to dump the jobs/actions right?

solarkennedy:
Don't worry about the failing test. Apparently it is DST dependent :( :( :(

PR #384, opened 2018-03-14
fix check_tron_job broken due to job misconfigure issue

chlgit:
use pysensu 0.3.5
catch exception if wrong configuration was using

### DIFF
--- a/bin/check_tron_jobs.py
+++ b/bin/check_tron_jobs.py
@@ -3,6 +3,7 @@ from __future__ import absolute_import
 from __future__ import unicode_literals
 
 import logging
+import sys
 import time
 
 from enum import Enum
@@ -237,8 +238,8 @@ def main():
         for job in jobs:
             try:
                 check_job_result(job=job, client=client, dry_run=args.dry_run)
-            except Exception:
-                log.info("check job result fails for job {}".format(job))
+            except Exception as e:
+                log.info("check job result fails for job {}: {}".format(job, e))
                 error_code = 1
     else:
         job_url = client.get_url(args.job)
@@ -249,4 +250,4 @@ def main():
 
 
 if __name__ == '__main__':
-    main()
+    sys.exit(main())
--- a/bin/check_tron_jobs.py
+++ b/bin/check_tron_jobs.py
@@ -231,14 +231,21 @@ def main():
     cmd_utils.load_config(args)
     client = Client(args.server)
 
+    error_code = 0
     if args.job is None:
         jobs = client.jobs(include_job_runs=True)
         for job in jobs:
-            check_job_result(job=job, client=client, dry_run=args.dry_run)
+            try:
+                check_job_result(job=job, client=client, dry_run=args.dry_run)
+            except Exception:
+                log.info("check job result fails for job {}".format(job))
+                error_code = 1
     else:
         job_url = client.get_url(args.job)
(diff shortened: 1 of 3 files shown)


qui:
So the script will exit 1 if any of the jobs fail, and alert us? Isn't that the same result as not catching the exception?

Not sure what the plan was, but I would expect it to alert the affected team, not us.

chlgit:
The main difference is if someone pushed a bad configuration, it would continue to check other jobs. Currently, it would only alert us. I think it's the same workflow like if someone pushed something bad to break paasta, it would notify ops (or us in the future) rather than the specific team.

bin/check_tron_jobs.py
```diff
@@ -231,15 +231,22 @@ def main():
...
     client = Client(args.server)
 
+    error_code = 0
     if args.job is None:
         jobs = client.jobs(include_job_runs=True)
         for job in jobs:
-            check_job_result(job=job, client=client, dry_run=args.dry_run)
+            try:
+                check_job_result(job=job, client=client, dry_run=args.dry_run)
+            except Exception:
```
solarkennedy: except Exception as e 

and then log the output of e so we know "why" it failed

bin/check_tron_jobs.py
```diff
@@ -231,15 +231,22 @@ def main():
...
     else:
         job_url = client.get_url(args.job)
         job = client.job_runs(job_url)
         check_job_result(job=job, client=client, dry_run=args.dry_run)
 
+    return error_code
+
 
 if __name__ == '__main__':
     main()
```
solarkennedy: sys.exit(main())

if you want to actually return the return code from the process

tests/bin/check_tron_jobs_test.py
```diff
@@ -4,13 +4,32 @@
...
 from check_tron_jobs import State
+from mock import PropertyMock
 from testify import assert_equal
 from testify import TestCase
 
 
 class CheckJobsTestCase(TestCase):
 
+    def test_check_job_result_exception(self):
+        with mock.patch('check_tron_jobs.parse_cli') as mock_args, \
```
solarkennedy: This isn't very pretty. Can you use the decorator method instead? I think it helps reduce the nestedness.

qui:
Okay, I see the benefit of continuing to check other jobs while alerting us to the failure. Just wanted to check who this change was supposed to alert.

tests/bin/check_tron_jobs_test.py
```diff
@@ -4,13 +4,32 @@
...
+                mock.patch('check_tron_jobs.cmd_utils'), \
+                mock.patch('check_tron_jobs.Client') as mock_client, \
+                mock.patch('check_tron_jobs.check_job_result') as mock_check_job_result:
+
+            type(mock_args.return_value).job = PropertyMock(return_value=None)
```
qui: These (lines 22-29) look like assignments, should they be assertions?

qui: anddd I completely misread that. Still confused why you need to set something on the type, though.

chlgit: According to the document here:  https://docs.python.org/3/library/unittest.mock.html#unittest.mock.PropertyMock

Since the way mock attributes are stored, a PropertyMock can't be attached to a mock object directly.

PR #385, opened 2018-03-15
Remove service code - first pass

chlgit:
I am tempting to remove service code part by part. As the first pass, I removed entire Tron/code/service.py, and made sure that "make test" still passes successfully. If this goes well, my next target is the part related to ServiceInstance.

Let me know if you have any concern about this, thx.


bin/tronctl
```diff
@@ -21,7 +21,7 @@ from tron.commands.cmd_utils import ExitCode
 
 
 COMMAND_HELP = (
-    ('start',           'Start the selected job, job run, action, or service'),
+    ('start',           'Start the selected job, job run, action'),
```
solarkennedy: Start the selected job, job run, or action

tron/mcp.py
```diff
@@ -156,12 +144,9 @@ def restore_state(self):
         to the configured Jobs and Services.
         """
         self.event_recorder.notice('restoring')
-        job_states, service_states = self.state_watcher.restore(
-            self.jobs.get_names(), self.services.get_names(),
-        )
+        job_states = self.state_watcher.restore(self.jobs.get_names())
```
solarkennedy: This is probably the this that worries me the most: can we survive the transition from services to no-services (with regard to state-loading)?

Probably.

bin/tronview
```diff
@@ -67,15 +67,11 @@ def display_events(data):
 
 
 def view_all(args, client):
-    """Retreive jobs and services and display them."""
+    """Retreive jobs and display them."""
```
qui: while we're fixing this, it's 'Retrieve'

qui:
Does tronweb still work, after removing /services from the API?

Also, config files still have 'services: []' in them, can Tron handle that when reconfiguring?

chlgit:
Good question and I don't know if it can survive tbh. I am thinking to make a release to test on playground. Any objection to doing that?

PR #387, opened 2018-03-20
Remove tron/utils/collections

keymone:
- remove backported Enum class
- remove MappingCollection
- also split up dev requirements from doc requirements


tron/core/job.py
```diff
@@ -435,32 +456,37 @@ class JobCollection(object):
     """A collection of jobs."""
 
     def __init__(self):
-        self.jobs = collections.MappingCollection('jobs')
+        self.jobs = {}
```
qui: There are other references to self.jobs in this class, do you need to update those too?

tox.ini
```diff
@@ -19,7 +19,7 @@ ignore = E501,E265,E241,E704
 
 [testenv:docs]
 deps =
-  --requirement={toxinidir}/dev/req_dev.txt
+  --requirement={toxinidir}/dev/req_docs.txt
```
solarkennedy: I'm surprised this works without req_dev.txt too. Can you confirm it actually does work as written?

keymone: can confirm, make docs passes provided services code is cleaned up

tron/core/job.py
keymone: i've checked - they all use dict-interface, so we're fine

PR #388, opened 2018-03-21
Dont try to localize time with tzinfo

keymone:
Fixes a crash when restoring a job with next run time non-existent in specified timezone.

### DIFF
--- a/tron/utils/trontimespec.py
+++ b/tron/utils/trontimespec.py
@@ -275,7 +275,8 @@ class TimeSpecification(object):
                 for _ in range(24):
                     out += datetime.timedelta(minutes=60)
                     try:
-                        out = self.timezone.localize(out)
+                        if out.tzinfo is None:
+                            out = self.timezone.localize(out)
                     except NonExistentTimeError:
                         return None
         return to_timezone(out, tzinfo)


qui:
https://github.com/Yelp/Tron/blob/c77967b7aa6659f490e45bab2c7798d3b7edaed1/tron/utils/trontimespec.py#L270-L279

This bug was mystifying me because it obviously makes it through L270 to get to L279, so in L270, out doesn't have a timezone...

We need to return `out` after 279 - that means there wasn't an error, so the time exists now and we should return. This bug was happening because we went through the loop again with the newly localized out.

tron/utils/trontimespec.py
```diff
@@ -275,7 +275,8 @@ def handle_timezone(self, out, tzinfo):
                 for _ in range(24):
                     out += datetime.timedelta(minutes=60)
                     try:
-                        out = self.timezone.localize(out)
+                        if out.tzinfo is None:
+                            out = self.timezone.localize(out)
```
qui: This whole method is a bit weird (do we ever need to add more than 1 hour?) but in theory, we should use self.timezone.localize(out, is_dst=None) to force it to check again if the time exists and raise NonExistentTimeError if it still doesn't.

keymone: well, i figured i'd rather add the minimal amount of code that implements the intended logic than change it and search for code paths that would break.

but i agree that adding 1 hour should always be enough, if you can give an advance shipit on such change - i'll merge that tomorrow.

keymone:
@qui yeah, it was really weird, but now i think it's fairly straightforward: on line 270 `out` doesn't have tz and is also non-existent in the local tz, so it jumps to the loop on 275, fixes itself in the first iteration and fails on the second (`.localize` on datetime with tz will raise `ValueError`)

qui:
Yeah, agreed about how it fails on the second iteration. I don't care if you solve it by returning out of the loop, or removing the loop entirely and just adding one hour. Either way, ship it!

PR #389, opened 2018-03-21
Implement and use backward compatible Py2Shelf

keymone:
Restoring from state file created by Python 2 fails in Python 3 due to reasons.

P2Shelf extends the base Shelf class, replicates DbfilenameShelf and passes around arguments required to make it work in py3.

manually tested:

- loading py2 state in py3 and verifying returned values using ShelveStateStore interface
- modifying state in py3, loading it in py2 and verifying the change

### DIFF
--- a/tron/serialize/runstate/shelvestore.py
+++ b/tron/serialize/runstate/shelvestore.py
@@ -17,14 +17,10 @@ log = logging.getLogger(__name__)
 
 
 class Py2Shelf(shelve.Shelf):
-    def __init__(self, filename, flag='c', protocol=2, writeback=False, keyencoding='ascii'):
-        self.keyencoding = keyencoding
-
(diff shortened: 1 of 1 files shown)


tron/serialize/runstate/shelvestore.py
```diff
@@ -38,7 +78,10 @@ class ShelveStateStore(object):
 
     def __init__(self, filename):
         self.filename = filename
-        self.shelve = shelve.open(self.filename)
+        self.shelve = Py2Shelf(
+            self.filename,
+            keyencoding='utf8',
```
solarkennedy: Are we ever going to need this?

solarkennedy: (the ability to adjust the keyencoding)

keymone: fair, dropped, thanks!

PR #390, opened 2018-03-21
Paasta executor configs

qui:
Master config: add clusters list
Jobs: add service, deploy_group
Actions: add executor, cluster, pool, cpus, mem, service, deploy_group

Validates that any action with executor: paasta either contains service and deploy_group, or its parent job does. Also validates that the cluster in actions is part of the list of clusters.

Next step will be using these values in the Job/ActionRun objects.

*Also includes a fix to the scheduler test that was failing.

### DIFF
--- a/tests/config/config_parse_test.py
+++ b/tests/config/config_parse_test.py
@@ -55,6 +55,10 @@ nodes:
 node_pools:
     - name: NodePool
       nodes: [node0, node1]
+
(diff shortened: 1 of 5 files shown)


qui:
Oops, rebasing on master. Scheduling test was already fixed.

solarkennedy:
A good start :)

PR #395, opened 2018-03-23
Use bsddb3 directly

keymone:
Apparently i didn't test Py2Shelf enough.

Python 2 through some db-detection logic in `anydbm` defaults to using bsddb module that was deprecated and removed from Python 3. It is recommended to use `bsddb3` module instead.

Since all our states are BDB, let's just use that directly and eventually replace it with another storage mechanism.

This adds some linux dependencies and makes life on OSX harder. To run tests we now need to `brew install berkeley-db` and export some env vars before running tox. I've included `osx-bdb.sh` script with relevant exports and made a section about it in README.md.

Also added some folders to .dockerignore because uploading half a gigabyte to docker daemon is not fun.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+.tox
+.git
+.idea
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ tron.egg-info
 *.swo
 docs/_build/
 .idea
+tron.iml
 docs/images/
 *.dot
 tronweb/js/cs/*.js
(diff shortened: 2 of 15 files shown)


Rob-Johnson:
lg2m - please add a test to cover the regression.

Rob-Johnson:
I kind of think we've hit breaking point and should create something more thorough to itest.sh (creating something akin to paasta's itests being the next step). But I won't block on that for this. shipit.

PR #397, opened 2018-03-23
Yelp's itest changes

keymone:
Run itest_* in docker container because package deps on jenkins hosts.


tox.ini
```diff
@@ -28,11 +28,6 @@ commands=
     dot -Tpng -odocs/images/action.png action.dot
     sphinx-build -b html -d docs/_build docs docs/_build/html
 
-[testenv:itest_deb]
```
qui: Remove from travis, then?

keymone: already done, travis is now using make target with the same name

PR #399, opened 2018-03-26
Chl/remove service code second pass

chlgit:
Remove the codes of Tron service configuration. It should be fine since we don't have any "services: []" in yaml file anymore.
Remove ServiceInstance related codes.
I have checked Tronfig, Tronweb, and Tronview wouldn't break.


tests/core/serviceinstance_test.py
```diff
@@ -7,16 +7,12 @@
 from testify import setup
```
solarkennedy: Can this whole file go away

tron/core/serviceinstance.py
```diff
@@ -2,17 +2,13 @@
 from __future__ import unicode_literals
 
 import logging
-import operator
```
solarkennedy: This too. Can this file be safely removed?

chlgit: Yeah, this is the target for next.

PR #401, opened 2018-03-27
First Stab at Action retries

keymone:
Action object gets new optional attribute: `retries=None`. ActionRun object gets two: `retries` and `retry_codes`. If `retries` is not `None`, any non-success event from action runner will decrease `retries`, append `exit_code` to `retry_codes`, reset state machine and initiate `start` transition.

When `retries` falls below zero, the usual `fail` transition logic will apply.

Termination methods like `stop` and `kill` will set `retries=-1`.


tron/config/schema.py
```diff
@@ -26,9 +26,12 @@ def config_object_factory(name, required=None, optional=None):
     """
     required = required or []
     optional = optional or []
+
     config_class = namedtuple(name, required + optional)
+    config_class.__new__.__defaults__ = (None,) * len(optional)
```
solarkennedy: Hmm. Is this to "DRY" out a lot of the defaults code stuff? The line itself looks magic to me. Can you add a... comment?

tron/core/actionrun.py
```diff
@@ -254,12 +274,18 @@ def start(self):
...
+            self.retries = -1
+
         stop_command = self.action_runner.build_stop_action_command(
             self.id, 'terminate',
         )
         self.node.submit_command(stop_command)
 
     def kill(self):
+        if self.retries is not None:
+            self.retries = -1
```
solarkennedy: I hope this is right. Seems like if an operator is giving this signal, they don't literally mean "kill this one try attempt", they probably actually mean "kill this attempt and cancel any retries"

tron/core/actionrun.py
```diff
@@ -165,6 +167,12 @@ def __init__(
         self.output_path = output_path or filehandler.OutputPath()
         self.output_path.append(self.id)
         self.context = command_context.build_context(self, parent_context)
+        self.retries = retries
+        self.retry_codes = retry_codes
```
solarkennedy: This `retry_codes` kinda confuses me. Can it just be called `exit_statuses` to mirror `exist_status`? And then just always be a list? (never None)

tron/core/actionrun.py
```diff
@@ -165,6 +167,12 @@ def __init__(
...
+        if self.retries is not None and self.retry_codes is None:
+            self.retry_codes = []
+
+        self.action_command = None
```
solarkennedy: Is this safe? How was this getting set before?

tron/core/actionrun.py
keymone: yeah, worst case they will have to also manually start the job again. it's still better than knowing you have to run kill command X times.

tron/config/schema.py
keymone: sure, i stole it from SO. it just makes last `len(optional)` arguments... optional :) kinda surprised namedtuple doesn't have a more convenient way to achieve this.

tron/core/actionrun.py
keymone: it is being set later on but it's a bad practice to initialize object properties outside __init__.

tron/core/actionrun.py
keymone: i didn't like how plural of status sounds, but you are right. will fix.

tron/core/actionrun.py
```diff
@@ -165,6 +167,12 @@ def __init__(
         self.output_path = output_path or filehandler.OutputPath()
         self.output_path.append(self.id)
         self.context = command_context.build_context(self, parent_context)
+        self.retries = retries
```
Rob-Johnson: this is really minor, but it took me reading the code to figure out whether this variable meant 'number of retries already attempted' or 'number of retries remaining'. could you instead rename this 'remaining_attempts', or seperate into ``self.current_attempt`` and ``self.max_attempts``?

tron/core/actionrun.py
```diff
@@ -307,14 +333,20 @@ def _done(self, target, exit_status=0):
             return self.machine.transition(target)
 
     def fail(self, exit_status=0):
+        if self.retries is not None and not self.retries < 0:
+            self.retries -= 1
+            self.retry_codes.append(exit_status)
+            self.machine.reset()
+            return self.start()
```
Rob-Johnson: should we add a backoff between attempts here?

keymone:
i've tested to primary functionality and it seems to work as expected. are there any particular test cases we'd like to cover? cc/ @Rob-Johnson @qui @solarkennedy

tron/core/actionrun.py
```diff
@@ -184,7 +192,7 @@ def check_state(self, state):
     @classmethod
     def from_state(
         cls, state_data, parent_context, output_path,
-        job_run_node, cleanup=False,
+        job_run_node, cleanup=False, retries_remaining=None, exit_statuses=None,
```
qui: Why not use retries_remaining and exit_statuses from the state_data, so you don't have to pass it separately?

Otherwise, you'll need to edit ActionRunFactory.action_run_from_state.

tests/core/actionrun_test.py
```diff
@@ -335,6 +339,23 @@ def test_is_broken(self):
         self.action_run.machine.state = ActionRun.STATE_QUEUED
         assert not self.action_run.is_broken
 
+    def test_retry(self):
+        self.action_run.retries_remaining = 2
```
qui: There are two retries allowed, so there should be 3 failures before the action_run becomes failed, right? How does this test work? (Why are there only two EXITING updates?)

The new math in actionrun.py does look right to me, though.

keymone: good point! retry logic actually had a subtle bug. i've moved it all to `.fail` method and fixed it.

PR #403, opened 2018-03-27
Timezone date context bug

qui:
macro_timedelta was failing if start date includes tzinfo. This was causing errors in DateArithmetic.
Made everything in timeutils handle timezones, and added tests.

Also changed manual_start to include tzinfo in the run_time if the job has it.

### DIFF
--- a/tron/utils/timeutils.py
+++ b/tron/utils/timeutils.py
@@ -20,7 +20,7 @@ def current_timestamp():
 
 def to_timestamp(time_val):
     """Generate a unix timestamp for the given datetime instance"""
-    return time.mktime(time_val.timetuple())
+    return time.mktime(time_val.utctimetuple())
 
 
 def delta_total_seconds(td):
@@ -47,7 +47,8 @@ def macro_timedelta(start_date, years=0, months=0, days=0, hours=0):
     end_date = datetime.datetime(
         start_date.year + years, new_month, start_date.day, start_date.hour,
     )
-    delta += end_date - start_date
+    month_and_year_delta = end_date - start_date.replace(tzinfo=None)
+    delta += month_and_year_delta
 
     return delta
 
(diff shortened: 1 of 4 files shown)


tron/utils/timeutils.py
```diff
@@ -59,6 +62,8 @@ def duration(start_time, end_time=None):
     if not start_time:
         return None
     last_time = end_time if end_time else current_time()
+    if last_time.tzinfo is None and start_time.tzinfo is not None:
```
solarkennedy: What about the other cases here, where a user has added a tz or removed one recently?
Won't this results in another naive/aware clash?

tron/utils/timeutils.py
```diff
@@ -20,6 +20,8 @@ def current_timestamp():
 
 def to_timestamp(time_val):
     """Generate a unix timestamp for the given datetime instance"""
+    if time_val.tzinfo is not None:
+        return time_val.tzinfo.utcoffset(time_val).total_seconds()
```
solarkennedy: I don't think this is correct. This returns literally "the number of seconds difference between the tzs", not the absolute unit timestamp.

Maybe it is

    time.mktime(time_val.utctimetuple())

?

tron/utils/timeutils.py
qui: Yes, you're right. I'm not sure how to handle this case in general if one has a tzinfo and the other doesn't. I was mostly thinking about the case where start does, and end_time=None so we get a naive time from current_time().

Now I realize that this wouldn't give you the right answer if start_time.tzinfo doesn't match the local timezone.

Good news is that we don't actually need this. duration is only used on naive datetimes now.
start time is set here:
https://github.com/Yelp/Tron/blob/4c7e637058f5e371053878fa7a1732cde6732dbb/tron/core/actionrun.py#L234
and end_time here: 
https://github.com/Yelp/Tron/blob/4c7e637058f5e371053878fa7a1732cde6732dbb/tron/core/actionrun.py#L305

What if I just take out this "fix" and add a warning that you shouldn't mix timezones?

tron/utils/timeutils.py
qui: Yes, this works! Thanks. Fixed and added another test that catches this.

tron/utils/timeutils.py
solarkennedy: A warning sounds good

PR #405, opened 2018-03-29
remove service code third pass

chlgit:
At this point, I think there won't be any services related codes. But feel free to point out if I miss anything.


qui:
🎉 almost there!
found a few more with grep:
https://github.com/Yelp/Tron/blob/master/tron/serialize/runstate/__init__.py
https://github.com/Yelp/Tron/blob/master/contrib/find_forgotten_procs.py
docs/
tronweb_tests/SpecRunner.html

I don't mind if you remove them now, or in a separate review.

PR #407, opened 2018-03-29
Added the action_runner for localhost on the example_cluster

solarkennedy:
This is a slight departure from how the example-cluster worked before.

Due to the (current) way that tron is installed here, I can't use the action_runner on remote boxes. 
But I can with localhost, which is "probably" good enough.

Heck, do we even need "distributed" scheduling in the example cluster? Or can we drop that complexity too? Is it enough for tron to try to schedule to localhost? It still exercises the ssh stuff. Who really cares if the hostname is localhost or is "batch1"?

But I think this may become more important as we invest more into the action_runner and action_status

### DIFF
--- a/example-cluster/config/MASTER.yaml
+++ b/example-cluster/config/MASTER.yaml
@@ -1,18 +1,24 @@
----
-
 ssh_options:
    agent: False
    identities:
-     - /work/example-cluster/insecure_key
+      - /work/example-cluster/insecure_key
+
+action_runner:
+    runner_type: "subprocess"
+    remote_status_path: "/tmp/tron"
+    remote_exec_path: "/work/bin/"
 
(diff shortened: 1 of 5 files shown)


qui:
This looks fine to me. example-cluster/config/* is only used in itest.sh, and itesh.sh doesn't seem to test remote execution anyway.

keymone:
this is slightly off-topic, but do we really want to make batch boxes have all these dependencies? what if we replaced action_runner.py with bash wrapper script and just sent it from trond process right after opening the channel? `echo <content> > $HOME/action_runner.sh` or whatever.

solarkennedy:
I think it is easier to write the code with the assumption that the action_runner.py script is already deployed and ready to go.

PR #409, opened 2018-03-30
add a script to clean up namespace

chlgit:
This will prompt the question "This will delete the configuration for the %s namespace. Proceed? (y/n):" before it deletes each namespace.

### DIFF
--- a/contrib/namespace_cleanup.sh
+++ b/contrib/namespace_cleanup.sh
@@ -1,0 +1,21 @@
+#/bin/bash
+
+ecosystem="stagef"
+
+read -p "Are you at tron-$ecosystem (y/n)?" RES
+echo
+if [ $RES = "y" ]; then
+        for file in /nail/tron/config/*
+        do
+                filename="$(basename $file)"
+                namespace="$(basename $file .yaml)"
+                if [ -f "/nail/etc/services/tron/$ecosystem/$filename" ]; then
+                        echo "$namespace is up to date"
+                else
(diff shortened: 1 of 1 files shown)


contrib/namespace_cleanup.sh
```diff
@@ -0,0 +1,21 @@
+#/bin/bash
+
+host="stagef"
+
+read -p "Are you at tron-$host (y/n)?" RES
+echo
+if [ $RES = "y" ]; then
+        for file in /nail/tron/config/*
+        do
+                filename="$(echo $file | cut -d'/' -f5)"
```
solarkennedy: use the `basename` command

contrib/namespace_cleanup.sh
```diff
@@ -0,0 +1,21 @@
...
+                namespace="$(echo $filename | cut -d'.' -f1)"
+                if [ -f "/nail/etc/services/tron/$host/$filename" ]; then
+                        echo "$namespace is up to date"
+                else
+                        echo "$namespace is left behind, delete the namespace?"
```
solarkennedy: just say "deleting the namespace", because that is what it is really doing

contrib/namespace_cleanup.sh
```diff
@@ -0,0 +1,21 @@
+#/bin/bash
+
+host="stagef"
```
solarkennedy: ecosystem

PR #411, opened 2018-04-02
update namespace cleanup script

chlgit:
Load namespace from _manifest.yaml. And it would cat the file before popping the question to confirm deleting namespace.


contrib/namespace_cleanup.sh
```diff
@@ -5,14 +5,20 @@ ecosystem="stagef"
...
+        for namespace in $(cat /nail/tron/config/_manifest.yaml | uq | jq -r 'keys[]')
         do
-                filename="$(basename $file)"
-                namespace="$(basename $file .yaml)"
+                file=$(cat /nail/tron/config/_manifest.yaml | uq | jq -r .\"$namespace\")
+                filename=$(basename $file)
                 if [ -f "/nail/etc/services/tron/$ecosystem/$filename" ]; then
                         echo "$namespace is up to date"
+                elif [ $namespace == "MASTER" ]; then
+                        echo "It is MASTER namepsace"
```
qui: This means that MASTER doesn't exist in /nail/etc/services/tron/$ecosystem/, right? If we're going to cover that case, we should make the message seem more dire :)

contrib/namespace_cleanup.sh
```diff
@@ -5,14 +5,20 @@ ecosystem="stagef"
...
                 else
                         echo "$namespace is left behind, delete the namespace"
```
qui: Can we say "deleting the namespace"? This makes it seem like action is required from you.

contrib/namespace_cleanup.sh
chlgit: No, it is just because of MASTER namespace mapping to master.yaml in _manifest.yaml. We could check if MASTER.yaml exists. I think the scenario with no MASTER.yaml is pretty rare, and if it does happen, we might know that before this script running.

contrib/namespace_cleanup.sh
chlgit: Sure.

contrib/namespace_cleanup.sh
```diff
@@ -5,14 +5,20 @@ ecosystem="stagef"
...
+                filename=$(basename $file)
                 if [ -f "/nail/etc/services/tron/$ecosystem/$filename" ]; then
                         echo "$namespace is up to date"
+                elif [ $namespace == "MASTER" ]; then
+                        echo "It is MASTER namepsace"
                 else
-                        echo "$namespace is left behind, delete the namespace"
+                        echo "$namespace is left behind, deleting the namespace"
+                        echo "========= $filename ========="
+                        cat /nail/tron/config/$filename
```
qui: Maybe we should skip deletion if there are jobs in the file?
Something like `cat /nail/tron/config/$filename | uq | jq -r ".jobs | length"`

chlgit: That sounds good, thx.

PR #412, opened 2018-04-03
install ssh-client for example-cluster/start.sh

chlgit:
Install ssh-client because it's used by example-cluster/start.sh


yelp_package/trusty/Dockerfile
```diff
@@ -15,6 +15,7 @@ RUN apt-get update > /dev/null && \
         libyaml-dev \
         libssl-dev \
         libffi-dev \
+        ssh-client \
```
solarkennedy: just junk the ssh agent stuff, we don' really need it inside example cluser

PR #415, opened 2018-04-03
Run the example_cluster under faketime 10x faster than real life

solarkennedy:
A step closer to a kind of "prod smoketest" where we can see if a release can handle all the jobs on prod for real "for the next month".


example-cluster/start.sh
```diff
@@ -2,4 +2,5 @@
 pip install -e .
 eval $(ssh-agent) || true
```
chlgit: Remove this as well?

PR #417, opened 2018-04-04
Disconnect state version from release version

keymone:
Currently when state file is created, it can only be read by the same or more recent version of tron even if there were no changes to the schema. Let's assume current version is `0` and increment in future when backward-incompatible changes are made. This will allow current (and previous) releases of tron to read the state file.

### DIFF
--- a/tron/serialize/runstate/statemanager.py
+++ b/tron/serialize/runstate/statemanager.py
@@ -8,7 +8,6 @@ from contextlib import contextmanager
 
 import six
 
-import tron
 from tron.config import schema
 from tron.core import job
 from tron.serialize import runstate
@@ -60,7 +59,8 @@ class StateMetadata(object):
     RunState interface as Jobs and Services.
     """
     name = 'StateMetadata'
-    version = tron.__version_info__
+    # version = tron.__version_info__
+    version = (0, 7, 0, 0)
 
     def __init__(self):
         self.state_data = {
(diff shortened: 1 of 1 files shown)


tron/serialize/runstate/statemanager.py
```diff
@@ -60,7 +59,8 @@ class StateMetadata(object):
     RunState interface as Jobs and Services.
     """
     name = 'StateMetadata'
-    version = tron.__version_info__
+    # version = tron.__version_info__
```
chlgit: I think you can delete this.

tron/serialize/runstate/statemanager.py
```diff
@@ -60,7 +59,8 @@ class StateMetadata(object):
...
+    version = (0, 7, 0, 0)
```
chlgit: Move it to tron/__init__.py maybe?

keymone: `tron/__init__.py` is for actual tron release version, this is state schema version (effectively only the first component).

qui: Does the 7 matter? Why have the tuple if you're only planning on using the first number?
Or is this supposed to be related to the Tron version number in some way?

keymone: making it not a tuple will break compatibility of state metadata with our 0.7.x releases. 7 doesn't really matter.

qui: Ah, right, rollback safety... Got it. Is there a way we can make it more obvious what this version means and when we should bump it?

PR #418, opened 2018-04-04
Add retry action

keymone:
This is still work in progress. `tronctl retry <actionrun>` "works", tronview doesn't really reflect this apart from started/finished times changing.

- [x] add retry action to the controller and tronctl
- [x] bump retries_remaining and kill/fail the actionrun
- [x] display retry info in tronview for jobrun/actionrun
- [x] make sure dependent actionruns resume after retry succeeds
- [x] repair actionrun state if it was cleaned up
- [x] repair jobrun state if it was cleaned up
- [x] show retry details in tronweb


bin/tronctl
```diff
@@ -23,6 +23,7 @@ from tron.commands.cmd_utils import ExitCode
 COMMAND_HELP = (
     ('start',           'Start the selected job, job run, or action'),
     ('rerun',           'Re-run a full job (all actions) with a new job id'),
+    ('retry',           'Restart a job action within the same job run'),
```
qui: maybe, "Re-run a job action within an existing job run"?

example-cluster/start.sh
```diff
@@ -1,7 +1,18 @@
-#/bin/sh
+#!/usr/bin/env bash
+
+if ! service ssh status; then
+  echo Setting up SSH
+  apt-get -qq -y install ssh
+  mkdir -p ~/.ssh
+  cp example-cluster/insecure_key ~/.ssh/id_rsa
+  cp example-cluster/insecure_key.pub ~/.ssh/aithorized_keys
```
qui: typo in authorized

tron/commands/display.py
```diff
@@ -346,7 +346,8 @@ class DisplayActionRuns(TableDisplay):
         ('Bare command',        'raw_command'),
         ('Start time',          'start_time'),
         ('End time',            'end_time'),
-        ('Exit status',         'exit_status'),
+        ('Last exit status',    'exit_status'),
```
qui: How about final instead of last? Last could mean "most recent".

tron/core/actionrun.py
```diff
@@ -327,10 +327,13 @@ def _done(self, target, exit_status=0):
             return self.machine.transition(target)
 
     def retry(self):
-        # kill if in flight
         if self.retries_remaining is None or self.retries_remaining <= 0:
             self.retries_remaining = 1
-        self.kill(final=False)
+
+        if self.is_done:
```
qui: Have you thought about what exit_status and exit_statuses will be in these two cases? And how that will look to the developer from tronview.

keymone: yeah good point, this is not the cleanest form, i'll change it to not use `fail`

tron/core/actionrun.py
```diff
@@ -329,18 +329,25 @@ def _done(self, target, exit_status=0):
             return self.machine.transition(target)
 
     def retry(self):
-        # kill if in flight
-        if self.retries_remaining is None or self.retries_remaining <= 0:
+        if self.retries_remaining is None or self.retries_remaining < 0:
```
qui: Are you sure this should still be < instead of <=, now?

PR #419, opened 2018-04-04
Fix namespace with non-ascii crashing tronfig

keymone:
`tronfig -p` fails when namespace contains non-ascii characters.

### DIFF
--- a/bin/tronfig
+++ b/bin/tronfig
@@ -166,7 +166,11 @@ if __name__ == '__main__':
     client = Client(args.server)
 
     if args.print_config:
-        print(client.config(args.source)['config'])
+        # TODO: use maybe_encode()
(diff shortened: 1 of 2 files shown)


bin/tronfig
```diff
@@ -166,7 +166,11 @@ if __name__ == '__main__':
     client = Client(args.server)
 
     if args.print_config:
-        print(client.config(args.source)['config'])
+        # TODO: use maybe_encode()
```
solarkennedy: Can we not just .encode('utf8') all the time?

keymone: that would break python3 compatibility, i'll clean it all up after we upgrade

keymone:
@qui i haven't noticed any problems in config manager yet

PR #421, opened 2018-04-05
monitoring alert if any action becomes failed or unknown

chlgit:


bin/check_tron_jobs.py
```diff
@@ -138,9 +138,20 @@ def get_relevant_run_and_state(job_runs):
...
         if state in ["failed", "succeeded", "unknown"]:
             return run, State(state)
+        if state in ["running"]:
+            action_state = is_action_failed_or_unknown(run)
+            if action_state != State.SUCCEEDED:
+                return run, action_state
     return job_runs['runs'][0], State.WAITING_FOR_FIRST_RUN
 
 
+def is_action_failed_or_unknown(action_runs):
```
qui: action_runs would more accurately be called job_run?

PR #422, opened 2018-04-05
Deal with requestargs encoding in get_string()

keymone:
encode/decode in requestargs.get_string, use get_string in other get_ funcs

### DIFF
--- a/tests/api/requestargs_test.py
+++ b/tests/api/requestargs_test.py
@@ -20,15 +20,17 @@ class RequestArgsTestCase(TestCase):
     @setup
     def setup_args(self):
         self.args = {
-            'number':   ['123'],
-            'string':   ['astring'],
-            'boolean':  ['1'],
-            'datetime': ['2012-03-14 15:09:26'],
+            b'number':   [b'123'],
+            b'string':   [b'astring'],
+            b'boolean':  [b'1'],
+            b'datetime': [b'2012-03-14 15:09:26'],
         }
         self.datetime = datetime.datetime(2012, 3, 14, 15, 9, 26)
         self.request = Turtle(args=self.args)
 
     def _add_arg(self, name, value):
+        name = name.encode()
+        value = value.encode()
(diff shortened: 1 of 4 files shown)


tests/api/requestargs_test.py
```diff
@@ -20,15 +20,17 @@ class RequestArgsTestCase(TestCase):
...
     def setup_args(self):
         self.args = {
-            'number':   ['123'],
-            'string':   ['astring'],
-            'boolean':  ['1'],
-            'datetime': ['2012-03-14 15:09:26'],
+            b'number':   [b'123'],
+            b'string':   [b'astring'],
+            b'boolean':  [b'1'],
+            b'datetime': [b'2012-03-14 15:09:26'],
```
chlgit: Making part of them be bytes and part of them be str might be more comprehensive for testing?

tron/api/requestargs.py
```diff
@@ -13,20 +13,29 @@ def get_integer(request, key):
...
     return int(value)
 
 
 def get_string(request, key):
     """Returns the first value in the request args for a given key."""
-    if not request.args or key not in request.args:
+    if not request.args:
+        return None
+
+    if key is not bytes:
```
chlgit: Using type(key) would be more consistent with type(val) in the following.

tron/api/requestargs.py
```diff
@@ -13,20 +13,29 @@ def get_integer(request, key):
...
+        key = key.encode()
+
+    if key not in request.args:
```
chlgit: You can check if key is not bytes first. So this can be something like:
```
    if not request.args or key not in request.args:
        return None
```
I guess it might be easier to read?

tests/api/requestargs_test.py
keymone: yeah, but this is emulating twisted's args map and that only ever contains keys/values as bytes

PR #425, opened 2018-04-10
Recover unknown batches

Rob-Johnson:
TODO:
- [x] fix the exit from the recovery script to not raise an exception (twisted related)
- [x] figure out how to recover actionruns without an action_runner in the state (we currently dont bother).
- [x] put the constants in the recovery script into tronfig, rather than hardcoding in the module. (I've actually just assumed that the status path and exec_path is the same as the action_runner).
- [x] figure out what to do with PaasTAActionRuns (ignore them)
- [x] add a check that the action_runner is still alive. if not, we'll be waiting forever for the status file to be updated. This is likely to happen if the batch machine is unexpectedly shut down.
- [ ] mark the batch being recovered as in progress once the recovery starts

Edit: I'm going to let this feature set go out as a first pass and bump recovering unknown jobs *without* a restart for another PR - this'll catch 99% of our issues, and I'd like to get it out fast and iterate.

I'm going to build a deb and install on tronplayground to start experimenting.

### DIFF
--- a/tests/core/job_test.py
+++ b/tests/core/job_test.py
@@ -23,6 +23,7 @@ from tests.testingutils import Turtle
 from tron import actioncommand
 from tron import event
 from tron import node
+from tron import scheduler
 from tron.core import job
 from tron.core import jobrun
 from tron.core.actionrun import ActionRun
@@ -47,6 +48,7 @@ class JobTestCase(TestCase):
                 run_collection=run_collection,
                 action_graph=action_graph,
                 node_pool=self.nodes,
+                action_runner=actioncommand.NoActionRunnerFactory
             )
             autospec_method(self.job.notify)
             autospec_method(self.job.watch)
@@ -140,17 +142,25 @@ class JobTestCase(TestCase):
         assert state_data['enabled']
 
     def test_restore_state(self):
-        run_data = ['one', 'two']
-        job_runs = [Turtle(), Turtle()]
-        self.job.runs.restore_state = lambda r, a, o, c, n: job_runs
-        state_data = {'enabled': False, 'runs': run_data}
+        job_runs = [
+            dict(
+                run_num=i,
+                job_name="thename",
+                run_time="sometime",
+                start_time="start_time",
+                end_time="sometime",
+                cleanup_run=None,
+                runs=[],
+            ) for i in range(0, 3)
+        ]
+        state_data = {'enabled': False, 'runs': job_runs}
 
-        self.job.restore_state(state_data)
+        returned_runs = self.job.restore_state(state_data)
 
         assert not self.job.enabled
-        calls = [mock.call(job_runs[i]) for i in range(len(job_runs))]
+        calls = [mock.call(returned_runs[i]) for i in range(0, len(job_runs))]
+        self.job.runs.runs.extend.assert_has_calls(mock.call(returned_runs))
         self.job.watch.assert_has_calls(calls)
-        self.job.event.ok.assert_called_with('restored')
 
     def test_build_new_runs(self):
         run_time = datetime.datetime(2012, 3, 14, 15, 9, 26)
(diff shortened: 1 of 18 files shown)


tron/core/recovery.py
```diff
@@ -0,0 +1,63 @@
+from __future__ import absolute_import
+from __future__ import unicode_literals
+
+import logging
+
+from tron.actioncommand import NoActionRunnerFactory
+from tron.core.actionrun import ActionRun
+from tron.core.actionrun import SSHActionRun
+
+RECOVERY_BINARY = "/work/bin/recover_batch.py"
```
solarkennedy: call this tron_recover_action or something, the normal python scripts should help make this work with the links file trick to give you a consistent full path.

Or you could do something like using
    remote_exec_path: "/opt/venvs/tron/bin"
to get the path

tron/core/recovery.py
```diff
@@ -0,0 +1,63 @@
...
+
+    recovery_run = SSHActionRun(
+        job_run_id=action_run.job_run_id,
+        name="recovery-%s" % action_run.id,
+        node=action_run.node,
+        bare_command=build_recovery_command(
+            recovery_binary=RECOVERY_BINARY,
+            path="%s/%s/status" % (STATUS_PATH, action_run.id),
+        ),
+        output_path=action_run.output_path,
```
qui: Is the recovery stdout going to the right place? Looks like this the action run constructor appends the action_id again, so it might go somewhere different.

tron/core/actionrun.py
```diff
@@ -179,6 +180,7 @@ def __init__(
         mem=None,
         service=None,
         deploy_group=None,
+        machine=None,
```
qui: Do you still need this?

tron/core/recovery.py
```diff
@@ -0,0 +1,63 @@
...
+    )
+    recovery_action_command = recovery_run.build_action_command()
+    recovery_action_command.write_stdout(
+        "recovering action run %s" % action_run.id,
+    )
+
+    # this line is where the magic happens.
+    action_run.watch(recovery_action_command)
```
qui: You might need to reset the action_run's state machine so that it can transition normally? Because UNKNOWN is considered a terminal state.

qui:
Approach looks sane :)

Recovering actionruns without an action_runner in the state - you mean, if there's no status file at all, try to find the process somehow?

For PaaSTAActionRuns, I think you can filter and ignore them, the recovery process will be different anyway.

Rob-Johnson:
@qui 

|Recovering actionruns without an action_runner in the state - you mean, if there's no status file at all, try to find the process somehow?

Kind of. Existing jobs won't have the action runner state persisted, so when we restore the jobs the NoActionRunnerFactory is assumed. But this'll mean that only jobs persisted after this release goes out will be recoverable - we won't be able to do anything with the runs that are left unknown after the initial restart. 

This is a bit of a false constraint though - really the action_runner is a global thing, rather than per action or job, so we can probably just fallback to getting the info from the global config.

tron/core/recovery.py
Rob-Johnson: seems to be ok. I watched the state machine go unknown -> success. Is that what you mean?

qui: Yep, I didn't think unknown could transition to success, but I was wrong!

qui:
Yeah, I think falling back to the global config is a good idea. You could even always assume the global config?

tron/core/actionrun.py
Rob-Johnson: I added this so I could inject a mock state machine in the tests.

tests/core/job_test.py
```diff
@@ -241,59 +251,53 @@ def test__eq__false(self):
...
+        mock_graph.action_map = {}
+        self.job = mock.Mock(autospec=True)
+        self.job.allow_overlap = False
+        self.job.max_runtime = datetime.timedelta(days=1)
+        self.job_scheduler = job.JobScheduler(job=self.job)
+
+    def test_restore_job_state_launches_recovery_runs(self):
+        job_runs = []
+        job_state_data = {'runs': job_runs, 'enabled': True}
+        # with mock.patch()
```
qui: I guess you should finish this test :)

tests/core/job_test.py
```diff
@@ -241,59 +251,53 @@ def test__eq__false(self):
...
-        self.job_scheduler.job = Turtle(runs=run_collection)
-        self.job_scheduler._set_callback = Turtle()
-        state_data = 'state_data_token'
-        self.job_scheduler.restore_state(state_data)
-        assert_call(self.job_scheduler.job.restore_state, 0, state_data)
-        assert_length(self.job_scheduler._set_callback.calls, 1)
-        assert_call(self.job_scheduler._set_callback, 0, 'a')
+
+        mock_graph.action_map = {}
+        self.job = mock.Mock(autospec=True)
```
qui: I'm not sure if autospec=True does anything when you're just creating a mock (as opposed to using mock.patch). spec_set might do what you want?

tests/core/jobrun_test.py
```diff
@@ -480,32 +479,20 @@ def test_restore_state(self):
                 runs=[],
             ) for i in range(3, -1, -1)
         ]
+        run_collection = jobrun.JobRunCollection(20)
```
qui: don't need anymore, right?

tests/core/recovery_test.py
```diff
@@ -0,0 +1,96 @@
...
+        action_runner = SubprocessActionRunnerFactory(
+            status_path='/tmp/foo',
+            exec_path='/bin/foo',
+        )
+        action_run = SSHActionRun(
+            job_run_id="test.succeeded",
+            name="test.succeeded",
+            node=Mock(),
+            action_runner=action_runner
+        )
```
qui: Missing the assertion

tests/core/job_test.py
```diff
@@ -241,59 +249,70 @@ def test__eq__false(self):
...
+        self.job.get_job_runs_from_state.return_value = mock_runs
+
+        with mock.patch(
+            'tron.core.job.recovery.launch_recovery_actionruns_for_job_runs'
+        ) as mock_launch_recovery:
+            mock_launch_recovery.return_value = mock.Mock(autospec=True)
+            self.job_scheduler.restore_state(job_state_data, mock.Mock())
+            assert self.job.runs.runs == collections.deque(mock_runs)
+
+    def test_restore_job_state_launches_recovery_runs(self):
```
qui: I guess you can just add an assertion about this in the test above, and get rid of this test?

tron/mcp.py
```diff
@@ -144,14 +144,14 @@ def get_job_collection(self):
...
-    def restore_state(self):
+    def restore_state(self, action_runner):
         """Use the state manager to retrieve to persisted state and apply it
         to the configured Jobs.
         """
         self.event_recorder.notice('restoring')
         job_states = self.state_watcher.restore(self.jobs.get_names())
 
-        self.jobs.restore_state(job_states)
+        self.jobs.restore_state(job_states, action_runner)
```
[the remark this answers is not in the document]
Rob-Johnson: yep, you're right. fixed

tron/core/job.py
```diff
@@ -272,12 +272,21 @@ def __init__(self, job):
         self.shutdown_requested = False
         self.watch(job)
 
-    def restore_state(self, job_state_data):
+    def restore_state(self, job_state_data, config_action_runner):
         """Restore the job state and schedule any JobRuns."""
-        self.job.restore_state(job_state_data)
+        job_runs = self.job.get_job_runs_from_state(job_state_data)
+        self.job.runs.runs.extend(job_runs)
```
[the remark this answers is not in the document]
Rob-Johnson: yeah, I renamed the function to deliberately make it less side-effecty. If anything, I should remove the watch from that function, too. Does that work for you?

tron/core/actionrun.py
```diff
@@ -396,24 +405,48 @@ def state_data(self):
...
+                self.service,
+            'deploy_group':
+                self.deploy_group,
+            'retries_remaining':
+                self.retries_remaining,
+            'exit_statuses':
+                self.exit_statuses,
+            'action_runner':
+                None
+                if type(self.action_runner) == NoActionRunnerFactory else {
```
keymone: probably worth moving this out into local var to reduce the diff

tron/core/job.py
```diff
@@ -277,12 +275,23 @@ def __init__(self, job):
...
         self.watch(job)
 
-    def restore_state(self, job_state_data):
+    def restore_state(self, job_state_data, config_action_runner):
         """Restore the job state and schedule any JobRuns."""
-        self.job.restore_state(job_state_data)
+        job_runs = self.job.get_job_runs_from_state(job_state_data)
+        for run in job_runs:
+            self.job.watch(run)
+        self.job.runs.runs.extend(job_runs)
```
keymone: i know we're single-threaded right now, but is this thread safe? i imagine we will eventually parallelize some parts of tron and this line kinda scares me :)

tron/core/recovery.py
```diff
@@ -0,0 +1,97 @@
...
+    if type(action_runner) == NoActionRunnerFactory:
+        log.info(
+            "unable to recover action_run %s: action_run has no action_runner"
+            % action_run.id,
+        )
+        return None
+
+    recovery_run = SSHActionRun(
+        job_run_id=action_run.job_run_id,
+        name="recovery-%s" % action_run.id,
```
keymone: if this object is supposed to be a drop-in replacement for original action run object, are we sure nothing depends on `name` attribute being the original action name?

tron/core/job.py
Rob-Johnson: probably not. but this isn't new code - just shuffled around.

tron/core/recovery.py
Rob-Johnson: afaik the name field is just a vanity field, but I'll double check.

PR #430, opened 2018-04-17
Some cleanups and fix to config manager

keymone:
this is pulled out of #418, will reduce the diff as a side-effect

- remove trond working directory in git
- remove _manifest from example-cluster/tronfig
- ensure ssh-agent is around for example-cluster and itest
- isolate tron working directory in itest
- fix an issue when initial MASTER config could not be created


keymone:
@solarkennedy this switches action_runner to use py3, is that too much to ask from batch hosts? i guess it could work with py27 too.

i'm not entirely sure why have i stumbled across the problem with yaml dependency just now, but action_runner *does* depend on PyYAML.

PR #432, opened 2018-04-18
Fix Xenial build

vkhromov:
Fix for
```
04:15:16 Step 8/8 : WORKDIR /work
04:15:16  ---> Using cache
04:15:16  ---> 5b6c272bd17f
04:15:16 Successfully built 5b6c272bd17f
04:15:16 docker run -t -v /ephemeral/jenkins/workspace/itest_xenial:/work:rw tron-builder-xenial python3.6 -m tox -vv --workdir .tox-docker -e py36
04:15:16 /usr/bin/python3.6: No module named tox
04:15:16 make: *** [test_in_docker_xenial] Error 1
```
Just copy-pasted `tox` installation commands for py3 from the trusty's `Dockerfile`.

### DIFF
--- a/yelp_package/xenial/Dockerfile
+++ b/yelp_package/xenial/Dockerfile
@@ -26,6 +26,10 @@ RUN apt-get -q update && \
         wget \
     && apt-get -q clean
 
+RUN wget https://bootstrap.pypa.io/get-pip.py -O /tmp/get-pip.py
+RUN python3.6 /tmp/get-pip.py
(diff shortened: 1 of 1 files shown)


Rob-Johnson:
There are already lines doing the same for python2.7 - I think you can replace those calls, rather than adding to them

keymone:
@Rob-Johnson it's kinda confusing because dh-virtualenv also requires some of those packages. took some trial and error to make sure no pips from different pythons conflict with each other. we're safer to include both until either dh-virtualenv is updated to use python3.6 or somebody decides to spend time sorting this mess out.

PR #433, opened 2018-04-23
Yapf formatting

keymone:
autopep8 and flake8 are nice, but all they do is make linting errors go away. let's go all the way and auto-format all python files using yapf with pep8 style? plugins for most editors are present and i've replaced relevant pre-commit hooks with yapf hook so it also happens on commit.


solarkennedy:
Hard to really "review" this, but I have faith.

.style.yapf
```diff
@@ -0,0 +1,240 @@
...
+#   }        # <--- this bracket is dedented and on a separate line
+#
+#   time_series = self.remote_client.query_entity_counters(
+#       entity='dev3246.region1',
+#       key='dns.query_latency_tcp',
+#       transform=Transformation.AVERAGE(window=timedelta(seconds=60)),
+#       start_ts=now()-timedelta(days=3),
+#       end_ts=now(),
+#   )        # <--- this bracket is dedented and on a separate line
+dedent_closing_brackets=False
```
Rob-Johnson: can we have this set to True please?

PR #434, opened 2018-04-23
Configuration and validation using pyrsistent

keymone:
Pyrsistent is a library that implements immutable implementation of record/collection types.

Rationale: currently tron internals are all mutable objects with schema and validation definitions in separate places, that form a hierarchy and have non-trivial interactions driven by events that cause state machine transitions. getting rid of mutation is a good first step to simplifying this setup, on top of that pyrsistent provides nice facilities to keep schema and validation in one place - next to record declarations.

Pros:

- config parsing and validation will pretty-much go away
- invariants will ensure objects can't be placed in invalid state
- immutable objects make concurrency, debugging and testing simpler

Cons:

- have worse performance
- less convenient interface

https://github.com/Yelp/Tron/blob/pyrsistent/tron/core/action.py is where the meat of the change is, the rest is making itests pass


tron/core/actiongraph.py
```diff
@@ -19,59 +14,38 @@ def __init__(self, graph, action_map):
...
-            for name, conf in six.iteritems(actions_config)
-        }
-        if cleanup_action_config:
-            cleanup_action = action.Action.from_config(cleanup_action_config)
-            actions[maybe_decode(cleanup_action.name)] = cleanup_action
+        if cleanup_action:
+            action_map = action_map.set(cleanup_action.name, cleanup_action)
 
-        return cls(cls._build_dag(actions, actions_config), actions)
+        return cls(*cls._build_dag(action_map))
```
solarkennedy: This unpacking is "tricky" to me, but I see how it is cleaner than the previous method

tron/node.py
```diff
@@ -464,7 +463,7 @@ def _connect(self):
         #  1. Transport is created (our client creator does this)
         #  2. Our transport is secure, and we can create our connection
         #  3. The connection service is started, so we can use it
-
+        from twisted.internet import reactor
```
solarkennedy: What's up with this? Was this required to get the tests to pass? do you know why?

keymone: basically now since action.py depends on node.py (because it validates node_pool being in node pool repo, which we will eventually refactor) and node.py depends on twisted reacotr, it was conflicting with another bit of code that was setting up the reactor. i've removed our own reactor setup as in newer version of twisted it already does the right thing, but there are still remnants of me trying to shuffle code around to keep it from breaking. i'll reduce this diff eventually if we decide to merge this.

tron/trondaemon.py
```diff
@@ -8,26 +8,18 @@
...
 import pkg_resources
 import six
 from twisted.internet import defer
-from twisted.internet.main import installReactor
 from twisted.python import log as twisted_log
 
 import tron
 from tron.utils import flockfile
 
-if platform.system() == 'Linux':
```
solarkennedy: Is this simply not needed anymore with the new version of twisted? 
I'm pretty sure we originally had this here to support OSX

qui:
I like the way this cleans up validation and parsing, and immutability sounds good. 
But I'm wondering what you all think this would look like if we wanted to fully integrate it. Would this work for ActionRuns? Would we end up with fully immutable state, and if not, where would the boundary be?

qui:
I think we could definitely replace a lot/all of the config code with this, where TronConfig is just a big PRecord.

What do you think about the rest of the state, though? I'm having trouble imagining how the watching/notifications on state changes would work. Fundamentally, we have to transition the run's state machine (https://github.com/Yelp/Tron/blob/master/tron/node.py#L597, similar stuff in tron/mesos.py on the MesosTask). How are we going to replace that action run in the parent job/job run?

PR #435, opened 2018-04-24
maybe_encode all data in the file serializer

qui:
Confirmed that this fixes the issue we saw on stage

### DIFF
--- a/tron/serialize/filehandler.py
+++ b/tron/serialize/filehandler.py
@@ -15,6 +15,8 @@ from subprocess import Popen
 
 import six
 
+from tron.utils import maybe_encode
+
 log = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


tron/serialize/filehandler.py
```diff
@@ -57,13 +59,13 @@ def write(self, content):
         """Write content to the fh. Re-open if necessary."""
         if self._fh == NullFileHandle:
             try:
-                self._fh = open(self.name, 'a')
+                self._fh = open(self.name, 'ab')
```
solarkennedy: I'm kinda afraid that going from text/binary will lead to other unintended consequences.
Does this handle newlines like before?

qui: I think newlines are fine, the output from the failing job on stage had newlines.

But I didn't think about cons of binary in general? I guess passing in `encoding=utf-8` instead of changing this to binary would probably fix the issue as well. I can test that.

tron/serialize/filehandler.py
```diff
@@ -57,13 +59,13 @@ def write(self, content):
...
             except IOError as e:
                 log.error("Failed to open %s: %s", self.name, e)
                 return
 
         self.last_accessed = time.time()
-        self._fh.write(content)
+        self._fh.write(maybe_encode(content))
```
solarkennedy: I think this is the "right" place to encode (just in time)

PR #436, opened 2018-04-24
Added retry docs

solarkennedy:


docs/jobs.rst
```diff
@@ -122,6 +122,10 @@ Optional Fields
     Node or node pool to run the action on if different from the rest of the
     job.
 
+**retries** (beta)
+    An integer representing how many times Tron is allowed to retry an action.
```
qui: will automatically retry an action, or something like that?

PR #438, opened 2018-04-27
add support for expected_runtime alerting

chlgit:
I have checked that the new check_tron_job script won't be broken without expected_runtime returned from API. So, updating check_tron_job.py only would be fine (no hurt but no improvement either).


tron/api/adapter.py
```diff
@@ -311,6 +313,12 @@ def get_runs(self):
     def get_max_runtime(self):
         return str(self._obj.max_runtime)
 
+    def get_expected_runtime(self):
+        return str(self._obj.expected_runtime)
+
+    def get_actions_expected_runtime(self):
+        return self._obj.action_graph.expected_runtime
```
qui: Does this need `str()`? Also, will this be a timedelta or the string that was passed in? If it's already a timedelta, might be easier to convert the timedelta to seconds here in the adapter, rather than doing it in the script.

tron/core/actionrun.py
```diff
@@ -194,6 +194,7 @@ def __init__(
         deploy_group=None,
         retries_remaining=None,
         exit_statuses=None,
+        expected_runtime=None,
```
qui: Is expected_runtime needed in actionrun? If it is, you might need to handle it to/from saved state.

tests/bin/check_tron_jobs_test.py
```diff
@@ -91,59 +91,141 @@ def test_get_relevant_action_picks_the_one_that_failed(self):
...
         assert_equal(actual["state"], "failed")
 
     def test_get_relevant_action_picks_the_first_one_succeeded(self):
         action_runs = [
             {
-                'id': 'MASTER.test.action1',
-                'action_name': 'action1',
-                'state': 'succeeded',
-                'start_time': time.localtime(time.time() - 1200),
+                'id':
```
solarkennedy: I don't usually nitpick on style, but can you put this back? Or is the auto-formatting doing this? This is not "normal" k/v style

docs/jobs.rst
```diff
@@ -83,6 +83,11 @@ Optional Fields
     None, which means it will use the default time_zone set in the master
     config.
 
+**expected_runtime** (default **None**)
+    A time interval (ex: "2 hours") that specified the maximal duration of each job run.
```
solarkennedy: say "maximum expected duration"

Then clarify that this is not a hard limit, and if they need a hard limit, use max_runtime

tests/bin/check_tron_jobs_test.py
chlgit: Yeah, it's a little weird. Yapf does this.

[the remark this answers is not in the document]
chlgit: It's auto-formatting from yapf.

tests/api/resource_test.py
```diff
@@ -219,6 +219,7 @@ def setup_resource(self):
             scheduler=mock.Mock(),
             node_pool=mock.create_autospec(node.NodePool, ),
             max_runtime=mock.Mock(),
+            expected_runtime=mock.Mock(),
```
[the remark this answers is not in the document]
chlgit: Hmmm....., it should be...... I don't know actually. It looks like an attribute on the job_runs, but so does max_runtime? And it saves a couple of bytes for the return values of API by being an attribute on an individual job rather than on the job_runs.

tron/api/adapter.py
chlgit: Yeah, it should have str(). I thought about and decided to keep it consistent with other timedelta objects, such as max_runtime.

tests/bin/check_tron_jobs_test.py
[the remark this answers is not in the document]
chlgit: Urrr.......better now?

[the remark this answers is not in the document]
chlgit: Cool, finally figure out the way to put them on the same line \o/

bin/check_tron_jobs.py
```diff
@@ -78,7 +80,10 @@ def compute_check_result_for_job_runs(client, job, job_content):
     )
     action_runs = client.job(job_run_id.url, include_action_runs=True)
     # A job action is like MASTER.foo.1.step1
-    relevant_action = get_relevant_action(action_runs["runs"], last_state)
+    actions_expected_runtime = job_content.get('actions_expected_runtime', {})
+    relevant_action = get_relevant_action(
+        action_runs["runs"], last_state, actions_expected_runtime
```
solarkennedy: please kwargs, at 3 arguments I want to make it immune to ordering issues

bin/check_tron_jobs.py
```diff
@@ -176,32 +181,81 @@ def is_job_scheduled(job_runs):
...
+
     for run in sorted(
         job_runs['runs'],
         key=lambda k: k['run_time'],
         reverse=True,
     ):
         if run.get('state', 'unknown') == "running":
+            # check if runtime exceeds expected_runtime
+            if expected_runtime_str is not None:
+                starttime = run.get('start_time', None)
```
solarkennedy: you made a `is_action_exceeding_expected_runtime`, can you wrap this up into an `is_job_exceeding_expected_runtime` ? I think it would make it easier to read

solarkennedy: job_run maybe

bin/check_tron_jobs.py
```diff
@@ -176,32 +181,81 @@ def is_job_scheduled(job_runs):
...
+        action_name = action.get('action_name', None)
+        if action_name in actions_expected_runtime and actions_expected_runtime[
+            action_name
+        ] is not None:
+            action_seconds = _get_delta_seconds_from_str(
+                actions_expected_runtime[action_name]
+            )
+            action_starttime = action.get('start_time', None)
+            if time.mktime(_timestamp_to_timeobj(action_starttime)
+                           ) + action_seconds < time.time():
```
solarkennedy: I'm surprised to see time.time() here. I guess a running action is running "now", but this would be easier if tron just told you how long it has been running for. This is easy to do, please do that instead. Otherwise this is fine I think.

tests/bin/check_tron_jobs_test.py
```diff
@@ -100,92 +100,104 @@ def test_get_relevant_action_picks_the_first_one_succeeded(self):
...
-        assert_equal(actual["id"], "MASTER.test.action2")
+        actions_expected_runtime = {'action3': '8m', 'action2': '12m', 'action1': '15m'}
+        actual = check_tron_jobs.get_relevant_action(action_runs, State.STUCK, actions_expected_runtime)
+        assert_equal(actual["id"], "MASTER.test.1.action3")
 
     def test_job_succeeded(self):
         job_runs = {
             'status':
-            'running',
+                'running',
```
solarkennedy: I still think this form of dictionary printing is weird.  The indentation is better though.

I guess I don't care so much as long as things are consistent.

tron/core/job.py
```diff
@@ -96,6 +97,7 @@ def __init__(
         time_zone=None,
         service=None,
         deploy_group=None,
+        expected_runtime=None
```
solarkennedy: Why *not* give this a sane default like 24h ?

It doesn't auto-kill, but any job/action can get stuck for lots of reasons.

tests/bin/check_tron_jobs_test.py
chlgit: Yeah, it's pretty strange. I can't figure out why yapf insists on a new line here.

tron/api/adapter.py
chlgit: Ok, I change my mind. It's not consistent for timedelta objects either. So I will convert it to seconds in Tron rather than in the script.

tron/core/job.py
chlgit: Yeah, make sense.

tron/core/action.py
```diff
@@ -57,6 +50,9 @@ def __init__(
         self.mem = mem
         self.service = service
         self.deploy_group = deploy_group
+        self.expected_runtime = expected_runtime or datetime.timedelta(
```
qui: Same, I think this should be in ValidateAction in config_parse instead.

docs/jobs.rst
```diff
@@ -83,6 +83,11 @@ Optional Fields
     None, which means it will use the default time_zone set in the master
     config.
 
+**expected_runtime** (default **24h**)
+    A time interval (ex: "2 hours") that specified the maximum expected duration of each job run.
```
qui: super nit but this really should be "specifies"

tron/core/job.py
```diff
@@ -113,6 +115,9 @@ def __init__(
         self.time_zone = time_zone
         self.service = service
         self.deploy_group = deploy_group
+        self.expected_runtime = expected_runtime or datetime.timedelta(
```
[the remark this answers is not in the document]
chlgit: Good idea.

chlgit:
@qui does Sensu only page when the status code equals to 2? I can't find documents about that. If it pages for any non-zero status code, teams would be paged if they set page=True in the configuration file.

solarkennedy:
@chlgit I'm sure it is documented "somewhere", but yes we only page for 2 (critical):
https://github.com/Yelp/sensu_handlers/blob/master/files/pagerduty.rb#L50

PR #441, opened 2018-05-11
set check_every in monitoring script

chlgit:
set check_every in monitoring script, so the developers can set alert_after in the config file.

### DIFF
--- a/bin/check_tron_jobs.py
+++ b/bin/check_tron_jobs.py
@@ -18,8 +18,7 @@ from tron.utils import timeutils
 
 log = logging.getLogger('check_tron_jobs')
 
-# This script run every 300 seconds currently
-RUN_INTERVAL = 300
+_run_interval = None
 
 
 class State(Enum):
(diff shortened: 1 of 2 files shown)


bin/check_tron_jobs.py
```diff
@@ -45,6 +44,13 @@ def parse_cli():
         default=None,
         help="Check a particular job. If unset checks all jobs",
     )
+    parser.add_argument(
+        "--run-interval",
+        help="Run interval for this monitoring script, default is 300 seconds",
```
qui: You can use %(default)s instead of hardcoding the default, it will get the value passed in below.

Also, can we make it more clear that this doesn't actually control how often the script runs? Maybe you can add something like, "Used to calculate re-alerting settings."

And add the units (seconds) to the help string.

bin/check_tron_jobs.py
```diff
@@ -367,6 +375,9 @@ def main():
     client = Client(args.server)
 
     error_code = 0
+
+    global _run_interval
```
solarkennedy: I think this is overkill. 

How often are we actually going to change this number? I bet never and think it is ok to be a constant hard-coded value.

chlgit: Maybe one day we want to set it as one minute to get alerting earlier 🤔

PR #442, opened 2018-05-14
fix a bug at check_tron_jobs scripts

chlgit:


tests/bin/check_tron_jobs_test.py
```diff
@@ -715,6 +715,96 @@ def test_job_running_action_exceeds_expected_runtime(self):
         assert_equal(run['id'], 'MASTER.test.2')
         assert_equal(state, State.STUCK)
 
+    def test_get_relevant_action_when_action_exceeds_expected_runtime(self):
```
qui: How is this case different from test_job_running_action_exceeds_expected_runtime ? Looks the same.

Does this test fail before you make the fix in bin/check_tron_jobs.py, and then pass afterwards?

chlgit: Oops, my bad. This should be removed.

PR #446, opened 2018-05-22
Fix reloading config ignores changes in `enabled`

vkhromov:


qui:
With this change, if someone manually enables/disables with `tronctl`, the next reconfiguration will overwrite that update (since `tronctl` won't update tronfig accordingly).  It was removed in https://github.com/Yelp/Tron/pull/154/files for that reason.

We should either use tronfig as the state for this, or the state file, but not both. If we want to use tronfig as the source of truth (which makes sense to me), we should remove tronctl enable/disable?

chlgit:
Ah, I see. qui++. Yeah, using tronfig and remove tronctl enable/disable sounds a reasonable approach for me.

PR #448, opened 2018-05-24
run tasks on Mesos

qui:
Updates the API to be more general (no more paasta, just mesos and docker arguments) and uses taskproc to submit tasks to the mesos cluster.

Works for the action configurations I've tried so far. To do:
- [x] unit tests
- [ ] ~~error handling/retries when creating the framework~~ I want to make a follow-up change for this, since the diff is big already.
- [x] more manual testing

### DIFF
--- a/tron/config/config_parse.py
+++ b/tron/config/config_parse.py
@@ -35,9 +35,12 @@ from tron.config.schedule_parse import valid_schedule
 from tron.config.schema import CLEANUP_ACTION_NAME
 from tron.config.schema import ConfigAction
 from tron.config.schema import ConfigCleanupAction
+from tron.config.schema import ConfigConstraint
 from tron.config.schema import ConfigJob
+from tron.config.schema import ConfigParameter
 from tron.config.schema import ConfigSSHOptions
 from tron.config.schema import ConfigState
+from tron.config.schema import ConfigVolume
 from tron.config.schema import MASTER_NAMESPACE
 from tron.config.schema import NamedTronConfig
 from tron.config.schema import NotificationOptions
@@ -145,12 +148,45 @@ def valid_node_name(value, config_context):
     return value
 
 
-def valid_cluster_name(value, config_context):
-    valid_string(value, config_context)
-    if not config_context.partial and value not in config_context.clusters:
-        msg = "Unknown cluster name %s at %s"
-        raise ConfigError(msg % (value, config_context.path))
-    return value
+class ValidateConstraint(Validator):
+    config_class = ConfigConstraint
+    validators = {
+        'attribute':
+            valid_string,
+        'operator':
+            config_utils.build_enum_validator(
+                # TODO: import from taskproc
+                ['EQUALS', 'NOTEQUALS', 'LIKE', 'UNLIKE']
+            ),
+        'value':
+            valid_string,
+    }
+
+
+valid_constraint = ValidateConstraint()
+
+
+class ValidateDockerParameter(Validator):
+    config_class = ConfigParameter
+    validators = {
+        'key': valid_string,
+        'value': valid_string,
+    }
+
+
+valid_docker_parameter = ValidateDockerParameter()
+
+
+class ValidateVolume(Validator):
+    config_class = ConfigVolume
+    validators = {
+        'container_path': valid_string,
+        'host_path': valid_string,
+        'mode': config_utils.build_enum_validator(['RO', 'RW']),
+    }
+
+
+valid_volume = ValidateVolume()
 
 
 class ValidateSSHOptions(Validator):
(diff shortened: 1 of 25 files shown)


qui:
Tested with the following `dev/config/MASTER.yaml` and `make dev`. The constraint format is a little different from what I put in paasta_tools, thought it would be easier to verify if it was more explicit.

```
ssh_options:
   agent: False

nodes:
  - hostname: localhost

jobs:
    - name: "test"
      node: localhost
      schedule: "interval 60min"
      actions:
        - name: "first"
          command: "echo hello world"
          executor: mesos
          docker_image: busybox
          mesos_address: mesos.paasta-norcal-devc.yelp
          cpus: 1
          mem: 100
          extra_volumes:
              - container_path: '/nail/etc/mrjob'
                host_path: '/nail/etc/mrjob'
                mode: RO
          constraints:
              - attribute: pool
                operator: LIKE
                value: default
          env:
              SERVICE: my-test-service
          docker_parameters:
            - key: memory-swap
              value: 1088m
            - key: cpu-period
              value: '100000'
            - key: cpu-quota
              value: '250000'
            - key: label
              value: service=my-test-service
            - key: label
              value: instance=tron
```

qui:
@keymone @Rob-Johnson finished adding tests and merging conflicts, ready for final reviews!

tron/config/config_parse.py
```diff
@@ -275,43 +306,81 @@ def valid_action_name(value, config_context):
...
     """Validate an action."""
     config_class = ConfigAction
 
     defaults = {
         'node': None,
         'requires': (),
+        'retries': None,
+        'expected_runtime': datetime.timedelta(hours=24),
         'executor': schema.ExecutorTypes.ssh,
-        'cluster': None,
```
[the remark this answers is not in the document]
qui: Yeah, I *think* this will be fine. Config parsing is only for current tronfig, so works as long as the old fields don't exist in tronfig.

For objects from state, we use restore_state, and that should just ignore fields that it doesn't use.

tron/mesos.py
```diff
@@ -0,0 +1,272 @@
...
+from tron.actioncommand import ActionCommand
+from tron.utils.dicts import get_deep
+from tron.utils.queue import PyDeferredQueue
+
+TASK_LOG_FORMAT = '%(asctime)s %(name)s %(levelname)s %(message)s'
+# TODO: put in configs
+MESOS_MASTER_PORT = 5050
+DOCKERCFG_LOCATION = "file:///root/.dockercfg"
+MESOS_SECRET = ''
+MESOS_ROLE = '*'
```
Rob-Johnson: I wonder if, in the future, we'll want to make this a per task thing. From the framework's point of you, it has to tell the master which roles it wants to subscribe to, but I think it can re-register with more roles (so you could re-register when an action with a previously-unseen role is added). Fine to pass on this for now and figure it out later, was just a thought.

tron/mesos.py
```diff
@@ -0,0 +1,272 @@
...
+OFFER_TIMEOUT = 300
+
+log = logging.getLogger(__name__)
+_processor = None
+frameworks = {}  # TODO: improve
+
+
+def get_mesos_processor():
```
Rob-Johnson: any need to have this as a seperate function, rather than just creating it in the constructor of the MesosCluster object?

tron/mesos.py
```diff
@@ -0,0 +1,272 @@
...
+
+
+def get_mesos_leader(master_address):
+    url = "http://%s:%s/redirect" % (master_address, MESOS_MASTER_PORT)
+    response = requests.get(url)
+    return '{}:{}'.format(urlparse(response.url).hostname, MESOS_MASTER_PORT)
+
+
+def get_mesos_cluster(master_address):
+    global frameworks
```
Rob-Johnson: I'm not a huge fan of the global use here - should the dict of frameworks belong to the caller (and shutdown_frameworks take the dictionary as a parameter?)

tron/mesos.py
```diff
@@ -0,0 +1,272 @@
...
+
+    def get_runner(self, mesos_address, queue):
+        executor = self.processor.executor_from_config(
+            provider='mesos',
+            provider_config={
+                'secret': MESOS_SECRET,
+                'mesos_address': get_mesos_leader(mesos_address),
+                'role': MESOS_ROLE,
+                # TODO: could also be in config, to include Tron cluster name
+                'framework_name': 'tron',
```
Rob-Johnson: should this be parameterised if we're leaving the possibility of >1 framework open (the frameworks object in this module is a dict, rather than a single object).

tron/mesos.py
qui: Hmm, the direct caller is an ActionRun. Keeping the dict there wouldn't enforce one framework per cluster, and it seems hard to get all the clusters back up to the MCP or trondaemon.

As a cleanup item, I was planning on making something similar to NodePoolRepository for frameworks to enforce the singleton-ness, instead of the current global. How does that sound?

tron/mesos.py
qui: The idea was that the processor might be expensive to create, but after reading the TaskProcessor code, seems fine. Also, only creating it once per cluster seems okay anyway. I can get rid of this.

tron/mesos.py
qui: There's only supposed to be one framework per cluster, so it shouldn't register the same framework name on the same cluster more than once. But adding the parameter seems like a good idea anyway, so I'll do that.

qui: Er, just realized that we don't really have the Mesos cluster name, just the fqdn. Is there a different parameter you think would be useful?

tron/mesos.py
Rob-Johnson: yep, that works.

tron/mesos.py
Rob-Johnson: how'd you mean? this is the gramework name, rather than the cluster name?

PR #449, opened 2018-05-29
deploy prod namespaces and jobs at playground

chlgit:
The code still needs some cleanup. 

To load the namespaces and jobs at tron-playground, we can copy the scripts to the playground, and run
  $ sync_namespaces_jobs.py
  $ deploy_all_prod_namespace.sh

To remove these namespaces and jobs, run 
  $ delete_all_prod_namespace.sh

With loading these namespaces and jobs from prod, tronview at the playground is still pretty fast:

real	0m1.750s
user	0m0.351s
sys	0m0.031s

I plan to run more experiments tomorrow. Let me know if you have any thought.


contrib/sync_namespaces_jobs.py
```diff
@@ -0,0 +1,49 @@
...
+#DST_DIR = "/nail/home/chl/tron-servdir"
+SRC_DIR = "/nail/etc/services/tron/prod"
+
+#SRC_DIR = "/nail/home/chl/git/yelpsoa-configs/tron/prod"
+#DST_DIR = "/nail/home/chl/Tron/contrib/tron-servdir"
+
+
+def main():
+    print(
+        "This script will load the same amount of namepance and jobs from tron-prod to tron-playground."
```
qui: Seems like this script doesn't do any loading, it just copies and modifies some files from SRC_DIR to DST_DIR. Probably don't need the warning? And this note (or the docstring) should describe what the script does better.

contrib/sync_namespaces_jobs.py
```diff
@@ -0,0 +1,49 @@
...
+                                del config[key]
+
+                    jobs = config.get("jobs", [])
+                    if jobs is not None:
+                        for job in jobs:
+                            job['node'] = "localhost"
+                            if 'monitoring' in job:
+                                del job['monitoring']
+                            for action in job.get("actions", []):
+                                action['command'] = '/bin/true'
```
qui: Some sleeping might be more realistic, keeping track of running jobs probably adds some load.

contrib/sync_namespaces_jobs.py
```diff
@@ -0,0 +1,49 @@
...
+#DST_DIR = "/nail/home/chl/Tron/contrib/tron-servdir"
+
+
+def main():
+    print(
+        "This script will load the same amount of namepance and jobs from tron-prod to tron-playground."
+    )
+    print("Please make sure you run it at playground")
+
+    for (dirpath, dirname, filenames) in os.walk(SRC_DIR):
```
qui: nit: listdir, since you're assuming that every file is directly in SRC_DIR (e.g. located at SRC_DIR/filename) anyway. Then you can get rid of this outer loop.

qui:
You can start a separate trond with a different `--working-dir` and `--config-path` instead doing this loading/deletion on the currently running trond. Seems easier, and that way you don't have to worry about conflicts with existing playground namespaces.

Also, do you want this to work after the migration to service directories starts, or if PaaSTA actions start getting added? It doesn't have to, but to handle those cases, you could copy (rsync?) configs directly from the prod working directory instead of from soaconfigs. so you don't have to worry about any of the translation steps.

contrib/sync_namespaces_jobs.py
```diff
@@ -0,0 +1,57 @@
...
+                        job['node'] = "localhost"
+                        if 'monitoring' in job:
+                            del job['monitoring']
+                        for action in job.get("actions", []):
+                            action['command'] = 'sleep 10s'
+                            if "node" in action:
+                                action['node'] = "localhost"
+            for i in range(args.multiple):
+                out_filepath = os.path.join(
+                    DST_DIR, 'prod_' + str(i) + '-' + filename
```
qui: maybe f-string would be nicer?

contrib/sync_namespaces_jobs.py
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env python
+import argparse
+import os
+
+import yaml
+
+DST_DIR = "/tmp/tron-servdir"
```
qui: Let's make destination and source arguments as well. This will make it easier to use, especially when we move configs out of tron/prod.

contrib/sync_namespaces_jobs.py
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env python
+import argparse
```
qui: How about we add a docstring that explains why you'd want to use this script? something about load testing, etc. You can have a longer description here, than what's in the argument parser description.

contrib/sync_namespaces_jobs.py
```diff
@@ -0,0 +1,57 @@
...
+
+def main():
+    args = parse_args()
+    for filename in os.listdir(SRC_DIR):
+        print('filename = {}'.format(filename))
+        filepath = os.path.join(SRC_DIR, filename)
+        if os.path.isfile(filepath) and filepath.endswith(".yaml"):
+            with open(filepath, "r") as f:
+                config = yaml.load(f)
+                if filename == "MASTER.yaml":
```
qui: After you load the config, the rest of this code doesn't have to be in the `with` context manager.

contrib/sync_namespaces_jobs.py
```diff
@@ -0,0 +1,57 @@
...
+
+import yaml
+
+DST_DIR = "/tmp/tron-servdir"
+SRC_DIR = "/nail/etc/services/tron/prod"
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+        description='Creating namespaces and jobs configuration from Tron prod'
```
qui: If the source is an argument, this description should be more general than "Tron prod".

contrib/sync_namespaces_jobs.py
```diff
@@ -0,0 +1,72 @@
...
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+        description='Creating namespaces and jobs configuration for load testing'
+    )
+    parser.add_argument(
+        '--multiple',
+        type=int,
+        default=1,
+        help='multiple workload of namespaces and jobs from Tron prod'
```
qui: from source directory

chlgit: Oops, qui++

PR #451, opened 2018-06-04
set the machine state to 'running' before recovery

Rob-Johnson:
without this, the state machine remains in the unknown state until the
recovery job terminates.

### DIFF
--- a/tron/core/actionrun.py
+++ b/tron/core/actionrun.py
@@ -150,6 +150,10 @@ class ActionRun(object):
     STATE_FAILED['skip'] = STATE_SKIPPED
     STATE_CANCELLED['skip'] = STATE_SKIPPED
 
+    STATE_UNKNOWN['running'] = STATE_RUNNING
+    STATE_UNKNOWN['failed'] = STATE_FAILED
+    STATE_UNKNOWN['success'] = STATE_SUCCEEDED
+
     # We can force many states to be success or failure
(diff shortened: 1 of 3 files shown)


tron/core/recovery.py
```diff
@@ -70,6 +70,14 @@ def recover_action_run(action_run, action_runner):
     # and updates its internal state acoording to its result.
     action_run.watch(recovery_action_command)
 
+    if not action_run.machine.check('running'):
+        log.error(
```
keymone: don't we want to error out of here? i assume if action run is not `running`, killing it won't kill the recovered process and other similar edge cases?

Rob-Johnson: I think it'll be ok? you can still move to a terminal state from unknown.

tron/core/actionrun.py
```diff
@@ -150,6 +150,10 @@ class ActionRun(object):
     STATE_FAILED['skip'] = STATE_SKIPPED
     STATE_CANCELLED['skip'] = STATE_SKIPPED
 
+    STATE_UNKNOWN['running'] = STATE_RUNNING
+    STATE_UNKNOWN['failed'] = STATE_FAILED
```
qui: There's some magic in `__getattr__` that makes all of the transition names correspond to a method on an action run.  This should be 'fail' since we already have a method called that, which handles that transition. (It's also done below in L157-160, but I don't mind making it explicit here.)

Rob-Johnson: oh, great, thanks.

PR #454, opened 2018-06-12
ClusterRepository for accessing Mesos clusters

qui:
Basically the same as using a global, but now it's namespaced under the class? So it's slightly nicer to change configuration and stuff like that.

Decided to do it this way instead of copying NodePoolRepository (https://github.com/Yelp/Tron/blob/62ebd9b8c45d4bce49c5943b1f319ee6c8f3aa85/tron/node.py#L60) since it doesn't actually work as a singleton... you can get different instances with NodePoolRepository() as many times as you want, it just enforces that get_instance() returns the same one. Didn't seem worth the extra code, but let me know if you like that better.

The enabling is proof-of-concept for how configuration would be done, need to follow up on using that configuration value properly in MesosCluster and adding the field to MASTER.

### DIFF
--- a/tron/mesos.py
+++ b/tron/mesos.py
@@ -23,7 +23,6 @@ MESOS_ROLE = '*'
 OFFER_TIMEOUT = 300
 
 log = logging.getLogger(__name__)
-frameworks = {}  # TODO: improve
 
 
 def get_mesos_leader(master_address):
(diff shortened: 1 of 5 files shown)


tron/mesos.py
```diff
@@ -32,17 +31,33 @@ def get_mesos_leader(master_address):
     return '{}:{}'.format(urlparse(response.url).hostname, MESOS_MASTER_PORT)
 
 
-def get_mesos_cluster(master_address):
-    global frameworks
-    if master_address not in frameworks:
-        frameworks[master_address] = MesosCluster(master_address)
-    return frameworks[master_address]
+class ClusterRepository:
```
keymone: `MesosClusterRepository` ?

tron/mesos.py
```diff
@@ -32,17 +31,33 @@ def get_mesos_leader(master_address):
...
 
+    clusters = {}
+    mesos_enabled = False
 
-def shutdown_frameworks():
-    global frameworks
-    for name, framework in frameworks.items():
-        framework.stop()
+    def __init__(self):
+        super(ClusterRepository, self).__init__()
```
keymone: why is this needed?

tron/mesos.py
```diff
@@ -32,17 +31,33 @@ def get_mesos_leader(master_address):
...
+
+    @classmethod
+    def get_cluster(cls, master_address):
+        if master_address not in cls.clusters:
+            cls.clusters[master_address] = MesosCluster(
```
keymone: if leader election results in different master address, will we have multiple frameworks in single cluster? we should probably index by cluster name?

tron/mesos.py
qui: It's not 🤷‍♀️

tron/mesos.py
qui: The master address is the fqdn, which points to all of the masters. It's like "mesos-norcal-prod.yelp" which is basically the cluster name.

PR #455, opened 2018-06-13
Bounded asynchronous API request handling

keymone:
- [x] run all get requests asynchronously
- [x] figure out locking for POSTs

all GET request handlers use semaphore with capacity=10 to limit number of concurrent requests, POST requests acquire exclusive lock and drain capacity from semaphore before proceeding.

in this PR also:

- Remove reactor initialization
we've upgraded twisted a while ago, it already has the right logic to pick reactor implementation by default.

### DIFF
--- a/tests/api/resource_test.py
+++ b/tests/api/resource_test.py
@@ -4,6 +4,8 @@ Test cases for the web services interface to tron
 from __future__ import absolute_import
 from __future__ import unicode_literals
 
+from functools import wraps
+
 import mock
 import six
 import twisted.web.http
@@ -26,7 +28,14 @@ from tron import event
 from tron import mcp
 from tron import node
 from tron.api import controller
-from tron.api import resource as www
+with mock.patch(
+    'tron.api.async_resource.AsyncResource.bounded', lambda fn: fn
+):
+    with mock.patch(
+        'tron.api.async_resource.AsyncResource.exclusive', lambda fn: fn
+    ):
+        from tron.api import resource as www
+
 from tron.core import job
 from tron.core import jobrun
 
--- a/tests/api/resource_test.py
+++ b/tests/api/resource_test.py
@@ -4,8 +4,6 @@ Test cases for the web services interface to tron
 from __future__ import absolute_import
 from __future__ import unicode_literals
 
-from functools import wraps
-
 import mock
 import six
 import twisted.web.http
--- a/tron/api/async_resource.py
+++ b/tron/api/async_resource.py
@@ -1,5 +1,7 @@
 import threading
 
+from twisted.internet import threads
+
 
 class AsyncResource():
     capacity = 10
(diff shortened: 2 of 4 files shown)


tests/api/resource_test.py
```diff
@@ -4,6 +4,8 @@
 from __future__ import absolute_import
 from __future__ import unicode_literals
 
+from functools import wraps
```
Rob-Johnson: don't think you need this?

tron/api/async_resource.py
```diff
@@ -0,0 +1,48 @@
+import threading
+
+from twisted.internet import threads
+from twisted.web import server
+
+
+class AsyncResource():
+    capacity = 10
```
Rob-Johnson: this seems kind of...low. Is this based on you testing?

keymone: just a wild guess. we can always bump this number if things are too slow and concurrency is the answer.

tests/api/resource_test.py
keymone: maybe i don't, mock.patch was making me jump through some hoops, let me check.

keymone:
@Rob-Johnson the problem right now is that every request is blocking the event loop for the whole duration of the request. we never do any async stuff in api handlers. the whole event loop just stops and waits.

tron/api/async_resource.py
```diff
@@ -0,0 +1,48 @@
...
+
+        return wrapper
+
+    @staticmethod
+    def exclusive(fn):
+        def wrapper(*args, **kwargs):
+            # ensures only one exclusive request starts consuming the semaphore
+            with AsyncResource.lock:
+                # this will wait until all bounded requests finished processing
+                for _ in range(AsyncResource.capacity):
```
chlgit: This might reduce the concurrency of the system. For example, it would acquire 9 semaphores and wait for the last one. During that period, the concurrency decreases to 1. And the period for acquiring all semaphores would become longer if we increase the capacity. What if we don't wait until all bounded requests finished? Maybe it's too aggressive...

tron/api/resource.py
```diff
@@ -6,8 +6,10 @@
 from __future__ import unicode_literals
 
 import collections
+import contextlib
```
qui: More unused imports here? Everything except AsyncResource

tron/api/async_resource.py
qui: https://twisted.readthedocs.io/en/twisted-18.4.0/core/howto/threading.html suggests that the max thread pool is 10. Maybe that's why larger numbers caused errors?

tron/api/async_resource.py
keymone: my goal here was to ensure that during any POST request no GET requests are being processed because that could cause data race issues. we can relax this constraint later as we implement more granular locks around critical pieces of data.

PR #459, opened 2018-06-20
Improve exit handling

Rob-Johnson:


bin/action_runner.py
```diff
@@ -85,8 +85,9 @@ def run_proc(output_path, command, run_id, proc):
         run_id=run_id,
         proc=proc,
     ):
-        proc.wait()
-    sys.exit(proc.returncode)
+        returncode = proc.wait()
+        log.warn(f'pid {proc.pid} exited with returncode {returncode}')
```
keymone: `.warn` is deprecated iirc

PR #463, opened 2018-06-20
[wip] Trigger scheduler

keymone:
Create a special scheduler `TriggerScheduler` that will only launch jobs after receiving a relevant event. Allow jobs to originate trigger events upon completion.

- [x] split up schedulers into separate modules
- [ ] create new kind of scheduler that reacts to trigger events from related jobs
- [ ] create job configuration option to publish a trigger event when job is finished
- [ ] change schedulers to publish job run events rather than passively waiting for `next_run_time` calls


tron/scheduler/common.py
```diff
@@ -0,0 +1,23 @@
+from __future__ import absolute_import
+from __future__ import unicode_literals
+
+import datetime
+import logging
+import random
+
+from tron.utils import timeutils
+
+log = logging.getLogger(__name__)
```
qui: Do we really want this in `common`, as opposed to different loggers per scheduler? This way, the logger will always be 'tron.scheduler.common'. It's not used right now anyway.

keymone: yeah, i kinda overlooked that, will make a separate logger in each scheduler module

PR #464, opened 2018-06-20
implement kill terminate for mesos actions

chlgit:
There's no difference between stopping a task or killing a task for task processing, so I use the same call except that "tronctl kill" would print the extra line saying "Warning: Mesos action might not be killed forcibly."

### DIFF
--- a/tron/api/controller.py
+++ b/tron/api/controller.py
@@ -77,8 +77,11 @@ class ActionRunController(object):
 
     def handle_termination(self, command):
         try:
-            getattr(self.action_run, command)()
+            # Extra message is only used for killing mesos action as warning so far.
+            extra_msg = getattr(self.action_run, command)()
             msg = "Attempting to %s %s"
+            if extra_msg is not None:
+                msg = msg + "\n" + extra_msg
             return msg % (command, self.action_run)
         except NotImplementedError as e:
             msg = "Failed to %s: %s"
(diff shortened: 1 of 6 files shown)


tron/api/controller.py
```diff
@@ -77,8 +77,11 @@ def handle_command(self, command):
 
     def handle_termination(self, command):
         try:
-            getattr(self.action_run, command)()
+            # Extra message is only used for killing mesos action as warning so far.
+            extra_msg = getattr(self.action_run, command)()
```
solarkennedy: This is cool, and in the future it would be nice to give some feedback to the user for the ssh-based killing too. (like.. did it work?)

tron/core/actionrun.py
```diff
@@ -616,13 +616,18 @@ def stop(self):
...
-        pass
+        self._kill_mesos_task()
 
     def kill(self, final=True):
         if self.retries_remaining is not None and final:
             self.retries_remaining = -1
 
-        pass
+        self._kill_mesos_task()
+        return "Warning: Mesos action might not be killed forcibly."
```
solarkennedy: I don't really understand, why wouldn't it? Mesos totally does a kill -9 if it doesn't die in like, 2 minutes or whatever.

chlgit: From https://github.com/Yelp/task_processing/blob/master/task_processing/plugins/mesos/logging_executor.py#L232, it looks like the kill function won't tell if it's a kill or a kill -9. Or am I missing anything?

solarkennedy: Its kinda deep into mesos land
http://mesos.apache.org/documentation/latest/configuration/agent/
executor_shutdown_grace_period (used to be docker_stop_timeout)
Really it is taken care of by docker, it will kill the container eventually, even if it has to do -9. We don't really get a choice.

PR #466, opened 2018-06-21
Add default volumes and some other Mesos task settings to master config

qui:
MESOS_MASTER_PORT, MESOS_SECRET, MESOS_ROLE would require changing the framework on reconfiguration, so leaving those for later.

### DIFF
--- a/tron/config/config_parse.py
+++ b/tron/config/config_parse.py
@@ -583,10 +583,20 @@ class ValidateMesos(Validator):
     option = True
     defaults = {
         'enabled': False,
+        'default_volumes': (),
+        'dockercfg_location': None,
+        'offer_timeout': 300,
     }
 
     validators = {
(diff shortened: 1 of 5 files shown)


tron/config/config_parse.py
```diff
@@ -583,10 +583,20 @@ class ValidateMesos(Validator):
     option = True
     defaults = {
         'enabled': False,
+        'default_volumes': (),
+        'dockercfg_location': None,
```
solarkennedy: I think the default got lost here. Is this where the `file:///` goes or is that piece of configuration moving somewhere else?

qui: Was planning to get it from system paasta config (https://github.com/Yelp/paasta/pull/1868) but I could put file:///root/.dockercfg back as a default if that's a fairly standard value.

PR #468, opened 2018-06-22
handle failures streaming  to stdout/stderr

Rob-Johnson:
if stdout or stderr disappear from under us and the thread reading them
crashes, then the stderr and stdout of the subprocess being run can
backup, and possibly prevent it exiting (see TRON-339 for an example of
docker doing this). catch any exceptions we get writing so that we
continue to read from the stderr and stdout of the subprocess.

example output in ``/tmp/{run_id}.log`` launching a pid that is interrupted:

```
WARNING:root:streaming 4 to <stdout>
WARNING:root:<stdout>: b'hi\n'
WARNING:root:streaming 6 to <stderr>
WARNING:root:foo-bar-123 running as pid 31721
WARNING:root:<stderr>: b'error\n'
WARNING:root:<stdout>: b'hi\n'
WARNING:root:<stderr>: b'error\n'
WARNING:root:<stdout>: b'hi\n'
WARNING:root:<stderr>: b'error\n'
WARNING:root:<stdout>: b'hi\n'
WARNING:root:<stderr>: b'error\n'
WARNING:root:failed writing to <_io.TextIOWrapper name='<stdout>' mode='w' encoding='UTF-8'>: [Errno 32] Broken pipe
WARNING:root:b'hi\n'
WARNING:root:failed writing to <_io.TextIOWrapper name='<stderr>' mode='w' encoding='UTF-8'>: [Errno 32] Broken pipe
WARNING:root:b'error\n'
WARNING:root:<stdout>: b'hi\n'
WARNING:root:<stderr>: b'error\n'
```

Note: this change means that we'll now log the stderr and stdout to the local disk, as well as streaming it back to tron. I thought it'd make for easier debugging, but will probably mean we need to bump the size of our disks?

### DIFF
--- a/bin/action_runner.py
+++ b/bin/action_runner.py
@@ -86,7 +86,7 @@ def run_proc(output_path, command, run_id, proc):
         proc=proc,
     ):
         returncode = proc.wait()
-        log.warn(f'pid {proc.pid} exited with returncode {returncode}')
+        log.warning(f'pid {proc.pid} exited with returncode {returncode}')
         sys.exit(returncode)
 
 
@@ -117,15 +117,27 @@ def run_command(command):
 
 
 def stdout_reader(proc):
+    is_connected = True
     for line in iter(proc.stdout.readline, b''):
-        sys.stdout.write(line.decode('utf-8'))
-        sys.stdout.flush()
+        if is_connected:
+            try:
+                sys.stdout.write(line.decode('utf-8'))
+                sys.stdout.flush()
+            except Exception as e:
+                log.warning(f'failed writing to stdout: {e}')
+                is_connected = False
 
 
 def stderr_reader(proc):
+    is_connected = True
     for line in iter(proc.stderr.readline, b''):
-        sys.stderr.write(line.decode('utf-8'))
-        sys.stdout.flush()
+        if is_connected:
+            try:
+                sys.stderr.write(line.decode('utf-8'))
+                sys.stderr.flush()
+            except Exception as e:
+                log.warning(f'failed writing to stderr: {e}')
+                is_connected = False
 
 
 if __name__ == "__main__":
--- a/bin/action_runner.py
+++ b/bin/action_runner.py
@@ -126,7 +126,7 @@ def stream(source, dst):
                 logging.warning(f'{dst.name}: {line}')
             except Exception as e:
                 logging.warning(f'failed writing to {dst}: {e}')
-                logging.warning(line)
+                logging.warning(f'{dst.name}: {line}')
                 is_connected = False
         else:
             logging.warning(f'{dst.name}: {line}')
--- a/bin/action_runner.py
+++ b/bin/action_runner.py
@@ -117,39 +117,32 @@ def run_command(command):
 
 def stream(source, dst):
     is_connected = True
-    logging.warning(f'streaming {source} to {dst}')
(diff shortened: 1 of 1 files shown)


bin/action_runner.py
```diff
@@ -116,30 +115,43 @@ def run_command(command):
...
+            try:
+                dst.write(line.decode('utf-8'))
+                dst.flush()
+                logging.warning(line)
+            except Exception as e:
+                logging.warning(f'failed writing to {dst}: {e}')
+                logging.warning(line)
+                is_connected = False
+        else:
+            logging.warning(line)
```
keymone: worth prefixing with stream type

bin/action_runner.py
```diff
@@ -116,30 +115,43 @@ def run_command(command):
...
+            is_connected = False
 
 
-def stderr_reader(proc):
-    for line in iter(proc.stderr.readline, b''):
-        sys.stderr.write(line.decode('utf-8'))
-        sys.stdout.flush()
+def configure_logging(run_id):
+    logging.basicConfig(filename=f'/tmp/{run_id}.log')
```
keymone: `/tmp/action_runner.<pid>.<run_id>.log`?

bin/action_runner.py
```diff
@@ -116,30 +115,43 @@ def run_command(command):
...
-        target=stdout_reader, args=(proc, ), daemon=True
-    )
-    stderr_printer_t = threading.Thread(
-        target=stderr_reader, args=(proc, ), daemon=True
-    )
-    stdout_printer_t.start()
-    stderr_printer_t.start()
+    streaming_threads = [
+        threading.Thread(
+            target=stream, args=(
```
keymone: `args=pair`?

bin/action_runner.py
```diff
@@ -116,30 +115,43 @@ def run_command(command):
...
     proc = run_command(args.command)
-    stdout_printer_t = threading.Thread(
-        target=stdout_reader, args=(proc, ), daemon=True
-    )
-    stderr_printer_t = threading.Thread(
-        target=stderr_reader, args=(proc, ), daemon=True
-    )
-    stdout_printer_t.start()
-    stderr_printer_t.start()
+    streaming_threads = [
```
keymone: i don't think we really need this array, lets just

    for p in [...]:
        threading.Thread(...).start()

?

bin/action_runner.py
Rob-Johnson: hm. we've not got the pid at this point, so can't do this.

bin/action_runner.py
```diff
@@ -116,30 +115,36 @@ def run_command(command):
...
+    logging.warning(f'streaming {source.name} to {dst.name}')
+    for line in iter(source.readline, b''):
+        if is_connected:
+            try:
+                dst.write(line.decode('utf-8'))
+                dst.flush()
+                logging.warning(f'{dst.name}: {line}')
+            except Exception as e:
+                logging.warning(f'failed writing to {dst}: {e}')
+                logging.warning(line)
```
qui: Add dst.name here too?

Rob-Johnson:
@qui this actually causes *all* stderr and stdout to be written to the batch host - (I've done that deliberately to make debugging a bit easier than having to stitch it together from 2 sources)

PR #473, opened 2018-06-28
fix tronfig -C bug

chlgit:
Not very elegant solution to fix this bug. But since Qui is working on new paasta validation (https://jira.yelpcorp.com/browse/TRON-326), so this part of codes would retire soon. Maybe it's not worth much time on this one.

### DIFF
--- a/tron/config/manager.py
+++ b/tron/config/manager.py
@@ -158,6 +158,9 @@ class ConfigManager(object):
     def get_namespaces(self):
         return self.manifest.get_file_mapping().keys()
 
+    def check_config(self, name, content):
+        log.info("checking config for {}".format(name))
+
 
 def create_new_config(path, master_content):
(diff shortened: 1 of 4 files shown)


solarkennedy:
I agree this isn't very elegant. Can you paste in a copy of the stack trace for the bug itself?

tron/config/manager.py
```diff
@@ -158,6 +158,9 @@ def __contains__(self, name):
     def get_namespaces(self):
         return self.manifest.get_file_mapping().keys()
 
+    def check_config(self, name, content):
```
qui: You can take out the call to `check_config` from the API controller, since this doesn't do anything.

chlgit: Yeah, this can work too. I was assuming that we can add some checks here. But since we will get rid of this soon, removing this call is better.

PR #474, opened 2018-06-29
Remove the graceful_shutdown aspect of Tron

solarkennedy:
We have recovery now. I don't think we would ever realistically wait for jobs to actually finish.

Also I did some yapf junk, I can undo if you want.

### DIFF
--- a/tron/trondaemon.py
+++ b/tron/trondaemon.py
@@ -150,8 +150,6 @@ class NoDaemonContext(object):
 class TronDaemon(object):
     """Daemonize and run the tron daemon."""
 
-    WAIT_SECONDS = 5
-
     def __init__(self, options):
         self.options = options
         self.mcp = None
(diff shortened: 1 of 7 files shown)


tron/trondaemon.py
```diff
@@ -214,23 +214,6 @@ def _handle_shutdown(self, sig_num, stack_frame):
...
-        self.mcp.graceful_shutdown()
-        self._wait_for_jobs()
-
-    def _wait_for_jobs(self):
-        if self.mcp.jobs.is_shutdown:
-            self._handle_shutdown(None, None)
-            return
-
-        log.info("Waiting for jobs to shutdown.")
-        reactor.callLater(self.WAIT_SECONDS, self._wait_for_jobs)
```
qui: Do we need WAIT_SECONDS anymore?

solarkennedy: nope, unused variable. removed.

PR #475, opened 2018-06-29
wait daemon thread finish before exit

chlgit:
It looks fine through running action_runner manually, but I am not sure how to reproduce the error in [TRON-355](https://jira.yelpcorp.com/browse/TRON-355). Any thought?

### DIFF
--- a/bin/action_runner.py
+++ b/bin/action_runner.py
@@ -85,7 +85,7 @@ def run_proc(output_path, command, run_id, proc):
     ):
         returncode = proc.wait()
         logging.warning(f'pid {proc.pid} exited with returncode {returncode}')
-        sys.exit(returncode)
+        return returncode
 
 
 def parse_args():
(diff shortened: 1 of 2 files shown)


Rob-Johnson:
hm, I don't think these threads will ever finish unless they're explicitly told to - we probably need to add some event signalling to tell the threads to finish when the subprocess is finished.

chlgit:
@qui is able to reproduce the error at the playground. And the daemon threads can exit after run_proc, which makes thread.join() work.

qui:
The repro was:
`  lines = ['line {}'.format(i) for i in range(1000)]`
`  print('\n'.join(lines))`
(See qui_test.hello_world in playground.) I think it works because there's a lot of stdout to flush at once. Anyway, this triggered `Fatal Python error: could not acquire lock for <_io.BufferedWriter name='<stdout>'> at interpreter shutdown, possibly due to daemon threads` every time it ran.

@chlgit and I tested the patch to action_runner here and the repro case didn't trigger the bug afterwards. So the threads evidently do finish. I think when the subprocess exits, it closes the pipes for stderr/stdout, so the iterators and thus the threads exit.

PR #478, opened 2018-07-06
Fix validation of full tronfig directory

qui:
tronfig needs one positional argument (source), so let's use that as the directory if -D is specified.

### DIFF
--- a/bin/tronfig
+++ b/bin/tronfig
@@ -61,7 +61,7 @@ def parse_cli():
     parser.add_argument(
         "-D",
         "--validate-dir",
-        action="store",
+        action="store_true",
         dest="validate_dir",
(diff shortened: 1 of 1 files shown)


bin/tronfig
```diff
@@ -128,7 +128,7 @@ def validate_dir(path):
         manifest.create()
         for fname in os.listdir(path):
             name, ext = os.path.splitext(fname)
-            if ext == 'yaml':
+            if ext == '.yaml':
```
solarkennedy: Gosh. did this ever work? :)

solarkennedy:
Looks fine. But is this enough to actually fix it?
https://pb.yelpcorp.com/320685

PR #479, opened 2018-07-09
Delay between actionrun retries

keymone:
- [x] add relevant action and actionrun attributes
- [x] delay start transition when delay attribute is non-zero
- [x] store information about delay in state
- [x] correctly restore delayed actionrun from state
- [x] display delay information in tronview/tronweb
- [x] write tests

### DIFF
--- a/tron/commands/display.py
+++ b/tron/commands/display.py
@@ -392,19 +392,19 @@ class DisplayEvents(TableDisplay):
     resize_fields = ['entity']
 
 
-def display_node(source, _):
+def display_node(source, _=None):
     if not source:
         return ''
     return '%s@%s' % (source['username'], source['hostname'])
 
 
-def display_node_pool(source, _):
+def display_node_pool(source, _=None):
     if not source:
         return ''
     return "%s (%d node(s))" % (source['name'], len(source['nodes']))
 
 
-def display_scheduler(source, _):
+def display_scheduler(source, _=None):
     if not source:
         return ''
     return "%s %s%s" % (source['type'], source['value'], source['jitter'])
--- a/tron/commands/display.py
+++ b/tron/commands/display.py
@@ -414,7 +414,7 @@ def display_state_delayed(_, obj):
     state = obj['state']
     in_delay = obj['in_delay']
     if in_delay:
-        return f"{state} (retry delayed for {in_delay}s)"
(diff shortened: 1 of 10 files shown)


tron/core/actionrun.py
```diff
@@ -324,6 +327,13 @@ def from_state(
 
     def start(self):
         """Start this ActionRun."""
+        if self.in_delay is not None:
```
chlgit: I think this part should go to start_after_delay(). Any reason to put them here instead?

tron/commands/display.py
```diff
@@ -394,22 +392,38 @@ class DisplayEvents(TableDisplay):
     resize_fields = ['entity']
 
 
-def display_node(source):
+def display_node(source, _=None):
+    if not source:
```
chlgit: Since the source has been checked for multiple display function, it might be worth to have a function for it.

tron/core/actionrun.py
keymone: we don't want `start` to be called when we're in delay, it could happen from other events.

tron/commands/display.py
keymone: i'd rather not make the diff even larger in less relevant parts of the codebase. this display functionality needs to be refactored anyway.

tron/core/actionrun.py
```diff
@@ -373,18 +383,35 @@ def _done(self, target, exit_status=0):
...
             return self.kill(final=False)
 
+    def start_after_delay(self):
+        log.info(f"Resuming action run {self.id} after backoff")
+        self.in_delay = None
+        self.start()
+
     def restart(self):
+        """Used by `fail` when action run has to be re-tried."""
         self.machine.reset()
```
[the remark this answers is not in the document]
keymone: why? i think "scheduled" represents state of action in delay fairly accurately. without adding a separate state - it's probably the best we got?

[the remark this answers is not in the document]
qui: Hmm, in my opinion, "running" makes more sense. Even if the command isn't currently running, the action has started. Otherwise, you see a job/action that already have a start time, but are still in the "scheduled" state.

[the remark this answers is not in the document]
keymone: i guess it's a matter of preference, i've updated the logic to keep action in running state. everything still works.

PR #480, opened 2018-07-09
fix _get_seconds_from_duration bug

chlgit:
I am sure there won't be any bug from _get_seconds_from_duration anymore.

### DIFF
--- a/bin/check_tron_jobs.py
+++ b/bin/check_tron_jobs.py
@@ -7,7 +7,7 @@ import sys
 import time
 from enum import Enum
 
-import pandas
+import pytimeparse
 from pysensu_yelp import send_event
 
 from tron.commands import cmd_utils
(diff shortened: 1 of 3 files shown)


qui:
Is there a test that failed before this fix, and now passes? If not, please update one of the test cases for expected_runtime.

I might prefer for this to be done with a regex so we don't need a complex dependency like pandas. (Although then we'd need a few more tests.) Happy either way, but wonder what others think.

chlgit:
test_get_relevant_action_pick_the_one_exceeds_expected_runtime_with_long_duration was supposed to test for this case, but it failed because of my bad. Now, it should take the responsibility it aimed at originally.
We could use python-dateutil and datetime for regex. But I don't see any trouble pandas would bring, for now at least.

Rob-Johnson:
pandas seems like a huge dependency to add for one function. Can we not get away with anything...simpler?

PR #482, opened 2018-07-13
send alerts based on job runtime

chlgit:


bin/check_tron_jobs.py
```diff
@@ -153,24 +153,29 @@ def pretty_print_actions(action_run):
...
+        reverse=True,
+    )
+    if len(job_runs) == 0:
         return None, State.NO_RUN_YET
     run = is_job_scheduled(job_runs)
     if run is None:
-        return job_runs['runs'][0], State.NOT_SCHEDULED
-    run = is_job_stuck(job_runs)
+        return job_runs[0], State.NOT_SCHEDULED
+    run = is_job_stuck(job_runs, job_content.get('expected_runtime', None), job_content.get('actions_expected_runtime', {}))
```
solarkennedy: can you break this out and kwargs this please

bin/check_tron_jobs.py
```diff
@@ -181,23 +186,15 @@ def is_action_failed_or_unknown(job_run):
...
 
 
-def is_job_stuck(job_runs):
+def is_job_stuck(job_runs, job_expected_runtime, actions_expected_runtime):
     next_run_time = None
-
-    job_expected_runtime = job_runs.get('expected_runtime', None)
-    actions_expected_runtime = job_runs.get('actions_expected_runtime', {})
-
-    for job_run in sorted(
```
solarkennedy: It is interesting to me that we were already doing this. 

The only issue that I can think of is that this function now "assumes" that the job_runs object is coming in pre-sorted by time.

I think that is "OK", but could be a source of bugs in the future.

bin/check_tron_jobs.py
```diff
@@ -153,24 +153,29 @@ def pretty_print_actions(action_run):
     return display.format_action_run_details(action_run)
 
 
-def get_relevant_run_and_state(job_runs):
-    if len(job_runs['runs']) == 0:
+def get_relevant_run_and_state(job_content):
+    job_runs = sorted(
+        job_content.get('runs', []),
+        key=lambda k: k['run_time'],
```
qui: What exactly is run_time, again? Is it possible for it to cause a key error or be None?

bin/check_tron_jobs.py
chlgit: Sure, fixed.

bin/check_tron_jobs.py
chlgit: run_time is the time that job is scheduled to run. It's assigned when the job is scheduled. So it's always there and won't be None.

bin/check_tron_jobs.py
chlgit: Hmm...maybe. But sorted job_runs is most close to our assumption that is monitoring the most recent one.

PR #483, opened 2018-07-23
Retries delay: kill delayed action correctly

keymone:


qui:
We also need this for stop, right? If you feel like it, it would be nice to refactor this retry logic out, similar to how `submit_command` is implemented by subclasses, but `start` is in the base class.

PR #484, opened 2018-07-24
sort the job run based on endtime and scheduled time

chlgit:
The job runs are sorted by end_time, and then scheduled time. It covers two edge cases: (1) rerun job, which has the same scheduled time with the first job run. (2) manually job run between two scheduled job runs. The details of these edge cases are in TRON-359.

### DIFF
--- a/bin/check_tron_jobs.py
+++ b/bin/check_tron_jobs.py
@@ -156,7 +156,7 @@ def pretty_print_actions(action_run):
 def get_relevant_run_and_state(job_content):
     job_runs = sorted(
         job_content.get('runs', []),
-        key=lambda k: k['run_time'],
+        key=lambda k: (k['end_time'] is None, k['end_time'], k['run_time']),
         reverse=True,
     )
     if len(job_runs) == 0:
--- a/bin/check_tron_jobs.py
+++ b/bin/check_tron_jobs.py
@@ -154,6 +154,11 @@ def pretty_print_actions(action_run):
 
 
 def get_relevant_run_and_state(job_content):
+    # The order of job run to check is as follows:
+    #   1. The scheduled but hasn't run one checked first
+    #   2. Then currently running ones are always checked (in case an action is failed/unknown)
+    #   3. If there are multiple running ones, then most recent run_time wins
+    #   4. If nothing is currently running, then most recent end_time wins
     job_runs = sorted(
         job_content.get('runs', []),
         key=lambda k: (k['end_time'] is None, k['end_time'], k['run_time']),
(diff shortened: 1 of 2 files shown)


tests/bin/check_tron_jobs_test.py
```diff
@@ -439,6 +454,11 @@ def test_most_recent_job_with_smaller_run_id_failed(self):
                             '%Y-%m-%d %H:%M:%S',
                             time.localtime(time.time() - 1800),
                         ),
+                    'end_time':
+                        time.strftime(
+                            '%Y-%m-%d %H:%M:%S',
+                            time.localtime(time.time() - 1700),
```
qui: Maybe this test should be called most_recent_end_time, since the run ID doesn't matter?

tests/bin/check_tron_jobs_test.py
```diff
@@ -562,6 +669,7 @@ def test_job_running_but_action_unknown_already(self):
                             '%Y-%m-%d %H:%M:%S',
                             time.localtime(time.time() - 1800),
                         ),
+                    'end_time': None,
```
qui: This shouldn't be None because it has succeeded.

bin/check_tron_jobs.py
```diff
@@ -156,7 +156,7 @@ def pretty_print_actions(action_run):
 def get_relevant_run_and_state(job_content):
     job_runs = sorted(
         job_content.get('runs', []),
-        key=lambda k: k['run_time'],
+        key=lambda k: (k['end_time'] is None, k['end_time'], k['run_time']),
```
qui: I wrote out a few examples for myself, and it seems like this sort key translates into:
1) Currently running ones are always checked first (in case an action is failed/unknown)
2) If there are multiple running ones, then most recent run_time wins
3) If nothing is currently running, then most recent end_time wins

Is that correct? Can we add a comment about the intended behavior so we/others can remember in the future?

chlgit: Yes, it's right. I will make a comment in the code.

tests/bin/check_tron_jobs_test.py
chlgit: Oops, nice catch.

tests/bin/check_tron_jobs_test.py
chlgit: Sure, that makes sense.

chlgit:
I reorganize these test cases by the monitoring states. All tests with the same state, such as STATE.STUCK, are grouped together.

PR #486, opened 2018-07-31
tronfig checks valid nodes

chlgit:
The command is something like:
   tronfig tron-playground.yaml -V -n robj-test-service -m ../tron/playground/MASTER.yaml

The output is something like:
  Unknown node name batch1-uswest1adevcfff at robj-test-service.NamedConfigFragment.jobs.Job.mesos-test.actions.Action.sleep.node

### DIFF
--- a/tests/config/config_parse_test.py
+++ b/tests/config/config_parse_test.py
@@ -455,6 +455,122 @@ class NamedConfigTestCase(TestCase):
         )
         assert_equal(test_config, expected)
 
+    def test_attributes_with_master_context(self):
+        expected = make_named_tron_config(
+            jobs=FrozenDict({
+                'test_job':
+                    make_job(
+                        name="test_job",
+                        namespace='test_namespace',
+                        schedule=ConfigIntervalScheduler(
+                            timedelta=datetime.timedelta(0, 20),
+                            jitter=None,
+                        ),
+                        expected_runtime=datetime.timedelta(1),
+                    )
+            })
+        )
+        master_context = dict(
+            nodes=[
+                dict(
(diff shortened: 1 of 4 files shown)


qui:
I don't think we should have to add validation code for this, Tron will already check that the nodes are valid if you use `validate_config_mapping`. 

Also the options are getting very convoluted here... I wish we had separate subcommands for validate, print, etc instead of making them all optional arguments, but maybe it's too late to change the interface now?

chlgit:
Separated subcommands is a good idea, but not sure how much time to refactor. Also, I imagine Tronfig is rarely used manually (it's used for deploy or pre-commit hook for most of the use cases), so it's okay to be a little convoluted?

tests/config/config_parse_test.py
```diff
@@ -455,6 +455,128 @@ def test_attributes(self):
...
+
+    def test_invalid_job_node_with_master_context(self):
+        master_context = dict(
+            nodes=[
+                dict(
+                    name="node0",
+                    hostname="node0",
+                )
+            ],
+            node_pools=[
```
qui: nit: let's take these lines out because they don't affect the test.

tron/config/config_parse.py
```diff
@@ -716,12 +716,18 @@ def post_validation(self, config, config_context):
 valid_named_config = ValidateNamedConfig()
 
 
-def validate_fragment(name, fragment):
+def validate_fragment(name, fragment, master_context=None):
```
qui: Can we call this master_config everywhere, since it's not a Context object?

PR #487, opened 2018-08-01
Update taskproc

qui:
* fixes logging (old taskproc was calling basicConfig before)
* stop using deprecated provider
* return feedback on killing tasks


tron/core/actionrun.py
```diff
@@ -681,7 +681,9 @@ def _kill_mesos_task(self):
             mesos_cluster = MesosClusterRepository.get_cluster(
                 self.mesos_address
             )
-            mesos_cluster.kill(self.task_id)
+            succeeded = mesos_cluster.kill(self.task_id)
```
solarkennedy: Now I see what you all were talking about with regards to error handling.

PR #488, opened 2018-08-03
catch command rendering type error

chlgit:
This catch type errors of rendering command, plus a special case like "$(variable) $(variable)s"

### DIFF
--- a/tron/config/config_parse.py
+++ b/tron/config/config_parse.py
@@ -25,6 +25,7 @@ from tron.config.config_utils import build_list_of_type_validator
 from tron.config.config_utils import ConfigContext
 from tron.config.config_utils import PartialConfigContext
 from tron.config.config_utils import valid_bool
+from tron.config.config_utils import valid_context_variable_expr
 from tron.config.config_utils import valid_dict
 from tron.config.config_utils import valid_float
 from tron.config.config_utils import valid_identifier
@@ -67,15 +68,15 @@ def build_format_string_validator(context_object):
         )
 
         try:
-            render_command = value % context
-            # This checks wrong format like "$(A) $(A)s"
-            if render_command != (render_command % context):
-                raise TypeError("type is not specified")
-
+            valid_context_variable_expr(value)
+            value % context
             return value
-        except (KeyError, ValueError, TypeError) as e:
+        except (KeyError, ValueError) as e:
             error_msg = "Unknown context variable %s at %s: %s"
             raise ConfigError(error_msg % (e, config_context.path, value))
+        except (TypeError) as e:
+            error_msg = "Wrong command format %s: %s at %s"
+            raise ConfigError(error_msg % (value, e, config_context.path))
 
     return validator
 
--- a/tron/config/config_parse.py
+++ b/tron/config/config_parse.py
@@ -67,9 +67,13 @@ def build_format_string_validator(context_object):
         )
 
         try:
-            value % context
+            render_command = value % context
+            # This checks wrong format like "$(A) $(A)s"
+            if render_command != (render_command % context):
+                raise TypeError("type is not specified")
+
             return value
-        except (KeyError, ValueError) as e:
+        except (KeyError, ValueError, TypeError) as e:
             error_msg = "Unknown context variable %s at %s: %s"
             raise ConfigError(error_msg % (e, config_context.path, value))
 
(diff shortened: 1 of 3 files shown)


tron/config/config_parse.py
```diff
@@ -67,9 +67,13 @@ def validator(value, config_context):
         )
 
         try:
-            value % context
+            render_command = value % context
+            # This checks wrong format like "$(A) $(A)s"
+            if render_command != (render_command % context):
```
qui: I think this might break commands that escape the string formatting. For example, if the command has `%%Y/%%m/%%d` in it, if you try to format it a second time, it will raise a ValueError.

This is a really tricky situation to catch... "The %(one) something %(stars)s" doesn't raise an error if you format it twice because `something` happens to start with s, which is a valid format type.

Maybe it's enough to detect if rendering the command will cause an error, as we do now. That will prevent people from breaking deployments, at least. But then we can add another feature later that allows you to preview your commands, so that people can manually check that their commands look correct?

tron/config/config_parse.py
```diff
@@ -67,9 +67,13 @@ def validator(value, config_context):
...
+                raise TypeError("type is not specified")
+
             return value
-        except (KeyError, ValueError) as e:
+        except (KeyError, ValueError, TypeError) as e:
```
qui: I agree we should also catch TypeError, but maybe in a separate case? The error_msg doesn't really make sense when it's a type error.

tron/config/config_parse.py
solarkennedy: I agree that we can't "do it twice". @chlgit why isn't the first pass sufficient to detect the bad mapping?

@qui omg you are right:
```
>>> "hello world %(foo)s something" % {'foo': foo}
'hello world hi something'
>>> "hello world %(foo) something" % {'foo': foo}
'hello world hiomething'
```

chlgit: OK, I rewrite the check function. It's not very elegant, but it works.....I hope.

tron/config/config_parse.py
chlgit: Yeah, good point.

tron/config/config_parse.py
chlgit: @solarkennedy  the first pass can't detect something like this:
 >>> "hello world %(foo) %(foo)s something" % {'foo': foo}
'hello world %(foo)s something'

solarkennedy: Ah. Yuck :(

We try so hard to validate this kind of easy mistake, users can easily make this mistake all the time (I have), and there are just so many gotchas.

Wouldn't it be cool if we could just snap our fingers and move to `f{}` format strings? Could here possibly be a migration path to this, given we control all our own code and configs?

tron/config/config_utils.py
```diff
@@ -157,6 +157,18 @@ def valid_name_identifier(value, config_context):
     return '%s.%s' % (config_context.namespace, value)
 
 
+def valid_context_variable_expr(command):
+    """ This funtion checks wrong format like '$(A) $(A)s' """
```
qui: *function

Also, we can be more specific about what "wrong" means. Something like, "verifies that all context variables have complete string format specifiers"?

tron/config/config_utils.py
```diff
@@ -157,6 +157,18 @@ def valid_name_identifier(value, config_context):
...
+    prefix = command.find("%(")
+    while prefix >= 0:
+        postfix = command.find(")s", prefix)
+        if postfix < 0:
+            raise TypeError("Context variable expression is invalid")
```
qui: Maybe this could just be a ConfigError?

tron/config/config_utils.py
chlgit: Yeah, that's better, thx.

tron/config/config_utils.py
chlgit: Sure.

solarkennedy:
So great that we finally have a way for users to catch these typos. Can you also preemptively run this over our existing config (just in case)?

chlgit:
I can confirm that all existing configs pass this test.

tron/config/config_utils.py
```diff
@@ -157,16 +157,18 @@ def valid_name_identifier(value, config_context):
     return '%s.%s' % (config_context.namespace, value)
 
 
-def valid_context_variable_expr(command):
-    """ This funtion checks wrong format like '$(A) $(A)s' """
+def valid_context_variable_expr(command, path):
```
qui: Other validators have the config_context passed in, it should be consistent.

PR #489, opened 2018-08-04
Spell check all the things

solarkennedy:
What could possibly go wrong.

### DIFF
--- a/tron/event.py
+++ b/tron/event.py
@@ -49,8 +49,8 @@ class EventLevel(object):
 
 
 LEVEL_INFO = EventLevel(0, "INFO")  # Troubleshooting information
-LEVEL_OK = EventLevel(1, "OK")  # Expected behaviour
-LEVEL_NOTICE = EventLevel(2, "NOTICE")  # Troubling behaviour
+LEVEL_OK = EventLevel(1, "OK")  # Expected behavior
+LEVEL_NOTICE = EventLevel(2, "NOTICE")  # Troubling behavior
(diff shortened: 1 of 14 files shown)


tron/utils/crontab.py
```diff
@@ -8,7 +8,7 @@
 
 PREDEFINED_SCHEDULE = {
     "@yearly": "0 0 1 1 *",
-    "@anually": "0 0 1 1 *",
+    "@annually": "0 0 1 1 *",
```
solarkennedy: Lol this never worked. Also we don't use this in our configs anywhere.

tron/event.py
```diff
@@ -49,8 +49,8 @@ def __hash__(self):
 
 
 LEVEL_INFO = EventLevel(0, "INFO")  # Troubleshooting information
-LEVEL_OK = EventLevel(1, "OK")  # Expected behaviour
-LEVEL_NOTICE = EventLevel(2, "NOTICE")  # Troubling behaviour
+LEVEL_OK = EventLevel(1, "OK")  # Expected behavior
+LEVEL_NOTICE = EventLevel(2, "NOTICE")  # Troubling behavior
```
EvanKrall: https://www.dictionary.com/browse/behaviour

solarkennedy: Yea but, it is safer if we agree to only one version of spelling, namely US/American.
Otherwise we get (have had) incidents where someone spells `datacentre`.

PR #490, opened 2018-08-07
add string format support

chlgit:
I would like to hear other thought before I proceed...
Instead of formatting string in old fashion way like:
```
%(PYTHON_ENV)s %(BATCHDIR)s/run.py
```
we can also format commands as:
```
{PYTHON_ENV} {BATCHDIR}/run.py
```

### DIFF
--- a/tests/config/config_parse_test.py
+++ b/tests/config/config_parse_test.py
@@ -1143,6 +1143,26 @@ class TestBuildFormatStringValidator(TestCase):
         )
         assert self.validator(template, context)
 
+    def test_validator_valid_format_string(self):
+        template = "The {one} thing I {seven} is {mars}"
+        context = config_utils.ConfigContext(
+            None,
+            None,
+            {'mars': 'ok'},
+            None,
+        )
+        assert self.validator(template, context)
+
+    def test_validator_valid_percent_string(self):
+        template = "The %(one)s %(seven)s thing is %(mars)s -config {'a': 1}"
+        context = config_utils.ConfigContext(
+            None,
+            None,
+            {'mars': 'ok'},
+            None,
+        )
+        assert self.validator(template, context)
+
 
 class TestValidateConfigMapping(TestCase):
     config = dict(**BASE_CONFIG, command_context=dict(some_var="The string"))
--- a/tests/config/config_parse_test.py
+++ b/tests/config/config_parse_test.py
@@ -1146,20 +1146,20 @@ class TestBuildFormatStringValidator(TestCase):
     def test_validator_valid_format_string(self):
         template = "The {one} thing I {seven} is {mars}"
         context = config_utils.ConfigContext(
-            None,
-            None,
-            {'mars': 'ok'},
-            None,
+            path=None,
+            nodes=None,
+            command_context={'mars': 'ok'},
+            namespace=None,
         )
         assert self.validator(template, context)
 
     def test_validator_valid_percent_string(self):
         template = "The %(one)s %(seven)s thing is %(mars)s -config {'a': 1}"
         context = config_utils.ConfigContext(
-            None,
-            None,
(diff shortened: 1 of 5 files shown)


tron/config/config_parse.py
```diff
@@ -67,7 +68,11 @@ def validator(value, config_context):
         )
 
         try:
-            value % context
+            if "%" in value:
```
solarkennedy: This is probably too generous. We have `%` all over the place.

Can we "just" support both? Where we do both types of string interpolation, regardless?

chlgit: I am ok to just support both here, but how could we decide which rendered command we should actually run? I guess we still need a condition to decide. If so, I think it's worth to keep it consistent between validation and action run.

solarkennedy: But why can't we literally do both?
Like format it in both ways. I don't think we need a condition to decide.
Also yes, however formatting happens it should be consistent between the actionrun and the validation.

chlgit: Ah, I see. That makes sense.

chlgit:
The code looks a little weird, but we need that to pass the case like [this](https://pb.yelpcorp.com/322425). Currently, there are 5 commands in prod and 3 commands in dev we need to fix before we can remove the weird try block.

qui:
If we format the command both ways, users will have to know to escape both % and {} in their commands, even if their command only uses one method. For example, you still need the extra % in `command: date -d {shortdate-1} +%%m-%%d-%%Y}` to prevent Tron from formatting the date format.

That makes me think we shouldn't keep both methods long-term. But if we agree that we want to migrate to string formatting and remove % formatting eventually, then keeping both is fine as an aid to migration.

chlgit:
As we discussed at Tron-sync today, I will merge this and try to migrate at devc first.

tron/core/actionrun.py
```diff
@@ -483,7 +484,11 @@ def state_data(self):
 
     def render_command(self):
         """Render our configured command using the command context."""
-        return self.bare_command % self.context
+        try:
+            parse_str = self.bare_command % self.context
+            return StringFormatter(self.context).format(parse_str)
```
solarkennedy: Can you help explain how we will deploy this code safely, given that we have existing `{}`'s?
If we ship this code, won't it break right away on the json stuff?
Then we will get an exception (:409) and then do the old behavior, which is fine.
Should be good to go, right?

It makes me wonder how we ever protected ourselves from exceptions before? idk.

tests/core/actionrun_test.py
```diff
@@ -328,6 +328,16 @@ def test_render_command(self):
         self.action_run.bare_command = "%(stars)s"
         assert_equal(self.action_run.render_command(), 'bright')
 
+    def test_render_format_string_command(self):
+        self.action_run.context = {'stars': 'bright'}
+        self.action_run.bare_command = "{stars}"
+        assert_equal(self.action_run.render_command(), 'bright')
+
+    def test_render_valid_percent_invalid_format_string_command(self):
```
solarkennedy: Nice

tests/config/config_parse_test.py
```diff
@@ -1152,6 +1152,26 @@ def test_validator_passes_with_context(self):
         )
         assert self.validator(template, context)
 
+    def test_validator_valid_format_string(self):
+        template = "The {one} thing I {seven} is {mars}"
+        context = config_utils.ConfigContext(
+            None,
```
solarkennedy: Can you kwargs these please?

tron/core/actionrun.py
chlgit: Yeah, you're right. If anything (exception) happens here, we would just fall back to the old behavior. So I think json stuff should be fine.

PR #491, opened 2018-08-08
Autospec all the things

solarkennedy:
Gotta stave off [the beast](https://engineeringblog.yelp.com/2015/02/assert_called_once-threat-or-menace.html) from going into autospace.

### DIFF
--- a/contrib/mock_patch_checker.py
+++ b/contrib/mock_patch_checker.py
@@ -1,0 +1,78 @@
+#!/usr/bin/env python3.6
+import ast
+import sys
+
+
+class MockChecker(ast.NodeVisitor):
+    def __init__(self):
+        self.errors = 0
+        self.init_module_imports()
+
+    def init_module_imports(self):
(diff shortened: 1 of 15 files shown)


contrib/mock_patch_checker.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def check_file(self, filename):
+        self.current_filename = filename
+        try:
+            with open(filename, 'r') as fd:
+                try:
+                    file_ast = ast.parse(fd.read())
+                except SyntaxError as error:
+                    print("SyntaxError on file %s:%d" % (filename, error.lineno))
+                    return
```
chlgit: Do you want to increase self.errors here so that the pre-commit hook exits with non-zero value?

solarkennedy: Hmm. We don't have that on paasta. I think it is "OK", other pre-commit hooks will catch this and fail way before.

PR #493, opened 2018-08-10
Save Mesos framework IDs and re-register on restart

qui:
Goes with https://github.com/Yelp/task_processing/pull/117

The bulk of this is updating framework IDs in the state file when they change. Did some hacks with MesosClusterRepository instead of subclassing Observable, would be great if you all have ideas on better ways to do this :)


chlgit:
Close this PR in favor of #498

PR #494, opened 2018-08-14
Added Tron + Mesos itest framework

solarkennedy:
If we are going to have this as a feature, I want confidence that I can make a PR to tron and not break it! :)

### DIFF
--- a/cluster_itests/docker-compose.yml
+++ b/cluster_itests/docker-compose.yml
@@ -1,0 +1,38 @@
+version: '2'
+
+services:
+  zookeeper:
+    build: ../yelp_package/itest_dockerfiles/zookeeper/
+    ports:
+      - 2181
+
+  mesosmaster:
+    build: ../yelp_package/itest_dockerfiles/mesos/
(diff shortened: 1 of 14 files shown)


cluster_itests/docker-compose.yml
```diff
@@ -0,0 +1,38 @@
...
+    volumes:
+      - /var/run/docker.sock:/var/run/docker.sock
+    depends_on:
+      - zookeeper
+
+  tronmaster:
+    build:
+      context: ../
+      dockerfile: ./yelp_package/itest_dockerfiles/tronmaster/Dockerfile
+    command: bash -c 'pip3 install -e /work && exec trond --debug -c /work/cluster_itests/tronfig/   -l /work/example-cluster/logging.conf'
```
keymone: just to be consistent with example-cluster, can you rename config folder to `cluster_itests/config`?

solarkennedy:
Amazingly, this thing now verifies that tron can connect to mesos and asserts that "tron" is in the framework list!

Now we have a place to add more stuff for the future.

PR #495, opened 2018-08-14
save task id of action run

chlgit:
This is useful to recover jobs running on Mesos.

### DIFF
--- a/tron/core/actionrun.py
+++ b/tron/core/actionrun.py
@@ -203,6 +203,7 @@ class ActionRun(object):
         env=None,
         extra_volumes=None,
         mesos_address=None,
+        task_id=None,
     ):
         self.job_run_id = maybe_decode(job_run_id)
         self.action_name = maybe_decode(name)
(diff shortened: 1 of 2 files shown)


tron/core/actionrun.py
```diff
@@ -681,6 +684,8 @@ def _kill_mesos_task(self):
             mesos_cluster = MesosClusterRepository.get_cluster(
                 self.mesos_address
             )
+            if self.task_id is None:
+                raise AttributeError
```
solarkennedy: Hehe. Please do a bit more. 
Actually, are you sure it should error hard like this? Maybe there won't be a task id sometimes. maybe it should just return an error string that there was no task id to kill?

solarkennedy:
Can you call it `mesos_task_id` ?

tron/core/actionrun.py
chlgit: Yeah, you're right. Return an error string makes more sense. I'll change that.

PR #496, opened 2018-08-15
RFC: Make tronview output the action graph to the terminal when available

solarkennedy:
Sorry, I spent a couple hours to scratch and itch instead of doing "real work".

Someone who knows more matplotlib can probably make the output prettier.

But isn't this cool?
Might be even more useful once we get cross-job dependencies.

screen shot 2018-08-14 at 7 02 35 pm

### DIFF
--- a/bin/tronview
+++ b/bin/tronview
@@ -105,7 +105,11 @@ def view_job(args, job_id, client):
     if args.show_events:
         return display_events(client.object_events(job_id.url))
 
-    job_content = client.job(job_id.url, count=args.num_displays)
+    job_content = client.job(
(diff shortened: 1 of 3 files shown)


drmorr0:
If you want to experiment with automagically nicer-looking visualizations, I've heard good things about [holoviews](http://holoviews.org/index.html).  I haven't played around with it much personally yet, but it's on my list of things I'd like to do.

solarkennedy:
@keymone you have to pass that env var through your ssh config. Search your email for TERM_PROGRAM and you see an email from me. Also you need tmux2.

PR #497, opened 2018-08-15
Event bus with pub/sub and persistance

keymone:
- persistance is triggered by time and/or number of stored events since last save
- events are dicts, indexed in the log by `id` key
- subscription is possible by id prefix


tron/eventbus.py
```diff
@@ -0,0 +1,162 @@
...
+        log.debug(f"subscription ({prefix}, {subscriber}) enqueued")
+
+    def has_event(self, event_id):
+        return event_id in self.event_log
+
+    def sync_load_log(self):
+        started = time.time()
+        with open(self.log_current, 'rb') as f:
+            self.event_log = pickle.load(f)
+        duration = time.time() - started
```
chlgit: Do we need to calculate duration every time? Do you concern the pickle performance here?

tron/trondaemon.py
```diff
@@ -180,10 +179,18 @@ def run(self):
             self._run_www_api()
             self._run_manhole()
             self._run_reactor()
+            self._run_eventbus()
+
+    def _run_eventbus(self):
+        self.eventbus = make_eventbus(f"{self.options.working_dir}/_events")
+        # self.eventbus.start()
```
chlgit: ?

tron/eventbus.py
keymone: saving is going to happen relatively rarely, i think we can spare a few cycles to track it's performance, just to be aware of it

tron/trondaemon.py
keymone: yeah, don't want to actually run it when this PR is merged. i'll add relevant args / configuration options separately.

tron/eventbus.py
```diff
@@ -0,0 +1,162 @@
...
+        log.info(f"log read from disk, took {duration:.4}s")
+
+    def sync_save_log(self, reason):
+        started = time.time()
+        new_file = os.path.join(self.log_dir, f"{int(started)}.pickle")
+        with open(new_file, 'xb') as f:
+            pickle.dump(self.event_log, f)
```
solarkennedy: Do we have to pickle? This decision could hurt us in the long run. Sure we can't get away with good ole json?

tron/eventbus.py
```diff
@@ -0,0 +1,162 @@
...
+        if new_subs:
+            self.event_subscribers[prefix] = new_subs
+        else:
+            del self.event_subscribers[prefix]
+        log.debug(f"subscription removed: {prefix} / {sub}")
+
+    def sync_notify(self, event_id):
+        event = self.event_log[event_id]
+        for prefix, subscribers in self.event_subscribers.items():
+            if event_id.startswith(prefix):
```
solarkennedy: This startswith seems funny to me. Does it need to be that loose?

tron/eventbus.py
keymone: json is orders of magnitude slower. i think we will be fine - we never rewrite these files, we store all history of backups (i'll also ship a cron that will implement some sensible retention policy) and most importantly - what goes into these files are dumb dicts of strings to strings (not intended to change). this is as safe as you can be with pickle.

tron/eventbus.py
keymone: it doesn't need to be, but why not?

tron/eventbus.py
```diff
@@ -0,0 +1,162 @@
...
+        self.log_save_interval = 60
+        self.log_save_updates = 10
+
+    def start(self):
+        log.info("starting")
+        self.setup_log_dir()
+        self.sync_load_log()
+        reactor.callLater(0, self.sync_loop)
+
+    def setup_log_dir(self):
```
Rob-Johnson: im not convinced this is the right place for this - can you just pass a file path into the constructor and expect the caller to deal with this?

tests/eventbus_test.py
```diff
@@ -0,0 +1,208 @@
...
+        self.eventbus.shutdown()
+        assert self.eventbus.must_shutdown
+
+    def test_publish(self):
+        evt = {'id': 'foo'}
+        self.eventbus.publish(evt)
+        assert self.eventbus.publish_queue.pop() is evt
+
+    def test_subscribe(self):
+        ps = ('foo', 'bar', 'cb')
```
Rob-Johnson: shouldn't the callback be a function? otherwise I think you'll have to do something [here](https://github.com/Yelp/Tron/pull/497/files#diff-bbce3bb8c085b14e354c7b20df758cbdR162) to convert a string to a function?

tron/eventbus.py
```diff
@@ -0,0 +1,162 @@
...
+            del self.event_subscribers[prefix]
+        log.debug(f"subscription removed: {prefix} / {sub}")
+
+    def sync_notify(self, event_id):
+        event = self.event_log[event_id]
+        for prefix, subscribers in self.event_subscribers.items():
+            if event_id.startswith(prefix):
+                for (sub, cb) in subscribers:
+                    log.debug(f"notifying {sub} about {event_id}")
+                    reactor.callLater(cb, event)
```
Rob-Johnson: is cb a string or a function here?

tests/eventbus_test.py
keymone: yep, it's a function. this is just a test to check we're getting the right thing on event queue, we don't really care about the values.

tron/eventbus.py
keymone: it's a function

tron/eventbus.py
```diff
@@ -0,0 +1,148 @@
...
+        event_id = event['id']
+        if event_id in self.event_log:
+            if self.event_log[event_id] != event:
+                log.info(f"replacing event: {event_id}")
+            else:
+                log.debug(f"duplicate event: {event}")
+                return
+
+        self.event_log[event['id']] = event
+        self.log_save_updates += 1
```
qui: log_updates?

tron/eventbus.py
```diff
@@ -0,0 +1,148 @@
...
+        started = time.time()
+        new_file = os.path.join(self.log_dir, f"{int(started)}.pickle")
+        with open(new_file, 'xb') as f:
+            pickle.dump(self.event_log, f)
+
+        tmplink = os.path.join(self.log_dir, "tmp")
+        os.symlink(new_file, tmplink)
+        os.rename(tmplink, self.log_current)
+
+        self.log_last_save = time.time()
```
qui: Should also reset log_updates here?

tron/eventbus.py
```diff
@@ -0,0 +1,148 @@
...
+            del self.event_subscribers[prefix]
+        log.debug(f"subscription removed: {prefix} / {sub}")
+
+    def sync_notify(self, event_id):
+        event = self.event_log[event_id]
+        for prefix, subscribers in self.event_subscribers.items():
+            if event_id.startswith(prefix):
+                for (sub, cb) in subscribers:
+                    log.debug(f"notifying {sub} about {event_id}")
+                    reactor.callLater(cb, event)
```
qui: Do you need a delay arg?

PR #500, opened 2018-08-16
Made tronview and tronctl provide suggestions on unknown identifiers

solarkennedy:
I felt silly when debugging somehing when I realized my flakey  key was to blame.

This adds suggestions instead of throwing a valueerror exception at our users.

I don't feel super great about copy/pasting code from paasta, but meh.


tron/commands/cmd_utils.py
```diff
@@ -194,3 +198,13 @@ def setup_logging(options):
         format='%(name)s %(levelname)s %(message)s',
         stream=sys.stdout,
     )
+
+
+def suggest_possibilities(word, possibilities, max_suggestions=3):
```
chlgit: Could we increase the value of max_suggestions a little bit, like 10? I feel we have many similar names at Tronfig files.

PR #502, opened 2018-08-16
[WIP] move mesos address to option section

chlgit:
I put MASTER.yaml as an example. This would be removed that before merging to master. I've tested locally and it works fine. If we feel it's the direction we gonna go, I'll update the unit tests.


solarkennedy:
Oh, so `MesosCluster` instead of `MesosClusterRepository`, yea I think that is ok. I've never used the singleton class pattern like this before (Repository was the same though).


I'm pretty sure that this is worth it and that we won't want to 'go back' tron being a "multi-framework" thingy. (making prod the special snowflake, sounds super error prone)


But yea! I hope that by doing this early we are going to make it easier to develop this functionality. Also YAGNI?

dev/config/MASTER.yaml
```diff
@@ -1,15 +1,50 @@
...
 nodes:
   - hostname: localhost
 
+mesos_options:
+  master_address: mesos.paasta-mesosstage.yelp
+  master_port: 5050
+  secret: ''
+  role: '*'
+  enabled: True
+  default_volumes:
```
qui: Let's take out the default volumes unless we need them, I just added these a while back when I was testing the default volumes feature. (Same with the mrjob one below in the action.) It could cause errors if those paths don't exist on the agent.

tron/config/schema.py
```diff
@@ -110,6 +110,7 @@ def config_object_factory(name, required=None, optional=None):
 ConfigMesos = config_object_factory(
     name='ConfigMesos',
     optional=[
+        'master_address',
```
qui: I think this should be required, especially if we're going to be initializing Mesos on start.

tron/config/config_parse.py
```diff
@@ -593,6 +587,7 @@ class ValidateMesos(Validator):
     config_class = ConfigMesos
     option = True
     defaults = {
+        'master_address': None,
```
qui: We should have a validator defined for this below.

tron/config/schema.py
chlgit: hmmm...I am not sure. It makes sense for me if "master_address = None" and "enabled = False."

tron/config/config_parse.py
chlgit: Ah, right.

PR #503, opened 2018-08-16
[WIP] Put the paasta command equivalent into an actionrun output

solarkennedy:
I want users to be able to run their (paasta) action commands outside of tron if they need to. I thought it would be nice to provide them with a command they can copy/paste edit.

However, because of the way we have designed this (strong separation of concerns) this actionrun object doesn't "know" anything about the service/instance/cluster, heck I don't even think it knows what job it is part of?

Any suggestions on what to do here? Fundamentally I think we want to keep "tron" from "knowing" too much about paasta, but for this (nice to have) feature... it kinda has to know.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,7 @@ debian/tron.postinst.debhelper
 debian/tron.postrm.debhelper
 debian/tron.prerm.debhelper
 debian/debhelper-build-stamp
+
+# Dev pg stuff
+dev/tron_state
+dev/tron.pid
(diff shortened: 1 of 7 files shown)


qui:
The job name should be available in the actionrun's context, as `self.context.name` (https://tron.readthedocs.io/en/latest/command_context.html).

But yeah, given the way we decided to separate PaaSTA and Tron, I don't think we should do this. I vote for adding this feature to the paasta CLI.

keymone:
i agree this should be part of paasta CLI. it's paasta that generates tron jobs and actions, so it's paasta that can render a command for a user.

solarkennedy:
Thank you for the input @qui and @keymone. I think you are right, I'm going to discard.

PR #504, opened 2018-08-17
Remove event.py and related code

keymone:
As I started working on eventbus to support cross-job dependencies I noticed that we have a lot of code that's doing similar stuff. `event.py` is more complex though and currently is only used to log of what happens to jobruns and  for crash reporting configured by `notification_options`.

We're not using `notification_options` and it's way more efficient (and less code) to just use `logging` module for logging.

We will lose `/events` api endpoint and related tronview/tronweb functionality. I still need to cleanup tronweb, but want to get some early opinions just in case  this is a no-go.


solarkennedy:
I'm ok with removing email / tronweb etc functionality.

But you know what would be cool: wouldn't it be neat if we could make all the run data for a paasta service show up in paasta logs? With the current design I don't know how we would do that, but it seems like this sort of event architecture would be the way to go. Are you sure the plain old logging module could do what I want?

chlgit:
Totally ok with removing these. The simple the better, yay. To stand on the safe side, could you remove the notification_options section from MASTER.yaml first? If we forgot to remove that before making a new release including this, it would block Tron deployment.

keymone:
@solarkennedy i think we can make it work. logging module is quite sophisticated and we already have non-trivial config for it: https://github.com/Yelp/Tron/blob/master/tron/logging.conf

problem is tron jobs don't carry paasta information (and shouldn't if we want to keep tron agnostic), so there's no easy way right now to tell which service/instance is running.

we could make each job/action have a generic `log_qualifiers` dict that would be attached to log stream and on the other end of that stream something paasta-y would parse those qualifiers and understand them (`service=X instance=Y` or w/e). best i can think of atm to have least coupling with least effort.

keymone:
@chlgit yes, valid point. i actually really don't like this aspect of our configs. we shouldn't really be erroring on having data we don't understand. possibility of uncaught typos (99% of which can be caught with trivial levenshtein distance check) is smaller issue than inability to extend the system easily.

Rob-Johnson:
I think this information could be useful, but I agree that the maintenance isn't worth it right now. I also agree that it highlights that 'event' is pretty overloaded in your other PR, which isn't really a generic event bus at all (though wouldn't that be nice)

PR #506, opened 2018-08-17
Added mesos framework authentication to itests

solarkennedy:
This fails! Here is the traceback:
https://pb.yelpcorp.com/322552

I need some help interpreting this. My best guess is that prints are unsafe in this loop somehow?
The symptom is that you run `tronctl start` and it hangs for a few seconds, and then trond crashes.


qui:
Not sure how this leads to the stack trace, but the principal is incorrect. We didn't make principal an option in https://github.com/Yelp/Tron/pull/492, and taskproc uses 'taskproc' by default.

cc @chlgit

chlgit:
Ah, it might be the root cause. But I have difficulty to link the principal and the stack trace together...

solarkennedy:
Update: added principal, logging, and fixed the secret vs secret_file issue.

PR #507, opened 2018-08-18
Added the prototype of the tronctl backfill command

solarkennedy:
Feedback from the CEP stuff, this apparently is something users are doing themselves a lot.
This is the very first pass of this feature, future version will actually run stuff, kick off dependent jobs, etc.

### DIFF
--- a/bin/tronctl
+++ b/bin/tronctl
@@ -26,10 +26,9 @@ COMMAND_HELP = (
     ('rerun', 'Re-run a full job (all actions) with a new job id'),
     ('retry', 'Re-run a job action within an existing job run'),
     ('cancel', 'Cancel the selected job run'),
+    ('backfill', 'Start many jobs for a particular date range'),
     ('disable', 'Disable selected job and cancel any outstanding runs'),
     ('enable', 'Enable the selected job and schedule the next run'),
-    ('disableall', 'Disable all jobs'),
-    ('enableall', 'Enable all jobs'),
     ('fail', 'Mark an UNKNOWN job as having failed'),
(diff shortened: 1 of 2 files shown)


solarkennedy:
Pretty sure new runs. @shazeline when you or others do these mega-backfills, it is with new runs (my guess from the code you showed me)

bin/tronctl
```diff
@@ -110,17 +121,42 @@ def control_objects(args):
...
-
         yield edit(args, tron_id.url)
 
 
+def backfill(args):
+    if args.start_date is None:
+        print("Error: For a backfill, --start-date must be set")
+        return 1
+    if args.end_date is None:
+        args.end_date = datetime.datetime.today()
```
chlgit: Will this be a little too dangerous? If someone forgot to set end_date, this would backfill more than he/she wants.

solarkennedy: I think it is a sane default, plus with the manual verification before you run something, we are always giving users a second chance to review what is about to happen.

chlgit: Ok, fair enough.

bin/tronctl
```diff
@@ -62,7 +61,19 @@ def parse_cli():
         "--run-date",
         type=parse_date,
         dest="run_date",
-        help="For job starts, what should run date be set to",
+        help="For job starts, what should-run date be set to",
```
qui: hyphen should be in run-date :)

solarkennedy:
Yea with `allow_overlap: false` they will be queued unless the user has set `queuing: false`.

PR #508, opened 2018-08-20
Nix testify

solarkennedy:
I was tired of writing Testify stuff.

* Most of this was done via the migration script, sed, and some outliers by hand
* testifycompat is a small library that we have internally. I'm just vendoring it because I think this is "fine"
* Sandbox stuff is skipped. I would be up for deleting it if you think I should.
* I had trouble with one of the fixtures, I didn't spend "that" much time on it and in the end marked it as skipped.
* Although I skipped the sandbox test cases, the integration and sandbox suites are now technically "running", so that is good?


solarkennedy:
I agree it would be cool. Yea another time.
Maybe the next time someone does something with tronctl they can remove the equivalent sandbox test and enable a matching behave test.

PR #509, opened 2018-08-21
move mesos address to mesos option section only

chlgit:


dev/config/MASTER.yaml
```diff
@@ -1,15 +1,46 @@
...
       actions:
         - name: "first"
-          command: "echo hi"
+          command: "/bin/sleep 5m"
+          executor: mesos
+          docker_image: docker-paasta.yelpcorp.com:443/services-robj-test-service:paasta-e9fc795ec881cfd1b9d4b68540504690deddd4ec
+          cpus: 1
+          mem: 100
+          extra_volumes:
+              - container_path: '/nail/etc/mrjob'
```
qui: Can we take this out too?

tron/config/schema.py
```diff
@@ -110,6 +110,7 @@ def config_object_factory(name, required=None, optional=None):
 ConfigMesos = config_object_factory(
     name='ConfigMesos',
     optional=[
+        'master_address',
```
qui: OK, can we either 1) make this required or 2) add validation that this exists if enabled is True?

tron/mesos.py
```diff
@@ -38,6 +38,7 @@ class MesosClusterRepository:
     """A class that stores MesosCluster objects and configuration."""
 
     clusters = {}
+    mesos_master_address = None,
```
qui: nit: don't need these commas

tron/config/schema.py
```diff
@@ -109,6 +109,9 @@ def config_object_factory(name, required=None, optional=None):
 
 ConfigMesos = config_object_factory(
     name='ConfigMesos',
+    required=[
+        'master_address',
```
solarkennedy: Does this mean you will have to add `mesos_address` to all our master files when we upgrade?

chlgit: Yes. I will update the master files later.

PR #511, opened 2018-08-23
Add eventbus and cross-job dep related config attributes

keymone:
Global: `eventbus_enabled`, per action/actionrun: `triggers_downstream`, `triggered_by`, `on_upstream_rerun`.

All default to `None`.

### DIFF
--- a/tests/eventbus_test.py
+++ b/tests/eventbus_test.py
@@ -47,7 +47,7 @@ class EventBusTestCase(TestCase):
     def teardown(self):
         self.log_dir.cleanup()
 
-    @mock.patch('tron.eventbus.reactor', autospec=None)
+    @mock.patch('tron.eventbus.reactor', autospec=True)
     def test_start(self, reactor):
         self.eventbus.sync_load_log = mock.Mock()
         reactor.callLater = mock.Mock()
@@ -75,7 +75,7 @@ class EventBusTestCase(TestCase):
         self.eventbus.event_log['foo'] = 'bar'
         assert self.eventbus.has_event('foo')
 
-    @mock.patch('tron.eventbus.time', autospec=None)
+    @mock.patch('tron.eventbus.time', autospec=True)
     def test_sync_load_log(self, time):
         time.time = mock.Mock(return_value=1.0)
         self.eventbus.event_log = {'foo': 'bar'}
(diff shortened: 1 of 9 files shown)


tron/config/config_parse.py
```diff
@@ -372,6 +375,12 @@ class ValidateAction(Validator):
             valid_dict,
         'extra_volumes':
             build_list_of_type_validator(valid_volume, allow_empty=True),
+        'triggers_downstream':
+            valid_bool,
+        'triggered_by':
+            valid_dict,
```
Rob-Johnson: shouldn't this be a list?

keymone: no, because we need to refer to triggers by both name and value to avoid situations where different triggers generate the same value

tests/eventbus_test.py
```diff
@@ -48,7 +75,9 @@ def test_has_event(self):
         self.eventbus.event_log['foo'] = 'bar'
         assert self.eventbus.has_event('foo')
 
-    def test_sync_load_log(self):
+    @mock.patch('tron.eventbus.time', autospec=None)
```
chlgit: why not autospec=True?

tron/core/actionrun.py
```diff
@@ -238,6 +244,9 @@ def __init__(
 
         self.action_command = None
         self.in_delay = None
+        self.triggers_downstream = None
```
chlgit: Should it be self.triggers_downstream = triggers_downstream?

tests/eventbus_test.py
keymone: this is actually cargo-culted, i don't even know what autospec does and why we need it :) i didn't notice any difference between having autospec=True vs autospec=None.

tron/config/config_parse.py
```diff
@@ -648,6 +666,7 @@ class ValidateConfig(Validator):
         'node_pools': {},
         'jobs': (),
         'mesos_options': ConfigMesos(**ValidateMesos.defaults),
+        'eventbus_enabled': None,
```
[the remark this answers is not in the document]
keymone: for an optional attribute i think it's better to default to `None` as it reflects absence of value better

tests/eventbus_test.py
chlgit: I just spent a couple of minutes to read the history of cargo cult. TIL

qui: autospec tries to prevent you from making method calls, etc that don't exist on the object you're mocking, it's usually a good idea :)

tron/config/config_parse.py
qui: What are the name and value in this case? I thought triggered_by would be a list of items like `MASTER.job.action.shortdate.%(shortdate-1)s`.

keymone: my bad, i mixed up the attributes. `triggered_by` should be a list, `triggers_downstream` should be bool/dict. thanks 👍

PR #512, opened 2018-08-23
Recover unknown mesos actions on startup

qui:
Goes with https://github.com/Yelp/task_processing/pull/120, works on restart!

Need to test more edge cases and write tests, but wanted to get the overall idea out there for feedback.

### DIFF
--- a/tron/core/actionrun.py
+++ b/tron/core/actionrun.py
@@ -690,7 +690,7 @@ class MesosActionRun(ActionRun, Observer):
             return
 
         self.watch(task)
-        mesos_cluster.submit(task, recover=True)
+        mesos_cluster.recover(task)
 
         # Reset status
         self.exit_status = None
--- a/tron/core/actionrun.py
+++ b/tron/core/actionrun.py
@@ -653,6 +653,52 @@ class MesosActionRun(ActionRun, Observer):
         mesos_cluster.submit(task)
         return task
 
+    def recover(self):
+        if self.mesos_task_id is None:
+            log.error(f'No task ID, cannot recover {self}')
+            return
+
+        if not self.machine.check('running'):
+            log.error(
+                f'Unable to transition {self} from {self.machine.state}'
+                'to running for recovery'
+            )
+            return
+
+        log.info(f'Recovering Mesos run {self}')
+
+        serializer = filehandler.OutputStreamSerializer(self.output_path)
+        mesos_cluster = MesosClusterRepository.get_cluster()
+        task = mesos_cluster.create_task(
+            action_run_id=self.id,
+            command=self.command,
+            cpus=self.cpus,
+            mem=self.mem,
+            constraints=self.constraints,
+            docker_image=self.docker_image,
+            docker_parameters=self.docker_parameters,
+            env=self.env,
+            extra_volumes=self.extra_volumes,
+            serializer=serializer,
+            task_id=self.mesos_task_id,
+        )
+        if not task:
+            log.warning(
+                f'Cannot recover {self}, Mesos is disabled or invalid task ID'
+            )
+            self.fail_unknown()
+            return
+
+        self.watch(task)
+        mesos_cluster.submit(task, recover=True)
+
+        # Reset status
+        self.exit_status = None
+        self.end_time = None
+        self.machine.transition('running')
+
+        return task
+
     def stop(self):
         if self.retries_remaining is not None:
             self.retries_remaining = -1
--- a/tron/core/recovery.py
+++ b/tron/core/recovery.py
@@ -18,10 +18,9 @@ def filter_action_runs_needing_recovery(action_runs):
     ]
 
 
-def filter_recoverable_action_runs(action_runs):
+def group_by_actionrun_type(action_runs):
     """
-    Given a list of action_runs, return those that can be recovered.
-    SSHActionRuns and MesosActionRuns are returned separately.
+    Given a list of action_runs, group them by type.
     """
     ssh_runs = []
     mesos_runs = []
@@ -34,7 +33,7 @@ def filter_recoverable_action_runs(action_runs):
 
 
 def filter_recovery_candidates(runs):
-    return filter_recoverable_action_runs(
+    return group_by_actionrun_type(
         action_runs=filter_action_runs_needing_recovery(action_runs=runs),
     )
 
--- a/tron/core/recovery.py
+++ b/tron/core/recovery.py
@@ -5,6 +5,7 @@ import logging
 
 from tron.actioncommand import NoActionRunnerFactory
 from tron.core.actionrun import ActionRun
+from tron.core.actionrun import MesosActionRun
 from tron.core.actionrun import SSHActionRun
 
 log = logging.getLogger(__name__)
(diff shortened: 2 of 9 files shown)


tron/mesos.py
```diff
@@ -309,13 +311,16 @@ def handle_next_event(self, deferred_result=None):
...
-    def submit(self, task):
+    def submit(self, task, recover=False):
         if not task:
             return
 
         if not self.enabled:
-            task.log.info('Task failed to start, Mesos is disabled.')
-            task.exited(1)
+            if recover:
+                task.log.info('Could not recover task, Mesos is disabled.')
```
solarkennedy: I guess we don't need to belabor the point by logging lots of advice, but can you confirm that we will be "unknown" in this case? (we should)
If we are disabling mesos for some reason, and there is a taskid, then it might be still running but who knows.

It just says "return" on line 324, does that mean it won't move the state machine at all and leave it `unknown`?

tron/core/actionrun.py
```diff
@@ -653,6 +653,47 @@ def submit_command(self):
         mesos_cluster.submit(task)
         return task
 
+    def recover(self):
+        if self.mesos_task_id is None:
+            log.error(f'No task ID, cannot recover {self}')
+            return
+
+        okay_to_recover = self.machine.transition('running')
```
solarkennedy: I don't really understand this. Isn't it premature to transition the machine to `running` before we have even talked to mesos?

Shouldn't we talk to mesos first and make sure it is in `TASK_RUNNING` before moving the state machine?

tron/core/actionrun.py
```diff
@@ -653,6 +653,47 @@ def submit_command(self):
         mesos_cluster.submit(task)
         return task
 
+    def recover(self):
+        if self.mesos_task_id is None:
+            log.error(f'No task ID, cannot recover {self}')
```
solarkennedy: I think we never get here because of recovery.py:32. (due to `None` filtering)
What do you think about removing the `None` filter so that this code does get invoked and logs the "why"?

solarkennedy:
I really want the itest framework to be able to restart tron for us so we can "prove" that this works and that it will continue to work as we improve things.

I can't think of a super great way to do that, but I don't want it to block you.

tron/mesos.py
qui: Good catch! Yes, the return means that the state machine isn't moved, and it was my intention for it be be unknown. But since it's transitioned to running in `tron/core/actionrun`, there's a chance it could be left in running. Will update.

tron/core/actionrun.py
qui: Yeah, that makes sense. Will do.

tron/core/actionrun.py
solarkennedy: But we are processing all these runs serially anyway, the actual transition happens in this same function.

How about running `check('running')` instead?

The thing I'm worried about is all the things that could go wrong later in the function, and I think that moving it right to "running" as the first thing we do in this function is presumptuous. We haven't even talked to mesos yet?

solarkennedy: For monitoring concerns, what if the script just checked the uptime of tron first? If it is <2m say, just skip out and give it time to settle in before checking jobs. (will be nice to go easy on the api anyway to give it some breathing room after a restart)

tron/mesos.py
```diff
@@ -327,14 +333,19 @@ def submit(self, task):
...
-            'Submitting task {} to {}'.format(
-                mesos_task_id,
-                self.mesos_address,
-            ),
-        )
+        self.tasks[mesos_task_id] = task  # handle case if task id already exists?
+        if recover:
+            task.log.info('Reconciling state for this task from Mesos')
+            task.started()
+            self.runner.check(task.get_config())
```
keymone: since  our task ids are unique, can we treat `.run` as idempotent somehow and just call it again with the same task config? this way we'd avoid needing `.check` at all.

tron/core/actionrun.py
```diff
@@ -653,6 +653,47 @@ def submit_command(self):
...
+            )
+            return
+
+        log.info(f'Recovering Mesos run {self}')
+        self.exit_status = None
+        self.end_time = None
+
+        serializer = filehandler.OutputStreamSerializer(self.output_path)
+        mesos_cluster = MesosClusterRepository.get_cluster()
+        task = mesos_cluster.create_task(
```
chlgit: Are we sure that mesos cluster would be able to reregister the framework at this point? Do we need to worry about the race condition between reregister and task recovery? Or am I missing any safeguard to prevent this somewhere?

tron/mesos.py
qui: Hm, are you suggesting that we always attempt to launch the task, and if the task ID has been used before, then Mesos will ignore it?

If Mesos doesn't ignore it, then how does taskproc figure out if it's submitted the task to Mesos before, or not?

tron/mesos.py
```diff
@@ -366,6 +378,7 @@ def create_task(
             volumes=volumes,
             uris=uris,
             offer_timeout=self.offer_timeout,
+            task_id_override=task_id,
```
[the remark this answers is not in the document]
qui: Yeah, I was trying to avoid assuming anything about how the task_id is constructed by TaskConfig, in Tron. Maybe I could add a method to TaskConfig like `set_task_id(self, task_id)` so the logic would be contained there?

tron/core/actionrun.py
qui: The .check method just tells task_processing to keep track of the task ID, the actual requests to Mesos don't happen until the framework is ready. So no race condition.

tron/mesos.py
keymone: yeah, ignore me, we don't need to enforce idempotency here.

tron/core/actionrun.py
qui: I moved the transition to the end of the method, so it doesn't transition if something goes wrong before submitting to taskproc.

At that point, I think it's fair to be optimistic and say it's running - we do the same for our SSH recovery. It helps with monitoring, and also matters if the job has `allow_overlap: False` (unknown wouldn't prevent another run from starting).

tron/mesos.py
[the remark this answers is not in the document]
qui: Updated to pass in name and uuid.

tron/core/recovery.py
```diff
@@ -19,13 +20,17 @@ def filter_action_runs_needing_recovery(action_runs):
 
 def filter_recoverable_action_runs(action_runs):
     """
-    Given a list of action_runs, create a filtered list that only includes those that can be recovered
-    For now, the only test is whether the run is an SSHActionRun
+    Given a list of action_runs, return those that can be recovered.
+    SSHActionRuns and MesosActionRuns are returned separately.
```
Rob-Johnson: if you're going to change what this function does, can you rename it? this doesn't really do any filtering anymore. just ``group_by_actionrun_type`` or something?

tron/mesos.py
```diff
@@ -328,13 +334,18 @@ def submit(self, task):
...
         mesos_task_id = task.get_mesos_id()
         self.tasks[mesos_task_id] = task
-        self.runner.run(task.get_config())
-        log.info(
-            'Submitting task {} to {}'.format(
-                mesos_task_id,
-                self.mesos_address,
-            ),
-        )
+        if recover:
```
Rob-Johnson: what do you think about moving this into a separate ``recover`` function on the task, rather than overloading this function?

tron/core/recovery.py
qui: Sure, done.

tron/mesos.py
qui: Also done.

tron/core/recovery.py
```diff
@@ -17,19 +18,22 @@ def filter_action_runs_needing_recovery(action_runs):
...
+    mesos_runs = []
+    for action_run in action_runs:
+        if isinstance(action_run, SSHActionRun):
+            ssh_runs.append(action_run)
+        elif isinstance(action_run, MesosActionRun):
+            mesos_runs.append(action_run)
+    return ssh_runs, mesos_runs
 
 
 def filter_recovery_candidates(runs):
```
Rob-Johnson: I'd probably just remove this function entirely now

PR #514, opened 2018-08-28
Yapfify

keymone:
i'll do this from time to time until we re-enable pre-commit hook


Rob-Johnson:
can you remind me why we disabled the pre-commit hook?

keymone:
we run them on python ~3.6.1 and yapf fails to validate f-strings on that version.

PR #515, opened 2018-08-28
Eventbus sync shutdown

keymone:
same logic as in state manager. when sighup comes - we don't have time to wait for next loop iteration and handle shutdown there.

### DIFF
--- a/tron/eventbus.py
+++ b/tron/eventbus.py
@@ -97,7 +97,8 @@ class EventBus:
         try:
             self.sync_process()
         except Exception as e:
-            log.error(e)
+            raise SystemExit(f"eventbus loop crashed with {e}")
+
         reactor.callLater(1, self.sync_loop)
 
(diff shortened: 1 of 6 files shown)


tron/eventbus.py
```diff
@@ -88,20 +91,28 @@ def sync_save_log(self, reason):
...
-            log.info("shutdown completed")
-        else:
+        if not self.enabled:
+            return
+
+        try:
             self.sync_process()
-            reactor.callLater(1, self.sync_loop)
+        except Exception as e:
+            log.error(e)
```
Rob-Johnson: think this could be dangerous? I'd hate for us to not notice these flying by - maybe we should exit if this fails?

keymone: i was considering crashing the process here but decided to just log the error because no essential functionality is affected, but it's not a strong opinion, i'll change it.

PR #518, opened 2018-09-05
Deprecate --nodaemon option

keymone:
We're only using NoDaemonContext anyway, also:

- move debug handler to MCP class where signal loop lives
- keep manhole.sock path in MCP's attribute
- set pthread mask on every iteration in signal loop
- more logging

### DIFF
--- a/bin/trond
+++ b/bin/trond
@@ -51,7 +51,7 @@ def parse_cli():
         "--nodaemon",
         action="store_true",
         default=False,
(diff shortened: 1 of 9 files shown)


Rob-Johnson:
instead of all the deprecation flags, can we just get rid of them?

keymone:
@Rob-Johnson that might break next rollout since our custom upstart script has that flag

PR #519, opened 2018-09-05
Remove enableall and disableall from JobCollections in tron controller

kawaiwanyelp:
### Description:
- @solarkennedy previously removed enableall/disableall from tronctl when he added the backfill command, but there was some related code in the api controller that was left behind, which I have now also removed.
- Note that the JobCollectionController can no longer handle any commands.
- The controller's tests have been updated to check that any command received will raise an UnknownCommandError.

### Testing done:
- pytest on the updated controller tests

### DIFF
--- a/tests/api/controller_test.py
+++ b/tests/api/controller_test.py
@@ -13,6 +13,7 @@ from tron import mcp
 from tron.api import controller
 from tron.api.controller import ConfigController
 from tron.api.controller import JobCollectionController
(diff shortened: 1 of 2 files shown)


qui:
@kawaiwanyelp The build is failing, make sure you have pre-commit installed!

kawaiwanyelp:
Fixed the precommit; I added an extra line by accident.

PR #521, opened 2018-09-05
Removed tronfig header code, as part of removing autogenerated comments

kawaiwanyelp:
### Description
- Blanked the config template except for the header terminator. 
- Modified the render_template function so that it does not do any string replacement to fill the header, as it is no longer around.
- Modified render_template tests to match.

### Testing done
- make test

### Note to reviewers
- The functionality to include or not include headers is already present in the API. Clients simply need to set a no_header boolean in the request to the API to true and the header/autogenerated comments will not appear. 
- I was going to remove all the header code, but I wanted to get @Rob-Johnson's feedback on whether or not the header code could be useful going forward.


Rob-Johnson:
hm, my personal experience is that it has just ended up as noise and always seems to fall out of date, so my vote is to get rid of it.

we certainly won't have such a thing in the new tron-on-paasta world.

kawaiwanyelp:
I removed all header code and updated the tests to match as per @Rob-Johnson's suggestion. Should be good to go!

tron/api/resource.py
```diff
@@ -321,17 +321,13 @@ def get_config_index(self):
     @AsyncResource.bounded
     def render_GET(self, request):
         config_name = requestargs.get_string(request, 'name')
-        no_header = requestargs.get_bool(request, 'no_header')
```
qui: I think you need to leave this for backwards-compatability and remove it in a later change. paasta_tools currently passes this param: https://github.com/Yelp/paasta/blob/master/paasta_tools/tron/client.py#L89

It can do nothing, but the API needs to accept that param without breaking.

tron/named_config_template.yaml
```diff
@@ -1,22 +1 @@
...
-#
-#
-# Command Context:
-# ---------------
-# %(command_context)s
-#
-#
-# All the lines above will be ignored and removed from your configuration.
-# Configuration starts below this line.
 ###################################END#HEADER###################################
```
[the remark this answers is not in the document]
kawaiwanyelp: Hm, I'm pretty sure I deleted this already. I'll re-delete.

tron/api/resource.py
kawaiwanyelp: Fair.

tron/api/controller.py
```diff
@@ -152,60 +149,27 @@ class ConfigController(object):
...
 
     def __init__(self, mcp):
         self.mcp = mcp
         self.config_manager = mcp.get_config_manager()
 
-    def render_template(self, config_content):
-        container = self.config_manager.load()
-        command_context = container.get_master().command_context or {}
-        context = {
-            'node_names': format_seq(container.get_node_names()),
```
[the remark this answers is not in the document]
kawaiwanyelp: Oops, forgot about those. Ty!

kawaiwanyelp:
Took @qui's suggestions and remove format functions for headers and deleted the template.

tron/api/resource.py
qui: Oops, I think you're right. No need to change, then.

PR #522, opened 2018-09-06
Better feedback on killing Mesos actions if not running

qui:
If the task hasn't started or is already done, the "Can't find task id" message is kind of confusing.


tests/core/actionrun_test.py
```diff
@@ -1058,7 +1060,8 @@ def test_kill_task(self, mock_cluster_repo):
         )
 
     @mock.patch('tron.core.actionrun.MesosClusterRepository', autospec=True)
-    def test_kill_task_not_running(self, mock_cluster_repo):
+    def test_kill_task_no_task_id(self, mock_cluster_repo):
```
chlgit: nit: you might want to add test_kill_task_not_running for better coverage

PR #523, opened 2018-09-06
Fix colors on tronweb

solarkennedy:
These CSS things were not (apparently) associated with events


tronweb/css/tronweb.less
```diff
@@ -792,6 +792,34 @@ div.refresh-view span {
     .success;
 }
 
+.enabled,
+.up,
```
qui: Let's combine this block with the one above.

solarkennedy: good idea

PR #524, opened 2018-09-07
Scheme optional for Mesos master address

qui:
This was failing to connect if the address already had http in it, e.g. 'http://leader.yelp'. 
I wanted to make it more flexible, so that `http` was optional, and it could end in a `/` without breaking things.

The logic turned out to be more complicated than I thought, but the test cases should cover it.

### DIFF
--- a/tron/config/config_parse.py
+++ b/tron/config/config_parse.py
@@ -10,6 +10,7 @@ import getpass
 import itertools
 import logging
 import os
+from urllib.parse import urlparse
 
 import pytz
 import six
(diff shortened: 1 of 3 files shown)


tron/config/config_parse.py
```diff
@@ -161,6 +162,41 @@ def valid_node_name(value, config_context):
...
+
+    if params or query or fragment:
+        msg = f"Mesos master address may not contain path components, got {value}"
+        raise ConfigError(msg)
+
+    # Only one of netloc or path allowed, and no / except trailing ones.
+    # netloc is empty if there's no scheme, then we try the path.
+    path = path.rstrip('/')
+    if (netloc and path) or '/' in path:
+        msg = f"Mesos master address may not contain path components, got {value}"
```
chlgit: nit: you might want to make the msg a little different from the previous one to make debugging easier.

qui: The solution is the same (take everything out except the host address) so I think it's fine. I think they're basically the same problem.

PR #525, opened 2018-09-10
Fix: restarting tron will restore outdated scheduled jobs

keymone:
When jobs are reconfigured and new runs are created, they don't trigger STATE_CHANGED handlers. If tron is restarted with such scheduled jobs, they will not be restored and older runs with stale config may be restored instead.

In this PR:

- eagerly save new scheduled runs
- f-stringify some log messages

### DIFF
--- a/tron/core/job.py
+++ b/tron/core/job.py
@@ -340,7 +340,10 @@ class JobScheduler(Observer):
         # also possible this job was run (or is running) manually by a user.
         # Alternatively, if run_queued is True, this job_run is already queued.
         if not run_queued and not job_run.is_scheduled:
-            log.info(f"Scheduling {job_run} in state {job_run.state}")
+            log.info(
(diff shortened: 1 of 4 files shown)


tron/core/job.py
```diff
@@ -336,20 +332,15 @@ def run_job(self, job_run, run_queued=False):
...
         # also possible this job was run (or is running) manually by a user.
         # Alternatively, if run_queued is True, this job_run is already queued.
         if not run_queued and not job_run.is_scheduled:
-            log.info(
-                "%s in state %s already out of scheduled state." % (
-                    job_run,
-                    job_run.state,
-                )
-            )
+            log.info(f"Scheduling {job_run} in state {job_run.state}")
```
qui: I think this might be misleading, since we're actually scheduling a different job run in the next line? 
How about, `f"{job_run} in state {job_run.state} is not scheduled, scheduling a new run instead of running"` or something?

PR #526, opened 2018-09-10
fix yelpsoa_config make test fail

chlgit:
Yelpsao_config pre-commit doesn't like anything except OK. So, I remove that.

### DIFF
--- a/tron/config/config_parse.py
+++ b/tron/config/config_parse.py
@@ -8,7 +8,6 @@ from __future__ import unicode_literals
 import datetime
 import getpass
 import itertools
-import logging
 import os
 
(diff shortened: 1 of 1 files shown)


solarkennedy:
The pre-commit hook spec says that it only cares about the return-code.

This warning output shouldn't alter the return-code, what is expecting the "OK" line?

chlgit:
@solarkennedy yeah, we do it at here https://github.com/Yelp/Tron/blob/master/bin/tronfig#L194
I will close this one and modify at yelpsoa-config there (https://reviewboard.yelpcorp.com/r/336689/).

PR #527, opened 2018-09-11
Cross-job deps: Actionrun triggers upon completion

keymone:
There are some non-trivial decisions to be made how to connect different pieces using eventbus, in this  PR i've decided to pass `eventbus_publish` function to various constructors and factories of jobs/jobruns/actionruns.

Also did some cleanups.


Rob-Johnson:
so the ``eventbus_publish`` function will look something like ``lambda x: self.eventbus.put(x)`` and pass that down from the mcp?

keymone:
@Rob-Johnson yep, it will also be noop if eventbus is disabled.

tron/core/jobrun.py
```diff
@@ -210,43 +211,45 @@ def _start_action_runs(self):
...
-    def handle_action_run_state_change(self, action_run, _):
+    def handle_action_run_state_change(self, action_run: ActionRun, _):
         """Handle events triggered by JobRuns."""
         # propagate all state changes (from action runs) up to state serializer
         self.notify(self.NOTIFY_STATE_CHANGED)
 
         if not action_run.is_done:
             return
 
+        if not action_run.is_broken and action_run.trigger_downstreams:
```
chlgit: We can put the logic about not action_run.is_broken together, so it would be easier to read.

tron/core/actionrun.py
```diff
@@ -390,10 +398,14 @@ def _done(self, target, exit_status=0):
...
             exit_status,
         )
+        # TODO: - state machine already does .check()
+        #       - only set exit_status / end_time if transition succeeds?
         if self.machine.check(target):
             self.exit_status = exit_status
             self.end_time = timeutils.current_time()
             return self.machine.transition(target)
+        else:
+            log.debug(f"{self} failed transition {self.state} ->")
```
chlgit: -> to where? Do you miss anything in the end?

tron/core/actionrun.py
```diff
@@ -442,7 +454,27 @@ def fail(self, exit_status=0):
 
         return self._done('fail', exit_status)
 
+    def emit_triggers(self):
+        if isinstance(self.trigger_downstreams, bool):
+            shortdate = self.render_template("%(shortdate)s")
```
chlgit: May I recommend {shortdate} rather than %(shortdate)s here? :)

tron/core/actionrun.py
```diff
@@ -442,7 +454,27 @@ def fail(self, exit_status=0):
...
+            triggers = [f"shortdate.{shortdate}"]
+        elif isinstance(self.trigger_downstreams, dict):
+            rendered = [
+                (k, self.render_template(v))
+                for k, v in self.trigger_downstreams.items()
+            ]
+            triggers = [f"{key}.{value}" for key, value in rendered]
+        else:
+            log.error(f"{self} trigger_downstreams must be true or dict")
```
chlgit: Do we need to worry about this error here? trigger_downstreams should be validated through the config parser already, right?

tron/core/job.py
```diff
@@ -446,6 +449,7 @@ def build(self, job_config):
             self.context,
             output_path,
             self.action_runner,
+            eventbus_publish=self.eventbus_publish,
```
chlgit: Since you've used kargs for eventbus_publish, why not using kargs for all of them?

tron/core/jobrun.py
```diff
@@ -228,25 +231,25 @@ def handle_action_run_state_change(self, action_run, _):
...
         if not cleanup_run or cleanup_run.is_done:
             return self.finalize()
 
-        # TODO: remove in (0.6), start() no longer raises an exception
-        # When a job is being disabled, or the daemon is being shut down a bunch
-        # of ActionRuns will be cancelled/failed. This would cause cleanup
-        # action to be triggered more then once. Guard against that.
-        if cleanup_run.check_state('start'):
-            cleanup_run.start()
+        cleanup_run.start()
```
chlgit: No need to do clenup_run.check_state('start') anymore?

tron/core/jobrun.py
```diff
@@ -395,62 +382,44 @@ def get_run_by_index(self, index):
...
-        def func(r):
-            return True if include_manual else not r.manual
-
-        return self._get_run_using(func)
+        return next_or_none(r for r in self.runs if include_manual or not r.manual)
 
     def get_pending(self):
         """Return the job runs that are queued or scheduled."""
-        return self._get_runs_using(lambda r: r.is_scheduled or r.is_queued)
+        return [r for r in self.runs if r.is_scheduled or r.is_queued]
```
chlgit: why not next_or_none(r for r in self.runs if r.is_scheduled or r.is_queued) ?

tron/core/jobrun.py
```diff
@@ -395,62 +382,44 @@ def get_run_by_index(self, index):
...
-
-            def func(r):
-                return (r.is_running or r.is_starting) and r.node == node
-        else:
-
-            def func(r):
-                return r.is_running or r.is_starting
-
-        return self._get_runs_using(func)
+        return [
```
chlgit: Same here. Why not next_or_none?

tron/core/actionrun.py
keymone: we don't *need* to, but it doesn't hurt to be a bit more explicit when something changes in future

tron/core/job.py
keymone: mostly laziness, i'll fix :)

tron/core/jobrun.py
keymone: that's what i  understood from the removed comment just above the `if`:

     # TODO: remove in (0.6), start() no longer raises an exception
    # When a job is being disabled, or the daemon is being shut down a bunch
    # of ActionRuns will be cancelled/failed. This would cause cleanup
    # action to be triggered more then once. Guard against that.

tron/core/jobrun.py
keymone: `next_or_none` is for cases when we only need first element of the sequence, in this  case we need  the sequence.

tron/mcp.py
```diff
@@ -115,8 +115,13 @@ def build_job_scheduler_factory(self, master_config):
             output_stream_dir,
             master_config.time_zone,
             action_runner,
+            self.eventbus_publish,
         )
 
+    def eventbus_publish(self, message):
```
Rob-Johnson: just to repeat what we spoke about offline:

I wonder if it might make for a cleaner interface if, rather than sending this callback down the stack, we expose a Queue object that the eventbus is on the other end of. It means you can cleanly seperate the consumer from the the publisher (so you can replace the eventbus object at will in the mcp), and I'd argue that it's less complex than dealing with callbacks as you've done here. 

The downside is that you'd have to create a NoopEventBus or something to consume off the queue and do nothing with the events to avoid the queue filling up.

tron/core/actionrun.py
```diff
@@ -442,7 +454,27 @@ def fail(self, exit_status=0):
 
         return self._done('fail', exit_status)
 
+    def emit_triggers(self):
+        if isinstance(self.trigger_downstreams, bool):
+            shortdate = self.render_template("{shortdate}")
+            triggers = [f"shortdate.{shortdate}"]
+        elif isinstance(self.trigger_downstreams, dict):
+            rendered = [
+                (k, self.render_template(v))
```
qui: We should have validation for these templates.

tron/core/jobrun.py
chlgit: Ah, gotcha.

tron/core/actionrun.py
keymone: should probably belong in config validation, like the command attribute?

PR #529, opened 2018-09-14
Make fail final and move retries and exit logic out

qui:
If someone uses `tronctl fail` on an action with retries, even if that action hasn't started yet, it will trigger restarts. So I've made fail move straight to the failed state, and replaced `fail` with `_exit_unsuccessful` everywhere we want to consider retries.

I also took back my change in https://github.com/Yelp/Tron/pull/522 so that you can still send kill requests to Mesos even if Tron thinks it's not running - for instance, if you manually failed it.

Internal ticket TRON-438.

### DIFF
--- a/tron/core/actionrun.py
+++ b/tron/core/actionrun.py
@@ -335,7 +335,7 @@ class ActionRun(object):
         if run.is_running:
             run._done('fail_unknown')
         if run.is_starting:
-            run.fail(None)
+            run._exit_unsuccessful(None)
         return run
 
     def start(self):
@@ -403,7 +403,7 @@ class ActionRun(object):
             self.retries_remaining = 1
 
         if self.is_done:
-            return self.fail(self.exit_status)
+            return self._exit_unsuccessful(self.exit_status)
         else:
             log.info(f"Killing action run {self.id} for a retry")
             return self.kill(final=False)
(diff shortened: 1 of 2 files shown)


tron/core/actionrun.py
```diff
@@ -733,27 +735,31 @@ def stop(self):
...
         if self.mesos_task_id is None:
-            return "Error: Can't find task id for the action."
-        succeeded = mesos_cluster.kill(self.mesos_task_id)
-        if not succeeded:
-            return "Error while killing task. Please try again."
+            msgs.append("Error: Can't find task id for the action.")
+        else:
+            succeeded = mesos_cluster.kill(self.mesos_task_id)
+            if not succeeded:
+                msgs.append("Error while killing task. Please try again.")
```
solarkennedy: Can you also emit the positive feedback that it did work?
Normally I think it is OK for no-news to be good news, but with more rare-commands like kill, it is nice to know that it did work. Heck even emitting the actual taskid it killed would be nice

tron/core/actionrun.py
```diff
@@ -403,7 +403,7 @@ def retry(self):
             self.retries_remaining = 1
 
         if self.is_done:
-            return self.fail(self.exit_status)
+            return self._exit_unsuccessful(self.exit_status)
```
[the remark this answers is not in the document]
qui: The retry function is only called externally, i.e. from the API. We assume that the user wants to force a retry, so we don't care if it was successful or not, and _exit_unsuccessful is where retries get triggered.

tron/core/actionrun.py
qui: Sure.

PR #534, opened 2018-09-19
remove percent string support

chlgit:
Maybe it's worth in a separated release, so we can roll back easily.

### DIFF
--- a/tests/config/config_parse_test.py
+++ b/tests/config/config_parse_test.py
@@ -1108,11 +1108,11 @@ class TestBuildFormatStringValidator(TestCase):
         self.validator = build_format_string_validator(self.context)
 
     def test_validator_passes(self):
-        template = "The %(one)s thing I %(seven)s is %(stars)s"
+        template = "The {one} thing I {seven} is {stars}"
         assert self.validator(template, NullConfigContext)
 
     def test_validator_unknown_variable_error(self):
-        template = "The %(one)s thing I %(seven)s is %(unknown)s"
+        template = "The {one} thing I {seven} is {unknown}"
         exception = assert_raises(
             ConfigError,
             self.validator,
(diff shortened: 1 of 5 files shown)


chlgit:
After testing, I think one release is fine.

tests/config/config_parse_test.py
```diff
@@ -1157,18 +1131,8 @@ def test_validator_passes_with_context(self):
...
-        template = "The {one} thing I {seven} is {mars}"
-        context = config_utils.ConfigContext(
-            path=None,
-            nodes=None,
-            command_context={'mars': 'ok'},
-            namespace=None,
-        )
-        assert self.validator(template, context)
-
     def test_validator_valid_percent_string(self):
```
qui: do we still need this test? if we do, let's rename

tron/config/config_parse.py
```diff
@@ -70,14 +70,7 @@ def validator(value, config_context):
...
             valid_context_variable_expr(value, config_context)
-            try:
-                value % context
-                StringFormatter(context).format(value)
-                return value
-            except Exception as e:
-                value % context
-                log.warning("Invalid format string {} at {} {}.\nFall back to % string evaluation".format(e, config_context.path, value))
-                return value
+            return StringFormatter(context).format(value)
```
qui: Do the error cases below still apply? i.e. does StringFormatter raise the same exceptions in the same situations?

chlgit: I think they are still needed here. For example, KeyError can be for unknown context variables, while TypeError can be not enough argument (Theoretically possible, but might never happen in our use cases).

tests/config/config_parse_test.py
chlgit: I've rewritten the test to see if we don't need a %% escape anymore.

PR #535, opened 2018-09-20
Simpler state machine

keymone:
- no longer observable, has no delegate object
- states are simple strings
- full transition map in machine directly
- transitions are single step only
- circular transition check removed
- actionrun and actioncommand are now observables instead


tron/actioncommand.py
```diff
@@ -41,48 +27,58 @@ class ActionCommand(object):
...
-    PENDING = ActionState('pending', start=RUNNING, exit=FAILSTART)
+    PENDING = 'pending'
+    RUNNING = 'running'
+    EXITING = 'exiting'
+    COMPLETE = 'complete'
+    FAILSTART = 'failstart'
+
+    STATE_MACHINE = Machine(
+        PENDING,
+        **{
```
solarkennedy: I know what `**` does, but looking at this code I don't really understand it.
Are you sure normal kwargs wouldn't be more clear here?

tron/actioncommand.py
```diff
@@ -41,48 +27,58 @@ class ActionCommand(object):
...
+            self.stderr = filehandler.NullFileHandle
 
     @property
     def state(self):
         return self.machine.state
 
-    @property
-    def attach(self):
-        return self.machine.attach
+    def transition_and_notify(self, target):
```
solarkennedy: These functions used to be able to return false, but now I think they will return an implicit None, is that OK?

tron/actioncommand.py
```diff
@@ -122,31 +112,15 @@ def is_failed(self):
     @property
     def is_complete(self):
         """Complete implies done and success."""
-        return self.machine.state == self.COMPLETE
+        return self.machine.state == 'complete'
```
solarkennedy: Why are these and and :120 the odd ones out?  These are still part of the same object, why not use the constants?

tests/api/resource_test.py
```diff
@@ -268,14 +268,6 @@ def test_get_run_from_identifier_number(self):
         assert_equal(job_run, self.job_runs.get_run_by_num.return_value)
         self.job_runs.get_run_by_num.assert_called_with(3)
 
-    def test_get_run_from_identifier_state_name(self):
-        job_run = self.resource.get_run_from_identifier('SUCC')
```
solarkennedy: Huh. I kinda wonder what the original authors' intent was with this. 
I agree that I think we can safely get rid of it.

tron/actioncommand.py
keymone: the problem with normal kwargs is that `PENDING=xyz` is interpreted as `"PENDING"=xyz` instead of `<value of PENDING>=xyz`

tron/actioncommand.py
keymone: i'm ok with that. even if there are too many things falsey in python, `None` is one i agree with.

tron/actioncommand.py
keymone: yep, that's a mistake, fixing

tron/utils/state.py
```diff
@@ -1,171 +1,70 @@
...
+    @staticmethod
+    def from_machine(machine, initial=None, state=None):
+        if initial is None:
+            initial = machine.initial
+        if state is None:
+            state = initial
+        new_machine = Machine(initial, **machine.transitions)
+        new_machine.state = state
+        assert machine.transitions == new_machine.transitions
+        assert machine.states == new_machine.states
```
chlgit: I don't understand here. Why do we need the assert? What gonna happen if the assert fails?

tron/utils/state.py
```diff
@@ -1,171 +1,70 @@
...
+        return new_machine
+
+    def __init__(self, initial, **transitions):
+        super().__init__()
+        self.transitions = transitions
```
chlgit: How about self.transitions = defaultdict(dict, transitions), so we don't need line 32 - 34?

PR #536, opened 2018-09-20
Allow actions to be a dict instead of a list

qui:
For Mesos actions, it would be nice to have actions be a dict so we can use YAML anchors.
e.g.
```
actions:
  one: &template
    command: echo 1
    cpus: 0.1
    ...
  two:
    <<: *template
    command: echo 2
```  
rather than
```
actions:
  - name: one
    command: echo 1
    cpus: 0.1
    ...
  - name: two
    command: echo 2
    cpus: 0.1
    ...
``` 

Will also need to make changes to paasta_tools to allow dicts, if we like this change.

### DIFF
--- a/tests/config/config_parse_test.py
+++ b/tests/config/config_parse_test.py
@@ -211,9 +211,9 @@ def make_master_jobs():
                 expected_runtime=datetime.timedelta(1),
                 cleanup_action=None,
             ),
-        'MASTER.test_job3':
+        'MASTER.test_job_actions_dict':
             make_job(
-                name='MASTER.test_job3',
+                name='MASTER.test_job_actions_dict',
                 node='node1',
                 schedule=ConfigConstantScheduler(),
                 actions=FrozenDict({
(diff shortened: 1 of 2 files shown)


solarkennedy:
OMG this is so great!

keymone:
i think we should do the same for jobs. the reason it's lists right now is mostly because it's possible to validate name uniqueness, while duplicate hashmap keys are ignored during yaml parsing. maybe we can make a validation that can detect duplicates in yaml?

solarkennedy:
Yes! I didn't want to say it, but yea I was hoping for jobs next.

qui:
@keymone @solarkennedy I didn't try to do this, but it works for jobs already! validate_jobs uses build_dict_name_validator too. 

@keymone good point on name uniqueness... this seems to be a general pyyaml problem. I can't think of a good way to validate a file *before* we load it with pyyaml. There are some hacks like https://gist.github.com/pypt/94d747fe5180851196eb to change pyyaml's parsing, or we could start using ruamel.

PR #537, opened 2018-09-20
Add handlers check in MesosTask logger to prevent duplicate log lines

kawaiwanyelp:
# Description
- Bug (behavior): Retrying Mesos actions would produce duplicate log lines by the same number of retries attempted so far.
- Bug (code): Whenever a Mesos task was created and its logger initialized, it adds a handler to the logger to log output to stderr. Unfortunately, there was no check in place to prevent duplicate handlers from being added, so each task (retry) added another handler, producing another duplicate log output.
- Fix: Whenever we initialize a Mesos task, we check the number of handlers in the logger, and only add the stderr handler if there are no handlers present.
- Additional changes: Added a test for MesosTask's log initializer function to ensure only 1 handler is ever added to the logger.

# Testing done
- make dev

### DIFF
--- a/tests/mesos_test.py
+++ b/tests/mesos_test.py
@@ -254,11 +254,11 @@ class TestMesosTask(TestCase):
         Ensures that only a single handler (for stderr) is added to the
         MesosTask event logger, to prevent duplicate log output.
         """
-        log1 = self.task.get_event_logger()
-        log2 = self.task.get_event_logger()
+        # Call 2 times to make sure 2nd call doesn't add another handler
+        logger = self.task.get_event_logger()
+        logger = self.task.get_event_logger()
 
-        assert len(log1.handlers) == len(log2.handlers)
-        assert len(log2.handlers) == 1
+        assert len(logger.handlers) == 1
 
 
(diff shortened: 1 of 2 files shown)


tests/mesos_test.py
```diff
@@ -249,6 +249,17 @@ def test_log_event_error(self):
...
 
+    def test_get_event_logger_add_unique_handlers(self):
+        """
+        Ensures that only a single handler (for stderr) is added to the
+        MesosTask event logger, to prevent duplicate log output.
+        """
+        log1 = self.task.get_event_logger()
+        log2 = self.task.get_event_logger()
+
+        assert len(log1.handlers) == len(log2.handlers)
```
chlgit: Out of curiosity, does "assert log1 == log2" work here?

kawaiwanyelp: Yes, it would. `self.task.get_event_logger()' always returns the same logger (since all MesosTasks share the same logger). So, actually, `assert len(log1.handlers) == len(log2.handlers)` will always be `True` since log1 and log2 are the same reference. I think I will remove this check and just keep the subsequent length = 1 check.

PR #539, opened 2018-09-24
Rollback script for v0.9.6.1 to 0.9.5.1

chlgit:
This is the script in case that anything goes wrong for the upgrade from 0.9.5.1 to 0.9.6.1.


chlgit:
This script requires modification for Tron source code. I will close it rather than merging to the /contrib.

PR #541, opened 2018-09-26
Remove http stacktraces from tronview and tronctl error messages

kawaiwanyelp:
### Description
- Bug (behavior): If `tronview` and `tronctl` were given erroneous commands in some way, the programs show HTTP stacktraces (and other related errors), which ideally should not be shown to the user. 
- Bug (code):
  1. `tronview`: Did not handle RequestErrors if raised when bad identifiers (for job runs or actions) were used.
  2. `tronctl`: Despite the fact that we deprecated non-id commands, we still allowed tronctl to run without an id.
  3. tron api: Although the tron client expects JSON error messages, the tron api returns HTML (with an error message embedded).
- Fix:
  1. `tronview`: Ensure RequestErrors are handled, and an error message from tron api is displayed to the user.
  2. `tronctl`: Enforce the need for at least job/jobrun/actionrun id as a positional argument
  3. tron api:
    - Create a new resource type that returns JSON messages in the event of an error
    - Since some errors are printed no matter what because they are logs, I modified the logging levels according to verbosity requested.
      - None: only critical
      - -v: warning, error, critical
      - -vv: info, warning, error, critical
      - -vvv: all

### Testing
- make test
- make dev
- Sample output:
1. In `tronview`, when querying the status for nonexistent jobs, there were no problems to begin with:
`---@---:~$ tronview foo`
`Error: Unknown identifier: foo`
    - But when querying a nonexistent job run (or action run) for an existing job, we originally got a stack trace:
`---@---:~$ tronview video.retry_video_processing.foo`
`tron.commands.client ERROR Received error response: HTTP Error 404: Not Found`
`tron.commands.client WARNING Incorrectly formatted error`
`<html>`
`<head><title>404 - No Such Resource</title></head>`
...
    - Now we get:
`---@---:~$ tronview MASTER.test.foo`
`Error: Cannot find job run 'foo' for 'Job:MASTER.test'`
    - Even if we query an action for a nonexistent job run, tron will recognize that the job run identifier, and not the action run identifier, is the problem:
`---@---:~$ tronview MASTER.test.foo.bar`
`Error: Cannot find job run 'foo' for 'Job:MASTER.test'`
    - But, if the job run identifier is good, but the action identifier is bad, tron also recognizes that:
`---@---:~$ tronview MASTER.test.0.bar`
`Error: Cannot find action 'bar' for 'JobRun:MASTER.test.0'`
2. For `tronctl`, when using a bad command, the program identifies that it is invalid:
`---@---:~$ tronctl foo`
...
`tronctl: error: argument command: invalid choice: 'foo' (choose from 'start', 'rerun', 'retry', 'cancel', 'backfill', 'disable', 'enable', 'fail', 'success', 'skip', 'stop', 'kill')`
    - But when using a command that requires an argument (i.e. all of them), tron originally saw them as 'Not Implemented':
`---@---:~$ tronctl start`
`tron.commands.client ERROR Received error response: HTTP Error 501: Not Implemented`
`tronctl ERROR Unknown command start`
    - Now, `tronctl` will require that at least 1 identifier is passed:
`---@---:~$ tronctl start`
`tronctl: error: the following arguments are required: id`
3. For `tronctl`, when using a command with an invalid job id, the program already identifies the job is invalid, but not when the job run (or action) is bad:
`---@---:~$ tronctl start foo`
`Error: Unknown identifier: foo`
`---@---:~$ tronctl start video.retry_video_processing.foo`
`tron.commands.client ERROR Received error response: HTTP Error 404: Not Found`
`tron.commands.client WARNING Incorrectly formatted error`
`<html>`
`<head><title>404 - No Such Resource</title></head>`
...
    - Now, tron will correctly identify that a job run (or action) is incorrect:
`---@---:~$ tronctl start MASTER.test.foo`
`Error: Cannot find job run 'foo' for 'Job:MASTER.test'`
    - And if a command doesn't work for a certain identifier (say, disable a job run, even though `disable` is only valid for a job), tron will also identify that:
`tronctl disable MASTER.test.0`
`Error: Unknown command 'disable' for 'JobRun:MASTER.test.0'`

### DIFF
--- a/bin/tronview
+++ b/bin/tronview
@@ -99,8 +99,7 @@ def view_job_run(args, job_run_id, client):
     try:
         actions = client.job_runs(job_run_id.url)
     except RequestError as e:
-        response = e.args[1]
-        raise SystemExit(f"Error: {response.content}")
+        raise SystemExit(f"Error: {e}")
 
     display_action = display.DisplayActionRuns()
     return display_action.format(actions)
@@ -113,8 +112,7 @@ def view_action_run(args, act_run_id, client):
             num_lines=args.num_displays,
         )
     except RequestError as e:
-        response = e.args[1]
-        raise SystemExit(f"Error: {response.content}")
+        raise SystemExit(f"Error: {e}")
 
     return display.format_action_run_details(content)
 
--- a/bin/tronview
+++ b/bin/tronview
@@ -12,6 +12,7 @@ from tron.commands import cmd_utils
 from tron.commands import display
 from tron.commands.client import Client
 from tron.commands.client import get_object_type_from_identifier
+from tron.commands.client import RequestError
 from tron.commands.client import TronObjectType
 from tron.commands.cmd_utils import ExitCode
 from tron.commands.cmd_utils import suggest_possibilities
@@ -95,16 +96,26 @@ def view_job(args, job_id, client):
 
 
 def view_job_run(args, job_run_id, client):
-    actions = client.job_runs(job_run_id.url)
+    try:
+        actions = client.job_runs(job_run_id.url)
+    except RequestError as e:
+        response = e.args[1]
+        raise SystemExit(f"Error: {response.content}")
+
     display_action = display.DisplayActionRuns()
     return display_action.format(actions)
 
 
 def view_action_run(args, act_run_id, client):
-    content = client.action_runs(
-        act_run_id.url,
-        num_lines=args.num_displays,
-    )
+    try:
+        content = client.action_runs(
+            act_run_id.url,
+            num_lines=args.num_displays,
+        )
+    except RequestError as e:
+        response = e.args[1]
+        raise SystemExit(f"Error: {response.content}")
+
     return display.format_action_run_details(content)
 
 
--- a/bin/tronview
+++ b/bin/tronview
@@ -96,24 +96,16 @@ def view_job(args, job_id, client):
 
 
 def view_job_run(args, job_run_id, client):
-    try:
-        actions = client.job_runs(job_run_id.url)
-    except RequestError as e:
-        raise SystemExit(f"Error: {e}")
-
+    actions = client.job_runs(job_run_id.url)
     display_action = display.DisplayActionRuns()
     return display_action.format(actions)
 
 
 def view_action_run(args, act_run_id, client):
-    try:
-        content = client.action_runs(
-            act_run_id.url,
-            num_lines=args.num_displays,
-        )
-    except RequestError as e:
-        raise SystemExit(f"Error: {e}")
-
+    content = client.action_runs(
+        act_run_id.url,
+        num_lines=args.num_displays,
+    )
     return display.format_action_run_details(content)
 
 
@@ -137,7 +129,11 @@ def get_view_output(name, args, client):
 
     if tron_id.type not in obj_type_to_view_map:
         return
-    return obj_type_to_view_map[tron_id.type](args, tron_id, client)
+
+    try:
+        return obj_type_to_view_map[tron_id.type](args, tron_id, client)
+    except RequestError as e:
+        raise SystemExit(f"Error: {e}")
 
 
 def main():
--- a/tron/api/resource.py
+++ b/tron/api/resource.py
@@ -155,8 +155,8 @@ class JobRunResource(resource.Resource):
             action_run = self.job_run.action_runs[action_name]
             return ActionRunResource(action_run, self.job_run)
 
-        msg = "Cannot find action '%s' for '%s'"
-        return ErrorResource(msg % (action_name, self.job_run))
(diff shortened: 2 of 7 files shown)


tron/api/resource.py
```diff
@@ -133,8 +155,8 @@ def getChild(self, action_name, _):
             action_run = self.job_run.action_runs[action_name]
             return ActionRunResource(action_run, self.job_run)
 
-        msg = "Cannot find action %s for %s"
-        return resource.NoResource(msg % (action_name, self.job_run))
+        msg = "Cannot find action '%s' for '%s'"
```
solarkennedy: Yuck. Just fstring this thing and merge with 159

tron/api/resource.py
```diff
@@ -184,8 +206,9 @@ def getChild(self, run_id, _):
         if run_id in job.action_graph.names:
             action_runs = job.runs.get_action_runs(run_id)
             return ActionRunHistoryResource(action_runs)
-        msg = "Cannot find job run %s for %s"
-        return resource.NoResource(msg % (run_id, job))
+
+        msg = "Cannot find job run '%s' for '%s'"
+        return ErrorResource(msg % (run_id, job))
```
solarkennedy: Well, idk same with these, I think fstrings will help make the code more readable, but is inconsequential to the user-facing behavior

tron/commands/client.py
```diff
@@ -183,7 +183,7 @@ def request(self, url, data=None):
         uri = urllib.parse.urljoin(self.url_base, url)
         response = request(uri, data)
         if response.error:
-            raise RequestError("%s: %s" % (uri, response))
+            raise RequestError("%s: %s" % (uri, response), response)
```
solarkennedy: This looks a little funny to me because there is the tuple that gets parsed by the string, and then there is a second argument to the exception? 

fstrings again I think will make this more clear, because the response object will be turned into a string, right?

bin/tronview
```diff
@@ -95,16 +96,26 @@ def view_job(args, job_id, client):
 
 
 def view_job_run(args, job_run_id, client):
-    actions = client.job_runs(job_run_id.url)
+    try:
+        actions = client.job_runs(job_run_id.url)
+    except RequestError as e:
+        response = e.args[1]
```
solarkennedy: Now I understand. I've never seen this pattern before for moving objects between exceptions.

I don't know man, can you... not do this and instead only pass a string and do whatever parsing you need to do at exception handling time?

bin/tronview
```diff
@@ -95,16 +96,26 @@ def view_job(args, job_id, client):
...
+        raise SystemExit(f"Error: {response.content}")
```
solarkennedy: In other words, I don't think it is "safe" for this function to "know" that we are going to be getting an object like this that is safe to inspect. 

If you would like to read another opinion on handling exceptions like this, you might like (the whole blog post, but this part in particular) 
https://blog.philipphauer.de/checked-exceptions-are-evil/#exposing-implementation-details-in-the-signature

tron/commands/client.py
qui: How about returning RequestError(response.content), since response.content is the error message you care about? Then tronview can just print the RequestError instead of inspecting it.

tron/api/resource.py
kawaiwanyelp: Oops, should've caught this.

tron/api/resource.py
kawaiwanyelp: And same here, should've caught it.

tron/commands/client.py
kawaiwanyelp: Yeah, I was thinking the same thing @qui.

bin/tronview
```diff
@@ -95,16 +96,24 @@ def view_job(args, job_id, client):
 
 
 def view_job_run(args, job_run_id, client):
-    actions = client.job_runs(job_run_id.url)
+    try:
+        actions = client.job_runs(job_run_id.url)
+    except RequestError as e:
```
qui: You could put this try/except in get_view_output, where the `obj_type_to_view_map` function is called on the client, to avoid repeating.

tron/commands/client.py
```diff
@@ -179,11 +179,12 @@ def http_get(self, url, data=None):
         return self.request(build_get_url(url, data))
 
     def request(self, url, data=None):
-        log.info("Request: %s, %s, %s", self.url_base, url, data)
+        log.info(f'Request: {self.url_base}, {url}, {data}')
         uri = urllib.parse.urljoin(self.url_base, url)
         response = request(uri, data)
         if response.error:
-            raise RequestError("%s: %s" % (uri, response), response)
+            print('in here', response.content)
```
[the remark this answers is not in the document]
kawaiwanyelp: Sigh....

PR #544, opened 2018-09-26
Convert Action to dataclass

keymone:
`dataclass` will be in stdlib in python 3.7 and there are clear benefits to using it as opposed to custom made objects with attributes so let's just gradually convert everything. I'm starting with Action here  because  that's the lowest level object and therefore smallest possible change.

### DIFF
--- a/tron/core/action.py
+++ b/tron/core/action.py
@@ -1,80 +1,37 @@
-from __future__ import absolute_import
-from __future__ import unicode_literals
-
+import datetime
 import logging
+
+from dataclasses import dataclass
+from dataclasses import field
 
 from tron import node
 from tron.config.schema import CLEANUP_ACTION_NAME
-from tron.utils import maybe_decode
 
 log = logging.getLogger(__name__)
 
 
-class Action(object):
+@dataclass
+class Action:
     """A configurable data object for an Action."""
-
-    equality_attributes = [
-        'name',
-        'command',
-        'node_pool',
-        'is_cleanup',
-        'retries',
-        'expected_runtime',
-        'executor',
-        'cpus',
-        'mem',
-        'constraints',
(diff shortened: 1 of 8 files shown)


tron/core/action.py
```diff
@@ -107,30 +52,29 @@ def from_config(cls, config):
...
-        return attributes_match and all(
-            self_act == other_act for (
-                self_act,
-                other_act,
-            ) in zip(self.required_actions, other.required_actions)
-        )
+        # Only convert config values if they are not None.
+        constraints = config.constraints
+        if constraints:
+            constraints = set((c.attribute, c.operator, c.value) for c in constraints)
```
qui: Do we need to turn them into normal tuples, if the constraint/parameter/volume configs are already namedtuples?

tron/core/actiongraph.py
```diff
@@ -61,10 +61,20 @@ def get_required_actions(self, name):
...
-        return self.action_map[name].required_actions
+
+        return (
+            self.action_map[action]
+            for action in self.action_map[name].required_actions
+        )
 
     def get_dependent_actions(self, name):
-        return self.action_map[name].dependent_actions
+        if name not in self.action_map:
```
solarkennedy: Can you help me understand why this funtion needs to change in this new world?

Is it because we can't return a set? It has a default of an empty set, does that not work or do we get a key error? I don't understand.

keymone: not entirely sure which change you're asking about so i'll address both:

- the `if` change isn't strictly necessary, i just copied it from `get_required_actions` because i like consistency

- the change in `return` statement is necessary because we need to return `Action` objects but we no longer have them in `required_actions / dependent_actions` - we only have action names

PR #545, opened 2018-09-27
Change tronfig updates to include the enabled attribute

kawaiwanyelp:
### Description
- Previously, except for creating new jobs, a tronfig's enabled attribute was irrelevant; enabling or disabling a job via tronctl took precedent.
- Now, if a new tronfig is deployed, the tronfig's enabled attribute will override a job's enabled state, as if it was enabled/disabled in tronctl. In other words, the latest change to a job's enabled state, be it from tronctl or tronfig, will be the one used.

### Testing done
- make test
- make dev (manual testing)

### DIFF
--- a/tests/core/job_scheduler_test.py
+++ b/tests/core/job_scheduler_test.py
@@ -245,19 +245,22 @@ class TestJobSchedulerSchedule(TestCase):
 class TestJobSchedulerOther(TestCase):
     """ Test other JobScheduler functions """
 
-    @setup
-    def setup_job(self):
-        self.scheduler = mock.Mock()
+    def _make_job_scheduler(self, job_name, enabled=True):
+        scheduler = mock.Mock()
         run_collection = mock.Mock()
         node_pool = mock.Mock()
-        self.job = job.Job(
-            "jobname",
-            self.scheduler,
+        new_job = job.Job(
+            job_name,
+            scheduler,
             run_collection=run_collection,
             node_pool=node_pool,
-            enabled=True,
+            enabled=enabled,
         )
-        self.job_scheduler = JobScheduler(self.job)
+        return new_job, JobScheduler(new_job)
+
+    @setup
+    def setup_job(self):
+        self.job, self.job_scheduler = self._make_job_scheduler('jobname', True)
 
     def test_disable(self):
         self.job.runs.cancel_pending = mock.Mock()
@@ -268,17 +271,7 @@ class TestJobSchedulerOther(TestCase):
         assert self.job.runs.cancel_pending.call_count == 1
 
     def test_update_from_job_scheduler_disable(self):
-        new_scheduler = mock.Mock()
-        new_run_collection = mock.Mock()
-        new_node_pool = mock.Mock()
-        new_job = job.Job(
-            "jobname",
-            new_scheduler,
-            run_collection=new_run_collection,
-            node_pool=new_node_pool,
-            enabled=False,
-        )
-        new_job_scheduler = JobScheduler(new_job)
+        new_job, new_job_scheduler = self._make_job_scheduler('jobname', False)
         self.job.update_from_job = mock.Mock()
         self.job_scheduler.disable = mock.Mock()
 
@@ -290,19 +283,10 @@ class TestJobSchedulerOther(TestCase):
         assert self.job_scheduler.disable.call_count == 1
 
     def test_update_from_job_scheduler_enable(self):
-        new_scheduler = mock.Mock()
-        new_run_collection = mock.Mock()
-        new_node_pool = mock.Mock()
-        new_job = job.Job(
-            "jobname",
-            new_scheduler,
-            run_collection=new_run_collection,
-            node_pool=new_node_pool,
-            enabled=True,
-        )
-        new_job_scheduler = JobScheduler(new_job)
+        new_job, new_job_scheduler = self._make_job_scheduler('jobname', True)
         self.job.update_from_job = mock.Mock()
         self.job.enabled = False
+        self.job.config_enabled = False
         self.job_scheduler.enable = mock.Mock()
 
         self.job_scheduler.update_from_job_scheduler(new_job_scheduler)
@@ -311,6 +295,22 @@ class TestJobSchedulerOther(TestCase):
             new_job_scheduler.get_job()
         )
         assert self.job_scheduler.enable.call_count == 1
+
+    def test_update_from_job_scheduler_no_config_change(self):
+        new_job, new_job_scheduler = self._make_job_scheduler('jobname', False)
+        new_job.config_enabled = True  # same as self.job
+        self.job.update_from_job = mock.Mock()
+        self.job_scheduler.enable = mock.Mock()
+        self.job_scheduler.disable = mock.Mock()
+
+        self.job_scheduler.update_from_job_scheduler(new_job_scheduler)
+
+        assert self.job.update_from_job.call_args == mock.call(
+            new_job_scheduler.get_job()
+        )
+        assert self.job_scheduler.enable.call_count == 0
+        assert self.job_scheduler.disable.call_count == 0
+        assert self.job.config_enabled == new_job.config_enabled
 
 
 class TestJobSchedulerFactory(TestCase):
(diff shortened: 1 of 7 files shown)


qui:
What about this case?
1. job is enabled in tronfig (by default)
2. job is disabled using tronctl
3. other part of job is changed in tronfig, but job is still enabled (might also be an issue if other part of job's namespace, or any job, is changed)

We want the job to remain disabled.

kawaiwanyelp:
Ah, it was my impression that we wanted to provide parity between tronfig and tronctl to enable/disable jobs, but it seems that we only want to allow tronfig to enable/disable when the enabled attribute in the job configuration changes.

tron/core/job.py
```diff
@@ -89,6 +90,9 @@ def __init__(
         self.queueing = queueing
         self.all_nodes = all_nodes
         self.enabled = enabled
+        # config_enabled is equivalent to the setting in file, whereas enabled
+        # is the current enable setting
+        self.config_enabled = enabled if config_enabled is None else config_enabled
```
qui: When a job is created, the `enabled` kwarg comes from the config, so we should always set config_enabled = enabled. That way, even if tronctl changes self.enabled, config_enabled will still reflect the original config value.

You also don't need to save it in state because it just comes from the config.

tron/core/job_scheduler.py
```diff
@@ -85,6 +85,26 @@ def schedule(self):
...
+        """ Update a job scheduler by copying another. """
+        old_job = self.get_job()
+        new_job = job_scheduler.get_job()
+
+        old_job.update_from_job(new_job)
+
+        # Since job upadting only copies equality attributes (defined in the Job
+        # class), we need to now enable or disable the job depending on the
+        # new job says so.
+        if old_job.enabled is not new_job.enabled and \
```
qui: Do you need this? I think you could simplify 98-107 to:
```
if old_job.config_enabled is not new_job.config_enabled:
  old_job.enabled = new_job.config_enabled
old_job.config_enabled = new_job.config_enabled
```

tron/core/job_scheduler.py
```diff
@@ -85,6 +85,26 @@ def schedule(self):
             return
         self.create_and_schedule_runs()
 
+    def update_from_job_scheduler(self, job_scheduler):
+        """ Update a job scheduler by copying another. """
+        old_job = self.get_job()
```
qui: I think this should be called current_job, because we keep it. old_job makes it sound like we don't use it anymore.

tests/core/job_scheduler_test.py
```diff
@@ -242,6 +242,77 @@ def test_run_queue_schedule(self, reactor):
...
+        self.job_scheduler.update_from_job_scheduler(new_job_scheduler)
+
+        assert self.job.update_from_job.call_args == mock.call(
+            new_job_scheduler.get_job()
+        )
+        assert self.job_scheduler.enable.call_count == 1
+
+    def test_update_from_job_scheduler_no_config_change(self):
+        new_job, new_job_scheduler = self._make_job_scheduler('jobname', False)
+        new_job.config_enabled = True  # same as self.job
```
qui: In the new job, config_enabled should always be the same as enabled. You should test the case where self.job.config_enabled = True, self.job.enabled = False (i.e. manually disabled with tronctl), and then new_job is enabled. That's essentially the case I described in my earlier PR comment.

tron/core/job.py
```diff
@@ -176,12 +180,22 @@ def state_data(self):
...
+            'enabled': {
+                'current': self.enabled,
+                'config': self.config_enabled,
+            }
         }
 
     def get_job_runs_from_state(self, state_data):
         """Apply a previous state to this Job."""
-        self.enabled = state_data['enabled']
+        if self.config_enabled is not state_data['enabled']['config']:
```
keymone: is difference between `False` and `None` important here?

tron/core/job.py
```diff
@@ -176,12 +180,22 @@ def state_data(self):
         """This data is used to serialize the state of this job."""
         return {
             'runs': self.runs.state_data,
-            'enabled': self.enabled,
+            'enabled': {
```
keymone: is this data used in  tronview / tronweb? the change in structure will have to be reflected.

tron/core/job_scheduler.py
```diff
@@ -85,6 +85,26 @@ def schedule(self):
...
+        new_job = job_scheduler.get_job()
+
+        old_job.update_from_job(new_job)
+
+        # Since job upadting only copies equality attributes (defined in the Job
+        # class), we need to now enable or disable the job depending on the
+        # new job says so.
+        if old_job.enabled is not new_job.enabled and \
+           old_job.config_enabled is not new_job.config_enabled:
+            if new_job.enabled:
```
keymone: something is off with indentation here or in previous line

tron/core/job.py
kawaiwanyelp: Originally, I believed that we didn't need to save config_enabled in state; however, there is a relevant edge case: 
1. Create job from config, enabled=False.
2. Turn off trond, enabled=False saved to state.
3. Change the config, enabled=True.
4. Turn on trond again.
    - No config_enabled: trond would recreate the job from config (enabled=True), then overwrite it from saved state (enabled=False). Here, tronfig will not acknowledge a config change that happens offline.
    - With config_enabled: trond would recreate the job from config (enabled=True, config_enabled=True), then attempt to overwrite it from state (enabled=False, config_enabled=False). It would see the difference in config_enabled between config and state, and therefore keep enabled=True.

tron/core/job_scheduler.py
kawaiwanyelp: Agreed.

tron/core/job_scheduler.py
kawaiwanyelp: I do want to keep thing as they are for 2 reasons:
1. Logging: Although reconfiguration is logged elsewhere, I'd like re-enabling/disabling by tronctl to be logged as well. And of course, those logs are sort of unnecessary if the job was already enabled/disabled to begin with, hence the conditions.
2. Enable()/disable(): Although setting job.enabled is enough technically, I use enable()/disable() so that tronfig is like tronctl when it comes to modifying job state. Additionally, those functions do things like schedule runs as well as cancel pending ones.

tron/core/job.py
kawaiwanyelp: It would've been, but given @qui's comment about initializing config_enabled, it can never be None now (unless self.enabled is somehow set to None).

tron/core/job.py
kawaiwanyelp: No, state_data is only used during serialization to tron_state. tronview and tronweb, by querying the api, only get access to the enabled attribute in a Job object.

tron/core/job_scheduler.py
kawaiwanyelp: You're right. I'll update it to follow the Python style guide.

tron/core/job.py
kawaiwanyelp: Yes, but neither should ever be `None` here, at least not now. Based on @qui's comments above, I've changed it so that `self.config_enabled` = `self.enabled` during initialization.

tests/core/job_scheduler_test.py
kawaiwanyelp: Fair enough.

tron/core/job.py
qui: This is an interesting edge case... I don't think you need to worry about it, because it is not possible to change the config while trond is off in any "normal" way. You can only do that by manually overwriting files in Tron's working directory.

kawaiwanyelp:
On @qui's suggestion, I have also reverted changes for preserving a job's config_enabled in state.

tron/core/job_scheduler.py
```diff
@@ -85,6 +85,26 @@ def schedule(self):
...
         self.create_and_schedule_runs()
 
+    def update_from_job_scheduler(self, job_scheduler):
+        """ Update a job scheduler by copying another. """
+        curr_job = self.get_job()
+        new_job = job_scheduler.get_job()
+
+        curr_job.update_from_job(new_job)
+
+        # Since job upadting only copies equality attributes (defined in the Job
```
Rob-Johnson: s/upadting/updating

PR #548, opened 2018-10-02
Skip state diagram in docs

keymone:
- skip generating state diagram in docs for now as it's broken since rework of state machine
- use separate requirements file for docs task


tox.ini
```diff
@@ -19,13 +19,15 @@ commands =
...
+deps =
+    --requirement={toxinidir}/requirements-docs.txt
 whitelist_externals=
     dot
     mkdir
 commands=
-    python tools/state_diagram.py
     mkdir -p docs/images
-    dot -Tpng -odocs/images/action.png action.dot
+    # python tools/action_dag_diagram.py
```
Rob-Johnson: just rm them

PR #549, opened 2018-10-02
Move bin/*.py to tron/bin to fix tests

keymone:
having things like `import action_runner` in `tests/bin` confuses me and also `pytest` when running after new tron release

- move bin/*.py to tron/bin and symlink back
- fix relevant tests


bin/action_runner.py
```diff
@@ -2,163 +2,8 @@
...
-        t.start()
-    returncode = run_proc(
-        output_path=args.output_dir,
-        run_id=args.run_id,
-        command=args.command,
-        proc=proc,
-    )
+from tron.bin.action_runner import main
 
-    for t in threads:
```
solarkennedy: You shouldn't need this, just modify setup.py to point to the new entrypoint

keymone: You are correct!

PR #550, opened 2018-10-02
External dependencies in tronview

keymone:
Mostly add expected and published triggers to tronview

also:
- remove six from touched files
- convert more formatting to f-strings
- remove some overly verbose logging
- fix example-cluster bin path

example lines from tronview:

    Waits for triggers  : MASTER.one.one.ymdhm.2018-10-10T04:10 (done)
    Publishes triggers  : ymdhm.2018-10-10T04:05


tron/api/adapter.py
```diff
@@ -171,6 +171,20 @@ def get_in_delay(self):
         if self._obj.in_delay is not None:
             return self._obj.in_delay.getTime() - time.time()
 
+    def get_triggered_by(self) -> str:
+        """ Returns a list of tuples: [(rendered_trigger, was_published), ...]
```
Rob-Johnson: doc looks incorrect

tron/commands/display.py
```diff
@@ -352,6 +352,8 @@ class DisplayActionRuns(TableDisplay):
         ('End time', 'end_time'),
         ('Final exit status', 'exit_status'),
         ('Retry exit statuses', 'exit_statuses'),
+        ('Waits for triggers', 'triggered_by'),
```
Rob-Johnson: this shows all of the triggers it waits for, right? is there a way to show the triggers it has received?

also, might be worth renaming as `Blocked on` and `Triggers Published:`

keymone: it will list all triggers and `(done)` next to those that were already published

PR #551, opened 2018-10-02
Fix constraints passing incorrectly in actionrun

kawaiwanyelp:
### Description
While switching actions to dataclasses (#544), @keymone changed where `constraints`, `docker_parameters` and `extra_volumes` were 'assembled' for a new action. While the code for the latter two attributes were copied correctly (using `e._as_dict()`), the code for `constraints` was not, resulting in a KeyError from taskproc when actions were scheduled to run on Mesos.

This PR is a small fix.

### Testing
`make dev`: test running a job using `tronctl start`


kawaiwanyelp:
I added constraints to the tests.

PR #552, opened 2018-10-05
Add option for alerting every job in check_tron_jobs

kawaiwanyelp:
### Description
For most jobs, alerting using their latest run is sufficient. Our current Sensu monitoring follows this model, such that if a run fails on one day and succeeds on the next, job owners will get an alert event followed by a resolved event. 

There are some jobs for which every run is precious; that is, if a run fails on one day but succeeds on the next, it does not mean the former run was fixed. Our monitoring currently DOES NOT cater to these jobs.

This PR modifies tron job monitoring via the check_tron_job script, such that if a job opts into it, owners can be alerted according to whether or not a there was a successful run on each given day, as opposed to a job's overall latest run was successful.

If a job does not opt in, job owners will still receive alerts the original way.

Key changes:
- Opt-in monitoring attribute: Setting the `check_that_every_day_has_a_successful_run` attribute to `True` in a job's monitoring config will make the job precious.
- Adding shortdate to job name in alert: Owners of precious jobs will have the runs bucketed by date, which will be reflected in the Sensu alerts they receive.

### Testing done
Note: Although this PR enforces daily run success, I used minute-based or hourly run success to gather data more quickly.
- `make dev` (local) and job in stage environment (remote): Running a job that was designed to succeed or fail every interval in an alternating manner. This interval was either 30 seconds, 1 minute, or 3 minutes.
- `make test`

### Notes to reviewers
- Example of a job opting into daily alerts: https://gist.github.com/kawaiwanyelp/f51041dfbd2256cb5ec4e2e326822a23
- Example of an ALERT event: https://gist.github.com/kawaiwanyelp/0b9ce31a9fa96d88e4f36f843b69cc8f
- Example of a RESOLVED event: https://gist.github.com/kawaiwanyelp/7bf35bae04967f904bfd0fedf000951d

### DIFF
--- a/docs/jobs.rst
+++ b/docs/jobs.rst
@@ -50,10 +50,6 @@ Optional Fields
       A comma-separated string of email destinations. Defaults to the "team"
       default.
 
-    **irc_channels**
-      A list of IRC channels to send alerts to. Defaults to the team setting.
-      Set an empty list to specify no IRC notifications.
-
     **slack_channels**
       A list of Slack channels to send alerts to. Defaults to the team setting.
       Set an empty list to specify no Slack notifications.
@@ -68,12 +64,6 @@ Optional Fields
     **tags** (default **None**)
       A list of arbitrary tags that can be used in handlers for different
       metadata needs.
-
-    **ttl** (default **None**)
-      A human readable time unit to set the check TTL. If the monitoring
-      framework does not hear from the check after this time unit, it will
-      spawn a new failing event. Defaults to None, meaning the framework will
-      only spawn events when it receives them.
 
     **component** (default **None**)
       A list of components affected by the event. A good example here would be
--- a/docs/jobs.rst
+++ b/docs/jobs.rst
@@ -30,8 +30,64 @@ Optional Fields
 ---------------
 
 **monitoring** (default **{}**)
-    (Beta Feature) Dictionary of key: value pairs to inform the monitoring
-    framework on how to alert teams for job failures.
+    Dictionary of key: value pairs to inform the monitoring framework on how to
+    alert teams for job failures.
+
+    **team**
+      Team responsible for the job. Must already be defined in the monitoring
+      framework.
+
+    **page** (default **False**)
+      Boolean on whether or not an alert for this job is page-worthy.
+
+    **runbook**
+      Runbook associated with the job.
+
+    **tip** (default **None**)
+      A short 1-line version of the runbook.
+
+    **notification_email**
+      A comma-separated string of email destinations. Defaults to the "team"
+      default.
+
+    **irc_channels**
+      A list of IRC channels to send alerts to. Defaults to the team setting.
+      Set an empty list to specify no IRC notifications.
+
+    **slack_channels**
+      A list of Slack channels to send alerts to. Defaults to the team setting.
+      Set an empty list to specify no Slack notifications.
+
+    **ticket** (default **False**)
+      A Boolean value to enable ticket creation.
+
+    **project** (default **None**)
+      A string representing the JIRA project that the ticket should go under.
+      Defaults to the team value.
+
+    **tags** (default **None**)
+      A list of arbitrary tags that can be used in handlers for different
+      metadata needs.
+
+    **ttl** (default **None**)
+      A human readable time unit to set the check TTL. If the monitoring
+      framework does not hear from the check after this time unit, it will
+      spawn a new failing event. Defaults to None, meaning the framework will
+      only spawn events when it receives them.
+
+    **component** (default **None**)
+      A list of components affected by the event. A good example here would be
+      to include the job that is being affected.
+
+    **description** (default **None**)
+      Human readable text giving more context on any monitoring events.
+
+    **check_that_every_day_has_a_successful_run** (default **False**)
+      If **True**, the latest job run each day will be checked to see if it was
+      successful.
+
+      If **False**, only the latest overall run will be checked to see if it was
+      successful.
 
 **queueing** (default **True**)
     If a job run is still running when the next job run is to be scheduled,
--- a/tests/bin/check_tron_jobs_test.py
+++ b/tests/bin/check_tron_jobs_test.py
@@ -986,7 +986,7 @@ class TestCheckJobs(TestCase):
 
 # These tests test job without succeeded/failed run scenarios
 
-    def test_job_waiting_for_first_run(self):
+    def test_job_no_runs_to_check(self):
         job_runs = {
             'status':
                 'scheduled',
@@ -1008,7 +1008,7 @@ class TestCheckJobs(TestCase):
         }
         run, state = check_tron_jobs.get_relevant_run_and_state(job_runs)
         assert_equal(run['id'], 'MASTER.test.1')
-        assert_equal(state, State.WAITING_FOR_FIRST_RUN)
+        assert_equal(state, State.NO_RUNS_TO_CHECK)
 
     def test_job_has_no_runs_at_all(self):
         job_runs = {
--- a/tests/bin/check_tron_jobs_test.py
+++ b/tests/bin/check_tron_jobs_test.py
@@ -1374,6 +1374,7 @@ class TestCheckPreciousJobs(TestCase):
         assert latest_run['run_num'] == 5
         assert state == check_tron_jobs.State.SUCCEEDED
 
+    @patch('time.time', mock.Mock(return_value=1539460800.0), autospec=None)
     def test_sort_runs_by_interval_day(self):
         run_buckets = check_tron_jobs.sort_runs_by_interval(self.job, 'day')
 
@@ -1415,9 +1416,11 @@ class TestCheckPreciousJobs(TestCase):
         assert results[0]['output'] == \
             "OK: fake_job is disabled and won't be checked."
 
+    @patch('time.time', mock.Mock(return_value=1539460800.0), autospec=None)
     @patch('check_tron_jobs.Client', autospec=True)
     def test_compute_check_result_for_job_enabled(self, mock_client):
         client = mock_client('fake_server')
+        self.job['monitoring']['check_every'] = 500
         check_tron_jobs.guess_realert_every = mock.Mock(return_value=1)
         check_tron_jobs.get_object_type_from_identifier = \
             mock.Mock(return_value=mock.Mock())
@@ -1435,3 +1438,6 @@ class TestCheckPreciousJobs(TestCase):
             'check_tron_job.fake_job-2018.10.12',
             'check_tron_job.fake_job-2018.10.13',
         ])
+        for res in results:
+            assert res['check_every'] == '300s'
+            assert check_tron_jobs.PRECIOUS_JOB_ATTR not in res
--- a/tests/bin/check_tron_jobs_test.py
+++ b/tests/bin/check_tron_jobs_test.py
@@ -1,12 +1,16 @@
 import time
 
+import mock
 from mock import patch
 from mock import PropertyMock
 
 from testifycompat import assert_equal
+from testifycompat import setup
 from testifycompat import TestCase
 from tron.bin import check_tron_jobs
 from tron.bin.check_tron_jobs import State
+
+_PRECIOUS_RUNS_ATTR = 'check_that_every_day_has_a_successful_run'
 
 
 class TestCheckJobs(TestCase):
@@ -939,6 +943,8 @@ class TestCheckJobs(TestCase):
                         ),
                 },
             ],
+            'monitoring':
+                {},
         }
         run, state = check_tron_jobs.get_relevant_run_and_state(job_runs)
         assert_equal(run['id'], 'MASTER.test.2')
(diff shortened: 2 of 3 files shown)


tron/bin/check_tron_jobs.py
```diff
@@ -57,6 +61,13 @@ def _timestamp_to_timeobj(timestamp):
     return time.strptime(timestamp, '%Y-%m-%d %H:%M:%S')
 
 
+def _are_runs_precious(monitoring_conf):
+    """ Checks if runs are 'precious', that is, we want to ensure there is a
+    successful run per day """
+    precious_jobs_attr = 'check_that_every_day_has_a_successful_run'
```
solarkennedy: I think this is OK if you are trying to save line length or whatever.
but if it is just a constant, just set it at the top.

Then bonus, the tests can reference it without copy/pasting

tron/bin/check_tron_jobs.py
```diff
@@ -115,23 +126,20 @@ def compute_check_result_for_job_runs(client, job, job_content):
         prefix = "UNKNOWN: The job is in a state that check_tron_jobs doesn't understand"
         status = 3
 
+    precious_runs_note = ''
+    if _are_runs_precious(job['monitoring']):
+        precious_runs_note = f"Note: Job {job_content['name']} is precious. Owners will be alerted if a day's, instead of the overall job's, latest run failed.\n"
```
solarkennedy: `Note: This alert is the run for 2018-10-04 (Friday). A resolve event will not occur until a job run for this date succeeds.`

How are we going to let people "skip" these?....

tron/bin/check_tron_jobs.py
```diff
@@ -115,23 +126,20 @@ def compute_check_result_for_job_runs(client, job, job_content):
...
+
     kwargs["output"] = (
-        "{}\n"
-        "{}'s last relevant run (run {}) {}.\n\n"
-        "Here is the last action:"
-        "{}\n\n"
+        f"{prefix}\n"
+        f"{job['name']}'s last relevant run (run {relevant_job_run['id']}) {relevant_job_run['state']}.\n"
```
solarkennedy: I would like to have the date here too if it is precious.

`name is the latest run for $date ($dom) {state}.`

tron/bin/check_tron_jobs.py
```diff
@@ -115,23 +126,20 @@ def compute_check_result_for_job_runs(client, job, job_content):
...
         "Here is the whole job view for context:\n"
-        "{}"
-    ).format(
-        prefix,
-        job['name'],
-        relevant_job_run['id'],
-        relevant_job_run['state'],
-        pretty_print_actions(action_run_details),
-        pretty_print_job_run(relevant_job_run),
-        pretty_print_job(job_content),
```
solarkennedy: Woot f-strings!

tron/bin/check_tron_jobs.py
```diff
@@ -293,41 +302,110 @@ def guess_realert_every(job):
...
 
-    status = job["status"]
-    if status == "disabled":
-        kwargs["output"] = "OK: {} is disabled and won't be checked.".format(
+    run_buckets = defaultdict(list)
+    if job_content is not None:
+        # Sort runs by shortdate
+        for run in job_content['runs']:
+            # If interval is invalid, will raise a KeyError
+            shortdate = time.strftime(
```
solarkennedy: If you are going to use the name "shortdate" as the variable name, then please remove the indirection of `interval_formats`

I realize you used a shorter variable for testing, but in practice we are never going to do anything other than a `shortdate`.

tron/bin/check_tron_jobs.py
```diff
@@ -293,41 +302,110 @@ def guess_realert_every(job):
...
+                interval_formats[interval],
+                _timestamp_to_timeobj(run['run_time']),
+            )
+            run_buckets[shortdate].append(run)
```
solarkennedy: Here we are missing a requirement that I neglected to voice.
I want logic that makes buckets for dates that don't even have runs yet, say for the past 7 days.

Right? Seems like if a run doesn't exist at all for a date, then we should alert on that too.

Kinda begs the question of "why do we only go backwards in time for as many days as there are jobruns that exist?

tron/bin/check_tron_jobs.py
```diff
@@ -293,41 +302,110 @@ def guess_realert_every(job):
...
-            job=job,
-            job_content=job_content,
-            client=client,
-        )
-        kwargs.update(results)
-        log.info(kwargs["output"].split("\n")[0])
-        return kwargs
+        ))
+
+        if _are_runs_precious(nonsensu_kwargs):
```
solarkennedy: Hehe. Using this as a dictionary "as if" it was monitoring.yaml feels a bit skecth, but I don't have any super great alternatives to recommend?

tron/bin/check_tron_jobs.py
```diff
@@ -293,41 +302,110 @@ def guess_realert_every(job):
...
+
+        for date, runs in dated_runs.items():
+            results = compute_check_result_for_job_runs(
+                job=job,
+                job_content=job_content.set('runs', runs),
+                client=client,
+            )
+            dated_kwargs = kwargs.update(results)
+            if date:  # if empty date, leave job name alone
+                dated_kwargs = dated_kwargs.set('name', f"{kwargs['name']}-{date}")
```
solarkennedy: This is an indirect way to do it.

Can you instead set the `name` in some if block where we are more explicitly doing a precious-runs branch?

tron/bin/check_tron_jobs.py
```diff
@@ -346,19 +424,21 @@ def check_job(job, client):
 
 
 def check_job_result(job, client, dry_run):
-    result = check_job(job, client)
-    if result is None:
+    results = check_job(job, client)
+    if not results:  # result is None or result == []
```
solarkennedy: I don't think you need this comment

solarkennedy:
This is looking super great. I really like how this looks with the "non-precious" logic branches as a kind of "trivial case" of the more general thing.

tron/bin/check_tron_jobs.py
```diff
@@ -293,41 +302,110 @@ def guess_realert_every(job):
...
+
+def compute_check_result_for_job(client, job):
+    kwargs = m(
+        name="check_tron_job.{}".format(job['name']),
+        source="tron",
+    )
+    if 'realert_every' not in kwargs:
+        kwargs = kwargs.set('realert_every', guess_realert_every(job))
+    kwargs = kwargs.set('check_every', "{}s".format(_run_interval))
+    sensu_kwargs, nonsensu_kwargs = filter_monitoring_config(job['monitoring'])
```
Rob-Johnson: I understand the pattern that you're going for, but there is a lot of indirection to deal with one non-standard attribute

at this point, you could just do

```
monitor_each_run = job['monitoring'].pop(PRECIOUS_JOB_ATTR, False)
kwargs.update(job['monitoring'])
....
if monitor_each_run:
    dated_runs = sort_runs_by_interval(job_content, interval='day')
...
```
and delete `filter_monitoring_config` and `_are_runs_precious`

sure, it might need reworking if the list of monitoring attributes that aren't sensu related grows, but we can tackle that when we come to it. For now I'd vote for the 'easy' option.

tron/bin/check_tron_jobs.py
chlgit: I think you can use something similar to guess_realert_every() to decide the number of days for going backward.

tests/bin/check_tron_jobs_test.py
```diff
@@ -1281,3 +1287,153 @@ def test_guess_realert_every_first_time_job(self):
...
+                'state': 'failed',
+                'exit_status': 1,
+            }, {
+                'id': f"{self.job_name}.4",
+                'job_name': self.job_name,
+                'run_num': 4,
+                'run_time': '2018-10-12 12:00:00',
+                'start_time': '2018-10-12 12:00:00',
+                'end_time': '2018-10-12 12:30:00',
+                'state': 'failed',
```
chlgit: Could you add the tests that the job run becomes unknown, or running longer than expected_runtime? I think the code can handle that, but it's nice to make sure that again.

tron/bin/check_tron_jobs.py
kawaiwanyelp: I agree with the easy approach; `filter_monitoring_config` and `_are_runs_precious` are easily removable.

tron/bin/check_tron_jobs.py
kawaiwanyelp: On @Rob-Johnson, I'm removing `_are_runs_precious`. Also, I don't have to use `nonsensu_kwargs`, I can still use `job['monitoring']`.

tron/bin/check_tron_jobs.py
kawaiwanyelp: Sure.

tron/bin/check_tron_jobs.py
kawaiwanyelp: I'll change the `shortdate` to something else, but I'd like to keep `interval_formats`, since it makes testing easier, now and in the future.

tron/bin/check_tron_jobs.py
kawaiwanyelp: Okay, removing it.

tron/bin/check_tron_jobs.py
kawaiwanyelp: I've decided to have buckets for all dates from the earliest run we have available to the current time. This way, we ensure that all dates have runs. If there are gaps in our run history (not just in the last 7 days), we will see them.

tron/bin/check_tron_jobs.py
```diff
@@ -164,7 +172,8 @@ def get_relevant_run_and_state(job_content):
     if len(job_runs) == 0:
         return None, State.NO_RUN_YET
     run = is_job_scheduled(job_runs)
-    if run is None:
+    # If runs are precious, then we do not care if there is a scheduled job
```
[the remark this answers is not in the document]
kawaiwanyelp: During testing, I found that if the latest run in a job is not scheduled, then the script will alert that the job has not been scheduled. This is not necessarily an error, as if a day already had a successful run, it does not make sense to schedule another for the same day. Moreover, scheduled runs only apply to the latest run in a job, not per day. 

However, it is an error if a day does NOT have a run whatsoever. That error we pick up elsewhere, before we enter this function.

I've updated the comment to be clearer about why I added the condition.

tests/bin/check_tron_jobs_test.py
kawaiwanyelp: Things to do with state live in `compute_check_result_for_job_runs`, which I have not meaningfully touched. After all, my changes aren't really concerned with state, just how we bucket runs. There are already plenty of tests for `compute_check_result_for_job_runs`, including for unknown states, so I don't think adding further tests for precious runs involving unknown is really necessary.

kawaiwanyelp:
Following @solarkennedy's comments, I've changed run bucketing so that all dates between the earliest available job run's date and the current date have their own buckets. This way, If any days don't have a run at all, owners will get an alert as well.

I also came across a bug where if a job was cancelled due to disabling a job, but not rerun when the job is reenabled, tron would think it was fine - "waiting for it's first run". I've added code to account for this state, as well as if a run is skipped. 

Lastly, I've removed some unnecessary functions and cleaned up things in other places according to @qui's, @robj's, and @chl's suggestions.

tron/bin/check_tron_jobs.py
```diff
@@ -293,41 +317,95 @@ def guess_realert_every(job):
     return realert_every
 
 
-def compute_check_result_for_job(client, job):
-    kwargs = {
-        "name": "check_tron_job.{}".format(job['name']),
-        "source": "tron",
+def sort_runs_by_interval(job_content, interval='day', until=None):
```
chlgit: I think I miss something here, but I can't find any place that until is not None.

tron/bin/check_tron_jobs.py
chlgit: Would it be better if we do this way instead:
    if job['monitoring'].get(PRECIOUS_JOB_ATTR, False):
        dated_runs = sort_runs_by_interval(job_content, interval='day')
        # compute_check_result_for_joj_runs for each run
    else:
        # do the old way

chlgit: Sorry, the format is messed up here. Let me know if you don't understand anything for my comment.

tron/bin/check_tron_jobs.py
kawaiwanyelp: On line 333, I check `until`. If it is set by the caller, then I do nothing. But if its still `None`, I set it to the current time, so we create buckets up until the current time.

chlgit: I mean... is there any caller that sets until as not None?

tron/bin/check_tron_jobs.py
```diff
@@ -99,9 +117,15 @@ def compute_check_result_for_job_runs(client, job, job_content):
     elif last_state == State.WAITING_FOR_FIRST_RUN:
         prefix = "OK: The job is 'new' and waiting for the first run"
         status = 0
+    elif last_state == State.SKIPPED:
+        prefix = "OK: The last job run was skipped"
+        status = 0
     elif last_state == State.STUCK:
         prefix = "WARN: Job exceeded expected runtime or still running when next job is scheduled"
         status = 1
+    elif last_state == State.CANCELLED:
```
qui: I think we still want to ignore cancelled runs. 
If a job was disabled on some day, they probably don't need an alert that it didn't run.
And for the non-precious case, if there's a cancelled run after a successful run, I think we should still consider the job OK.

tron/bin/check_tron_jobs.py
```diff
@@ -1,17 +1,24 @@
...
+from pyrsistent import m
+from pyrsistent import pmap
 from pysensu_yelp import send_event
 
 from tron.commands import cmd_utils
 from tron.commands import display
 from tron.commands.client import Client
 from tron.commands.client import get_object_type_from_identifier
 
+PRECIOUS_JOB_ATTR = 'check_that_every_day_has_a_successful_run'
```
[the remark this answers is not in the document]
kawaiwanyelp: Well, no one seems to mind how long it is LOL.

tron/bin/check_tron_jobs.py
kawaiwanyelp: I'm not sure it would be better to do it this way because the 'old way' is a trivial case of the 'new way'. In other words, the new way involves bucketing by day, while you can essentially treat the old way as putting everything in one bucket, a bucket of eternity. 

In any event, even if we put it into its own branch, the code would more or less be the same in the branch.

tron/bin/check_tron_jobs.py
solarkennedy: I agree

tron/bin/check_tron_jobs.py
```diff
@@ -293,41 +317,95 @@ def guess_realert_every(job):
...
+def compute_check_result_for_job(client, job):
+    kwargs = m(
+        name="check_tron_job.{}".format(job['name']),
+        source="tron",
+    )
     if 'realert_every' not in kwargs:
-        kwargs['realert_every'] = guess_realert_every(job)
-    kwargs['check_every'] = "{}s".format(_run_interval)
+        kwargs = kwargs.set('realert_every', guess_realert_every(job))
+    kwargs = kwargs.set('check_every', "{}s".format(_run_interval))
```
solarkennedy: Whoa, this is incorrect. We check every 5 minutes and this is "fixed"

kawaiwanyelp: Given that the intent is that we don't want to allow users to set 'check_every' in any monitoring config files, I'll need to remove that attribute from `sensu_kwargs` on the next line of code.

tron/bin/check_tron_jobs.py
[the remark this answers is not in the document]
chlgit: I like it shorter if it's possible. Also, the use cases are not only for every day. It can be any interval, right?

solarkennedy:
Can you add some docs too please?

tron/bin/check_tron_jobs.py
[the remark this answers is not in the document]
kawaiwanyelp: According to @solarkennedy, we will only allow a daily interval.

As for the long name, got any suggestions? @solarkennedy and I talked for a bit and we couldn't come up with a shorter name that wholly captured the intent of this attribute.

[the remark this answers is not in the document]
chlgit: I will use 'precious_check' and explain this configuration option in the doc [here](https://yelpwiki.yelpcorp.com/pages/viewpage.action?pageId=57614946). Same thing as we did for "realert_every". You can make the call.

tron/bin/check_tron_jobs.py
qui: Hold on, this will still be incorrect. Imagine one run is failed, and the next is cancelled. We want the check to fail.

What about changing WAITING_FIRST_RUN to something like NO_RUN_TO_CHECK: no job run to check ? I think that captures both the non-precious case (we just have a scheduled run, nothing to check yet) and the precious case (there's no run for that interval that we need to check, because the only one for that day is cancelled).

docs/jobs.rst
```diff
@@ -30,8 +30,64 @@ Optional Fields
...
+
+    **project** (default **None**)
+      A string representing the JIRA project that the ticket should go under.
+      Defaults to the team value.
+
+    **tags** (default **None**)
+      A list of arbitrary tags that can be used in handlers for different
+      metadata needs.
+
+    **ttl** (default **None**)
```
solarkennedy: TTL should not be something users can set

docs/jobs.rst
```diff
@@ -30,8 +30,64 @@ Optional Fields
...
+      Runbook associated with the job.
+
+    **tip** (default **None**)
+      A short 1-line version of the runbook.
+
+    **notification_email**
+      A comma-separated string of email destinations. Defaults to the "team"
+      default.
+
+    **irc_channels**
```
solarkennedy: don't bother mentioning irc

tron/bin/check_tron_jobs.py
kawaiwanyelp: Ah, okay. I agree with that. I'll change it.

PR #554, opened 2018-10-05
create pidfile for manhole socket

chlgit:
When we restarted Tron at prod yesterday, Tron crashed after running for 2 mins. And the error message from the log at /upstart was shown as follows:

https://fluffy.yelpcorp.com/i/CDhTDjD58zp2WmVSg9F3hZ3HlfLT4mNx.html

Setting wantPID = 1 should fix this issue.

### DIFF
--- a/tron/trondaemon.py
+++ b/tron/trondaemon.py
@@ -169,7 +169,7 @@ class TronDaemon(object):
 
     def _run_manhole(self):
         self.manhole = make_manhole(dict(trond=self, mcp=self.mcp))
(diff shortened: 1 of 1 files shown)


solarkennedy:
Have you tested this in the playground?

chlgit:
I did. The way I did was to restarting Tron with an existing PID file. By doing that, Tron would restart many times since a Tron pidfile is already present. Removing the pidfile with wantPID=0, it got the same "address already in use" error. However, removing the pidfile with wantPID=1, Tron was able to run.

PR #555, opened 2018-10-09
add tronctl move command

chlgit:
For example, if we want to move MASTER.test to test.test. the procedure would be:
  1. review for the change at yelpsoa-configs
  1. stop cron
  2. tronctl move MASTER.test test.test
  3. merge the change at yelpsoa-configs
  4. start cron

Need to test: Is it OK to let tron_deployment_tool to deploy the new change? I think so, but need to test it.

### DIFF
--- a/bin/tronctl
+++ b/bin/tronctl
@@ -146,7 +146,7 @@ def move(args):
     if old_name not in job_index.keys():
         raise SystemExit(f"Error: {old_name} doesn't exist")
     if new_name in job_index.keys():
-        raise SystemExit(f"Error: {new_name} has existed already")
+        raise SystemExit(f"Error: {new_name} exists already")
 
     return edit(args, '/api/jobs')
 
--- a/bin/tronctl
+++ b/bin/tronctl
@@ -34,7 +34,7 @@ COMMAND_HELP = (
     ('skip', 'Skip a failed action, runs dependent actions.'),
     ('stop', 'Stop the action run (SIGTERM)'),
     ('kill', 'Force kill the action run (SIGKILL)'),
-    ('move', 'Move a job to the new namespace'),
+    ('move', 'Rename a job'),
 )
 
 log = logging.getLogger('tronctl')
--- a/bin/tronctl
+++ b/bin/tronctl
@@ -34,6 +34,7 @@ COMMAND_HELP = (
     ('skip', 'Skip a failed action, runs dependent actions.'),
     ('stop', 'Stop the action run (SIGTERM)'),
     ('kill', 'Force kill the action run (SIGKILL)'),
+    ('move', 'Move a job to the new namespace'),
 )
 
 log = logging.getLogger('tronctl')
@@ -97,6 +98,9 @@ def edit(args, uri):
     data = {'command': args.command}
     if args.command == "start" and args.run_date:
         data['run_time'] = str(args.run_date)
+    elif args.command == "move":
+        data['old_name'] = args.id[0]
+        data['new_name'] = args.id[1]
 
     uri = urljoin(args.server, uri)
     response = client.request(uri, data=data)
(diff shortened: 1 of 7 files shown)


tron/core/job_scheduler.py
```diff
@@ -195,6 +195,12 @@ def get_runs_to_schedule(self, ignore_last_run_time):
         next_run_time = self.job.scheduler.next_run_time(last_run_time)
         return self.job.build_new_runs(next_run_time)
 
+    def update_name(self, name):
+        self.job.name = name
+        self.job.update_from_job(self.get_job())
```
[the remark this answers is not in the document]
chlgit: Ah ha, you're right. I've removed this one, but haven't pushed the updated code yet.

chlgit:
I am not sure if we need to change the names of the job runs. I like to keep it so I know where the job was moved from. But might it confuse others?

After restarting, the old job runs still show up in tronview because it had been removed from the configuration file. And yes, the new job runs show up along with the old ones.

chlgit:
Here is the example: 

$ tronview
Name                                          State         Scheduler             Last Success
test1.test                                    enabled     interval 0:00:10    None

$ tronview test1.test
.20       failed      chl@localhost                 2018-10-10 16:59:41
          Start: 2018-10-10 17:04:14  End: 2018-10-10 17:04:44  (0:00:30)
.19       failed      chl@localhost                 2018-10-10 16:59:31
          Start: 2018-10-10 17:03:44  End: 2018-10-10 17:04:14  (0:00:30)
.18       failed      chl@localhost                 2018-10-10 16:59:21

$ tronctl move test1.test test2.test
Moving test1.test to test2.test succeeded.

$ tronview
Name                                          State         Scheduler             Last Success
test2.test                                    enabled     interval 0:00:10    None

//sync configuration files, restart Tron
$ tronview
Name                                          State         Scheduler             Last Success
test2.test                                    enabled     interval 0:00:10    None

$ tronview test2.test
.21       failed      chl@localhost                 2018-10-10 16:59:51
          Start: 2018-10-10 17:04:44  End: 2018-10-10 17:05:14  (0:00:30)
.20       failed      chl@localhost                 2018-10-10 16:59:41
          Start: 2018-10-10 17:04:14  End: 2018-10-10 17:04:44  (0:00:30)
.19       failed      chl@localhost                 2018-10-10 16:59:31
          Start: 2018-10-10 17:03:44  End: 2018-10-10 17:04:14  (0:00:30)
.18       failed      chl@localhost                 2018-10-10 16:59:21

bin/tronctl
```diff
@@ -129,6 +133,24 @@ def control_objects(args):
...
+    try:
+        old_name = args.id[0]
+        new_name = args.id[1]
+    except IndexError as e:
+        raise SystemExit(f"Error: Move command needs two arguments.\n{e}")
+
+    tron_client = client.Client(args.server)
+    url_index = tron_client.index()
+    job_index = url_index['jobs']
+    if old_name not in job_index.keys():
```
qui: These checks should happen in the controller. If someone hits the API directly, they could bypass these checks and cause an error.

tron/core/job_scheduler.py
```diff
@@ -195,6 +195,11 @@ def get_runs_to_schedule(self, ignore_last_run_time):
         next_run_time = self.job.scheduler.next_run_time(last_run_time)
         return self.job.build_new_runs(next_run_time)
 
+    def update_name(self, name):
```
qui: I wonder if you need this. You could just set the name in JobCollection.move and call schedule_reconfigured afterwards.

chlgit: The reason I didn't call schedule_reconfigured is the variable, "ignore_last_run_time", should be False when we move the job. I can either keep update_name function here or refactor schedule_reconfigured function. No preference.

bin/tronctl
chlgit: OK, I will keep these checks here to report the error as early as possible. And I'll add the check at controller too.

bin/tronctl
```diff
@@ -34,6 +34,7 @@ COMMAND_HELP = (
     ('skip', 'Skip a failed action, runs dependent actions.'),
     ('stop', 'Stop the action run (SIGTERM)'),
     ('kill', 'Force kill the action run (SIGKILL)'),
+    ('move', 'Move a job to the new namespace'),
```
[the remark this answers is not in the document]
chlgit: Yeah, that sounds better.

PR #558, opened 2018-10-10
Triggers in tronweb

keymone:
Display waiting for and publishing triggers data for actionruns in tronweb.


solarkennedy:
Huh. Is it really that simple. There are no... data bindings?

keymone:
yeah, already did those in `tronview` PR

PR #559, opened 2018-10-11
Initial metrics endpoint

qui:
Need to add some tests, but otherwise ready for some comments.

Includes:
* timers for successful requests
* meters for all HTTP status codes
* meters for action run state changes
* counters for how many resources are being requested from Mesos

Example response: https://fluffy.yelpcorp.com/i/0VMHJDKh42PpNbkN3sFTbVJ8hZVffCx4.html

I would have to make some changes to taskproc for it to be able to use these gauges/timers instead of yelp_meteorite. Review soon! Either way, I think it's worth having these Tron metrics available. Planning to add a collector to put them into our system later.

### DIFF
--- a/tron/api/resource.py
+++ b/tron/api/resource.py
@@ -20,9 +20,12 @@ except ImportError:
 
 from twisted.web import http, resource, static, server
 
+from tron import __version__
 from tron.api import adapter, controller
 from tron.api import requestargs
 from tron.api.async_resource import AsyncResource
+from tron.metrics import view_all_metrics
+from tron.metrics import meter
 from tron.utils import maybe_decode
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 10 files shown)


tron/api/resource.py
```diff
@@ -395,7 +398,22 @@ def __init__(self, master_control):
...
-        return respond(request, {'status': "I'm alive."})
+        return respond(request, {
+            'status': "I'm alive.",
+            'version': __version__,
+        })
+
+
+class MetricsResource(resource.Resource):
+
+    isLeaf = True
```
chlgit: What is this for?

tron/core/jobrun.py
```diff
@@ -227,6 +228,7 @@ def handle_action_run_state_change(self, action_run: ActionRun, event):
 
         # propagate all state changes (from action runs) up to state serializer
         self.notify(self.NOTIFY_STATE_CHANGED)
+        metrics.meter(f'tron.actionrun.{event}')
```
solarkennedy: This is cool.
What do you think about actually putting individual namespace.job.action names in here?
We would explode the metrics to "a thousand or so", but it means we could action runtimes on our charts. Wouldn't that be cool?

tron/api/resource.py
qui: It's for twisted resources. If True, it means that there aren't any subpaths (e.g. /api/metrics/something_else), and twisted won't try to call getChild.

tron/core/jobrun.py
qui: I **could** put something in `_done` and emit the difference between start_time and end_time for each action. But one data point/day doesn't seem like the best fit for these metrics, which use exponentially decaying samples and stuff.

We can already query the API and get the most recent duration for an action. Maybe in the future we could add a different API endpoint that provides actual stats from the state file for jobs and actions.

PR #560, opened 2018-10-15
Prevent Pymesos thread abort from interrupting main thread

kawaiwanyelp:
### Description
- Problems:
    - In the event a pymesos process encounters an error, it will attempt to shutdown the main thread via `thread.interrupt_main` (which raises a `KeyboardInterrupt`)
    - @keymone, in #516, moved `reactor.run` and signal handlers around, which prevented us from properly handling signals ourselves.
        - Instead of explicitly setting signal handlers, he opted to use `signal.sigwait`, which blocks the main thread until a signal is received. However, if a `KeyboardInterrupt` is thrown (say, by pymesos), it is never translated to a `SIGINT` because the main thread is blocked. In some ways, this defeats the purpose of using `signal.sigwait`.
        - Although the intent was to prevent reactor shutdown before state changes complete and are saved, @qui and I found that, through testing, this wasn't true. **This will need to be addressed in a separate PR.**
- Solution:
    - Explicitly set signal handlers (as `DaemonContext` before it was deprecated) so signals can be properly handled (instead of waiting forever on `signal.sigwait`)
    - Add a meta-function that uses the previous SIGINT handler to set our own. To prevent pymesos from shutting down the main thread, we need to differentiate a SIGINT thrown by pymesos from one thrown when, say, a user does a Ctrl-C. By default, pymesos sets its own SIGINT handler, which can differentiate by checking if pymesos threw an exception that led to a SIGINT. In order to preserve our power to handle a SIGINT, my new meta-function generates a handler that calls a 'previous handler' (in this case, pymesos' handler) to see if it thinks the SIGINT was caused by a user-generated `KeyboardInterrupt` or some internal exception from another thread. If it is the former, we shutdown; otherwise, we simply log and continue on.

### Testing
- `tox -e trond_inside_container`: In a simulated, dockerized cluster, caused an error that triggered a `thread.interrupt_main` from pymesos.
- `make test` (including some new tests to test the new meta-function)

### DIFF
--- a/tests/trondaemon_test.py
+++ b/tests/trondaemon_test.py
@@ -3,6 +3,7 @@ from __future__ import unicode_literals
 
 import os
 import tempfile
+from collections import defaultdict
 
 import lockfile
 import mock
@@ -78,6 +79,29 @@ class TronDaemonTestCase(TestCase):
     @teardown
     def teardown(self):
         self.tmpdir.cleanup()
+
+    @mock.patch('tron.trondaemon.PIDFile', mock.Mock(), autospec=None)
+    def test_init(self):
+        daemon = TronDaemon.__new__(TronDaemon)  # skip __init__
+        daemon._make_sigint_handler = mock.Mock()
+        options = mock.Mock()
+        _sig_handlers = defaultdict(lambda: 'original_handler')
+
+        with mock.patch(
+            'signal.getsignal',
+            mock.Mock(side_effect=_sig_handlers.__getitem__),
+            autospec=None
+        ), mock.patch(
+            'signal.signal',
+            mock.Mock(side_effect=_sig_handlers.__setitem__),
+            autospec=None
+        ):
+            daemon.__init__(options)
+
+            assert daemon._make_sigint_handler.call_count == 1
+            assert daemon._make_sigint_handler.call_args == mock.call(
+                'original_handler'
+            )
 
     def test_make_sigint_handler_keyboardinterrupt(self):
         daemon = TronDaemon.__new__(TronDaemon)  # skip __init__
--- a/tests/trondaemon_test.py
+++ b/tests/trondaemon_test.py
@@ -78,3 +78,30 @@ class TronDaemonTestCase(TestCase):
     @teardown
     def teardown(self):
         self.tmpdir.cleanup()
+
+    def test_make_sigint_handler_keyboardinterrupt(self):
+        daemon = TronDaemon.__new__(TronDaemon)  # skip __init__
+        daemon._handle_shutdown = mock.Mock()
+
+        def prev_handler(signum, frame):
+            raise KeyboardInterrupt
+
+        handler = daemon._make_sigint_handler(prev_handler)
+        handler('fake_signum', 'fake_frame')
+
+        daemon._handle_shutdown.call_args == mock.call(
+            'fake_signum',
+            'fake_frame',
+        )
+
+    def test_make_sigint_handler_exception(self):
+        daemon = TronDaemon.__new__(TronDaemon)  # skip __init__
+        daemon._handle_shutdown = mock.Mock()
+
+        def prev_handler(signum, frame):
+            raise Exception
+
+        handler = daemon._make_sigint_handler(prev_handler)
+        handler('fake_signum', 'fake_frame')
+
+        daemon._handle_shutdown.call_count == 0
--- a/tron/trondaemon.py
+++ b/tron/trondaemon.py
@@ -9,7 +9,6 @@ import logging
 import logging.config
 import os
 import signal
-import threading
 import time
 
 import ipdb
@@ -126,6 +125,9 @@ class NoDaemonContext(object):
         self.pidfile = kwargs.pop('pidfile', None)
         self.working_dir = kwargs.pop('working_directory', '.')
 
+        self.signal_map = kwargs.pop('signal_map', {})
+        self._set_signal_handlers(self.signal_map)
+
     def __enter__(self):
         os.chdir(self.working_dir)
         if self.pidfile:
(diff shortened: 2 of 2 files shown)


tests/trondaemon_test.py
```diff
@@ -78,3 +78,30 @@ def setup(self):
     @teardown
     def teardown(self):
         self.tmpdir.cleanup()
+
+    def test_make_sigint_handler_keyboardinterrupt(self):
+        daemon = TronDaemon.__new__(TronDaemon)  # skip __init__
```
qui: You can use @mock.patch.object(TronDaemon, _handle_shutdown) to mock out the method instead. Then you should be able to use self.trond in the test.

(same for the test below)

kawaiwanyelp:
Had to do a forced re-push because when I branched, @chlgit hadn't yet fixed the changelog, which was causing tests to fail.

tron/trondaemon.py
```diff
@@ -149,13 +156,48 @@ def __init__(self, options):
...
+                    prev_handler(signum, frame)
+            except KeyboardInterrupt:
+                # Previous signal handler didn't raise another exception,
+                # so must be user requesting shutdown.
+                pass
+            except Exception as e:
+                # We received a SIGINT, but was caused by another thread
+                # aborting due to its own error. In this case, we don't want to
+                # stop running.
+                log.info(f"Non-reactor thread raised: {e}")
```
qui: I think this should be log.error. At *least* warning-level.

One thought - what do people think about disabling Mesos with MesosClusterRepository.shutdown here? It would be nice that Mesos actions would automatically start failing, instead of getting stuck in "starting". On the other hand, that assumes that these interrupts are only coming from pymesos.

tests/trondaemon_test.py
kawaiwanyelp: But part of creating `self.trond` is that it uses `_make_sigint_handler` during initialization. Doesn't it go against the spirit of testing to use an object to test something we have already used without testing?

tron/trondaemon.py
kawaiwanyelp: I'm not sure about disabling Mesos here precisely because, as you mentioned, we assume the interrupts come only from pymesos, @qui. In practice, that's true because its the only other place (we know of) that sets its own signal handler, but it doesn't preclude it from happening in the future (even though it really shouldn't be allowed).

qui: But in the future, if some other library calls interrupt_main, will they even necessarily set their own handler? I think we're already making some implicit assumptions about pymesos. (We could add more comments to make our assumptions clear.)

tron/trondaemon.py
```diff
@@ -149,13 +156,48 @@ def __init__(self, options):
...
+        :param prev_handler: The previous SIGINT handler, set by another source.
+                             We use it to verify whether or not a SIGINT we
+                             received is a genuine shutdown request.
+        """
+        def handler(signum, frame):
+            try:
+                if prev_handler is not None:
+                    prev_handler(signum, frame)
+            except KeyboardInterrupt:
+                # Previous signal handler didn't raise another exception,
```
Rob-Johnson: rather than try/catch and having to know the internals of the default handler, what do you think of checking if ``prev_handler == signal.default_int_handler``?

keymone:
personally i dislike going back to signal handlers for few reasons:

- they are executed asynchronously on a separate thread(s)
- they run in parallel with other signal handlers which means you have to be careful to coordinate the side effects
- handler is invoked every time signal  comes in even if previous invokation didn't finish, again coordination
- we rely on faith that no other code will replace our handler (bad idea)

this PR removes opting out of twisted reactor's signal handlers (the `kwargs=dict(installSignalHandlers=0)` line) - was that intentional and do we know if reactor's handlers play nice with pymesos' handlers and our handlers?

sigwait loop approach solves all of these and makes processing signals safe and easy to understand

but the real problem we need to address right now is that libraries  (pymesos, twisted, etc) *have no business managing our signals* unless explicitly asked to do so (with twisted we can at least opt out).

we own the app, it's app's responsibility to handle signals.

tron/trondaemon.py
```diff
@@ -195,27 +211,66 @@ def _run_mcp(self):
...
+                # so must be user requesting shutdown.
+                pass
+            except Exception as e:
+                # We received a SIGINT, but was caused by another thread
+                # aborting due to its own error. In this case, we don't want to
+                # stop running.
+                log.info(f"Non-reactor thread raised: {e}")
```
qui: Can you make this an error?

tron/trondaemon.py
```diff
@@ -146,6 +154,8 @@ class TronDaemon(object):
     def __init__(self, options):
         self.options = options
         self.mcp = None
+        self._sigint_handler = self._make_sigint_handler(
```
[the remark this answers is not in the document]
kawaiwanyelp: Ah, you're right. I actually ran into this issue before I put up the PR. I'll update the tests.

tron/trondaemon.py
```diff
@@ -155,7 +165,13 @@ def _build_context(self, options):
             working_directory=options.working_dir,
             umask=0o022,
             pidfile=pidfile,
-            signal_map={},
+            signal_map={
```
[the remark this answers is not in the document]
kawaiwanyelp: @qui, we do only need to reset the `SIGINT` handler, but its actually not true that setting a default handler is different. The reason is that by default, all signals have handlers, namely `SIG_DFL` (except `SIGINT`, which has `default_int_handler` instead). You can actually check this by running `signal.getsignal` for any signal you want in a Python interpreter. 

I decided to set all of them to default for brevity, because in the event anyone else attempts to set things (which shouldn't happen but you never know), it shouldn't mess with us weirdly.

tron/trondaemon.py
```diff
@@ -195,27 +211,66 @@ def _run_mcp(self):
...
+                # KeyboardInterrupt will be thrown. A sigwait will not unblock,
+                # but a sigtimedwait will.
+                signum = signal.sigtimedwait(set(signal_map.keys()), 0)
+                if signum is not None:
+                    signum = signal.Signals(signum.si_signo)
+            except KeyboardInterrupt:
+                signum = signal.SIGINT
+
             if signum in signal_map:
+                logging.info(f"Got signal {str(signum)}")
```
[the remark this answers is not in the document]
kawaiwanyelp: @qui, fair. That's the way it was before, so I can restore that.

tron/trondaemon.py
kawaiwanyelp: Oops! I thought I had already done this. I'll make sure it happens this time, @qui.

PR #564, opened 2018-10-17
Tronctl publish/discard events

keymone:
Add API resource and controller for publishing/discarding events. Add tronctl subcommands.

I've also moved around some code in tronctl around to keep related bits together and cleaned up  tron.api.client a bit.


tron/api/resource.py
```diff
@@ -424,6 +424,41 @@ def render_GET(self, request):
...
 
 
+class EventsResource(resource.Resource):
+    isLeaf = True
+
+    def __init__(self):
+        super().__init__()
+        self.controller = controller.EventsController()
+
+    @AsyncResource.exclusive
```
qui: Is this right? Only one client can read the event log at a time, while multiple clients can publish/discard?

tron/api/controller.py
```diff
@@ -197,3 +198,40 @@ def delete_config(self, name, content, config_hash):
...
+
+        if not EventBus.publish(event):
+            msg = f'could not publish {event}'
+            log.error(msg)
+            return dict(error=msg)
+
+        return dict(response='OK')
+
+    def discard(self, event):
+        if not EventBus.instance:
```
chlgit: Do you really want to check EventBus.instance twice, at here and eventbus.py?

bin/tronview
```diff
@@ -60,6 +60,14 @@ def parse_cli():
         help="Solely displays stderr",
         default=0,
     )
+    parser.add_argument(
+        "--events",
+        "-E",
+        action="count",
```
qui: Are you planning on adding a count argument to the events GET endpoint later?

keymone: no, i  just copied the definition from other args, this should really be store_true or something.

tron/api/controller.py
keymone: this check isn't necessary for discard to function correctly, but i want to respond with correct error message in case eventbus is disabled.

PR #566, opened 2018-10-19
If runs are allowed to overlap, don't consider that case stuck

qui:


solarkennedy:
Thanks for making our customers happy :)

PR #567, opened 2018-10-19
Fix check_tron_jobs to discard instead of remove precioius attr

kawaiwanyelp:
### Description
- Fixing a small bug in check_tron_jobs.py where I wrote it to `remove` instead of `discard` the precious attribute when converting the monitoring fields to Sensu args.
- Added a test to catch it.

### Tests
- make dev (and run the patched script on a test job)
- make test

### DIFF
--- a/tests/bin/check_tron_jobs_test.py
+++ b/tests/bin/check_tron_jobs_test.py
@@ -1404,6 +1404,24 @@ class TestCheckPreciousJobs(TestCase):
         assert run_buckets['2018.10.14'] == []
 
     @patch('check_tron_jobs.Client', autospec=True)
+    def test_compute_check_result_for_job_not_precious(self, mock_client):
+        client = mock_client('fake_server')
+        del self.job['monitoring'][check_tron_jobs.PRECIOUS_JOB_ATTR]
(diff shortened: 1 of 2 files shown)


tests/bin/check_tron_jobs_test.py
```diff
@@ -1403,6 +1403,23 @@ def test_sort_runs_by_interval_day_empty_buckets(self):
...
 
+    @patch('check_tron_jobs.Client', autospec=True)
+    def test_compute_check_result_for_job_not_precious(self, mock_client):
+        client = mock_client('fake_server')
+        del self.job['monitoring'][check_tron_jobs.PRECIOUS_JOB_ATTR]
+        check_tron_jobs.guess_realert_every = mock.Mock(return_value=1)
+        check_tron_jobs.get_object_type_from_identifier = \
+            mock.Mock(return_value=mock.Mock())
+        client.job = mock.Mock(return_value=self.job)
+        check_tron_jobs.compute_check_result_for_job_runs = mock.Mock(
```
qui: Can you also assert that compute_check_result_for_job_runs is only called once in the non-precious case?

kawaiwanyelp: Sure.

PR #568, opened 2018-10-19
add migration namespace script

chlgit:
The push plan:
  -  make a branch and do the yelpsoa-configs change
  -  get ship it from the team
  -  run this script, something like "./migration_script.py /nail/home/chl/yelpsoa-configs/tron/mesosstage/test2.yaml --server=http://tronplayground-uswest1adevc.dev.yelpcorp.com:8089 --old-ns=test2  --new-ns=test"
  -  the script should wait with the message "Merge and push yelpsoa-configs branch. Ready to continue? [y/n]"
  - merge yelpsoa-config branch
  - enter 'y' to continue this script

This script does these things:
  - read the source file to get the list of the jobs
  - check if all jobs are not running
  - disable all jobs
  - stop cron
  - wait for the yelpsoa-configs merge
  - rsync yelpsoa-configs
  - tronctl move each job
  - update the new namespace
  - update the old namespace
  - start cron

### DIFF
--- a/contrib/migration_script.py
+++ b/contrib/migration_script.py
@@ -161,6 +161,9 @@ def main():
         # start cron
         ssh_command(hostname, "sudo service cron start")
 
+        #clean up namespace
+        ssh_command(hostname, "sudo paasta_cleanup_tron_namespaces")
+
         print(bcolors.OKBLUE + "Jobs migration are done. Enable all jobs" + bcolors.ENDC)
         if command_jobs('enable', jobs, args, ns=args.new_ns) is True:
             print(bcolors.OKBLUE + "Jobs are running at namespace {}".format(args.new_ns) + bcolors.ENDC)
--- a/contrib/migration_script.py
+++ b/contrib/migration_script.py
@@ -31,19 +31,16 @@ def parse_args():
     parser.add_argument(
         '--server',
         required=True,
-        # default='http://tronplayground-uswest1adevc.dev.yelpcorp.com:8089',
         help='specify the location of tron master'
     )
     parser.add_argument(
         '--old-ns',
         required=True,
-        # default='dsc-test-service',
         help='Old namespace'
     )
     parser.add_argument(
         '--new-ns',
         required=True,
-        # default='test_namespace',
         help='New namespace'
     )
     parser.add_argument(
@@ -69,12 +66,18 @@ def check_job_if_running(jobs_status, job_name):
     return False
 
 
-def command_jobs(command, jobs, args):
+def command_jobs(command, jobs, args, ns=None):
+    """ This function run tronctl command for the jobs
+    command: the tronctl command it will run
+    jobs: a list of jobs
+    args: the args for this script
+    ns: the namespace to use as the prefix for each job, if None, the scrip would use args.old_ns instead
+    """
     data = {'command': command}
     command_flag = True
     for job in jobs:
-        if command == 'enable':
-            job_name = args.new_ns + '.' + job['name']
+        if ns is not None:
+            job_name = ns + '.' + job['name']
         else:
             job_name = args.old_ns + '.' + job['name']
 
@@ -95,11 +98,14 @@ def command_jobs(command, jobs, args):
 
 
 def ssh_command(hostname, command):
+    print(bcolors.BOLD + 'Executing the command: ssh -A {} {}'.format(hostname, command) + bcolors.ENDC)
     ssh = subprocess.Popen(["ssh", "-A", hostname, command], shell=False, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    exitcode = ssh.wait()
     result = ssh.stdout.readlines()
     error = ssh.stderr.readlines()
-    if error != []:
-        print(bcolors.BOLD + 'Execute command {}: {}'.format(command, error) + bcolors.ENDC)
+    if exitcode != 0:
+        print(bcolors.FAIL + 'Execute command {} failed: {}'.format(command, error) + bcolors.ENDC)
+        exit(exitcode)
     return result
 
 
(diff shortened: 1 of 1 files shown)


contrib/migration_script.py
```diff
@@ -0,0 +1,170 @@
...
+
+def main():
+    args = parse_args()
+    filename = args.source
+    hostname = urlparse(args.server).hostname
+    if filename.endswith(".yaml"):
+        tron_client = client.Client(args.server)
+        jobs_status = tron_client.jobs()
+
+        migration_flag = True
```
solarkennedy: maybe `is_migration_safe` ?

contrib/migration_script.py
```diff
@@ -0,0 +1,170 @@
...
+    if error != []:
+        print(bcolors.BOLD + 'Execute command {}: {}'.format(command, error) + bcolors.ENDC)
+    return result
+
+
+def main():
+    args = parse_args()
+    filename = args.source
+    hostname = urlparse(args.server).hostname
+    if filename.endswith(".yaml"):
```
solarkennedy: For this bulk migration, working at a whole file at a time is good and efficient. In the long term we will probably need a more fine-grained way to do this (per-job) because MASTER will basically never been ready to move (always some job will be running probably)

contrib/migration_script.py
```diff
@@ -0,0 +1,170 @@
...
+                print(bcolors.OKBLUE + "Recovery successfully: Jobs are enabled again." + bcolors.ENDC)
+            else:
+                print(bcolors.WARNING + "Recovery failed: Please manually check the status of jobs" + bcolors.ENDC)
+            return
+
+        # stop cron
+        ssh_command(hostname, "sudo service cron stop")
+
+        # wait unitil yelpsoa-configs branch is merged
+        res = input("Merge and push yelpsoa-configs branch. Ready to continue? [y/n]")
```
solarkennedy: Is a `git mv` implied here? If so can you print out the exact git mv command to reduce mistakes?

contrib/migration_script.py
```diff
@@ -0,0 +1,170 @@
...
+        if res == 'y':
+            # wait for 10 seconds after pushing the branch
+            time.sleep(10)
```
solarkennedy: do 30

contrib/migration_script.py
```diff
@@ -0,0 +1,170 @@
...
+        response = client.request(uri, data=data)
+        if response.error:
+            print(bcolors.FAIL + 'Failed to {} {}'.format(command, job_name) + bcolors.ENDC)
+            command_flag = False
+        else:
+            print(bcolors.OKGREEN + 'Succeed to {} {}'.format(command, job_name) + bcolors.ENDC)
+    return command_flag
+
+
+def ssh_command(hostname, command):
```
solarkennedy: Printing out the command that is running will be helpful for debugging and recovery.

This script is for us, it is OK to be verbose

contrib/migration_script.py
```diff
@@ -0,0 +1,170 @@
...
+    parser.add_argument(
+        '--old-ns',
+        required=True,
+        # default='dsc-test-service',
+        help='Old namespace'
+    )
+    parser.add_argument(
+        '--new-ns',
+        required=True,
+        # default='test_namespace',
```
solarkennedy: go ahead and remove these defaults

contrib/migration_script.py
```diff
@@ -0,0 +1,170 @@
...
+        help='New namespace'
+    )
+    parser.add_argument(
+        'source',
```
solarkennedy: When would this ever not be `$old_ns.yaml`?

contrib/migration_script.py
```diff
@@ -0,0 +1,170 @@
...
+    args = parse_args()
+    filename = args.source
+    hostname = urlparse(args.server).hostname
+    if filename.endswith(".yaml"):
+        tron_client = client.Client(args.server)
+        jobs_status = tron_client.jobs()
+
+        migration_flag = True
+        with open(filename, "r") as f:
+            jobs = yaml.load(f)['jobs']
```
qui: To make this easier to run on partial namespaces (or single jobs), can we extract out the rest of the code into a separate function, that takes a list of jobs?
(could do in a follow-up, if you want to start on full namespaces now)

contrib/migration_script.py
```diff
@@ -0,0 +1,170 @@
...
+            print(bcolors.OKGREEN + 'Succeed to {} {}'.format(command, job_name) + bcolors.ENDC)
+    return command_flag
+
+
+def ssh_command(hostname, command):
+    ssh = subprocess.Popen(["ssh", "-A", hostname, command], shell=False, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    result = ssh.stdout.readlines()
+    error = ssh.stderr.readlines()
+    if error != []:
+        print(bcolors.BOLD + 'Execute command {}: {}'.format(command, error) + bcolors.ENDC)
```
qui: I think the script should exit if there are any errors in the SSH commands.

contrib/migration_script.py
chlgit: I was thinking "git merge --no-ff [branch]" and "git push origin master".

contrib/migration_script.py
chlgit: The $source should have the full path, while $old_ns is only the name of the namespace. I was thinking the path can be 'yelpsoa-configs/namespace' or  'yelpsoa-configs/tron/norcal-prod', although I am not sure if we have use cases for the first one.

contrib/migration_script.py
```diff
@@ -0,0 +1,170 @@
...
+            print(bcolors.OKBLUE + "Jobs are not running. Disable all the jobs." + bcolors.ENDC)
+        else:
+            print(bcolors.WARNING + "Some jobs are still running, abort this migration," + bcolors.ENDC)
+            return
+
+        if command_jobs('disable', jobs, args) is True:
+            print(bcolors.OKBLUE + "Jobs disabled" + bcolors.ENDC)
+        else:
+            print(bcolors.WARNING + "Some jobs aren't disabled, starting recovery" + bcolors.ENDC)
+            if command_jobs('enable', jobs, args) is True:
```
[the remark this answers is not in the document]
chlgit: Ah, you're right.

contrib/migration_script.py
chlgit: Roger that

contrib/migration_script.py
chlgit: Yeah, will do in a follow-up PR.

contrib/migration_script.py
chlgit: Unfortunately, the some messages would show up in stderr rather than stdout. For example, paasta_setup_tron_namespace would output this message "[b"INFO:paasta_tools.setup_tron_namespace:Updated following namespaces: ['test'], failed: [], skipped: 0\n"]" in stderr.

chlgit: Maybe we should check the return code instead?

chlgit: Yay, return code works better.

PR #570, opened 2018-10-24
Pin and cleanup dependencies

keymone:
- move dependencies from requriements.txt to setup.py (use `-e .` trick instead)
- pin to specific versions
- remove unused deps: enum34, python-daemon, yelp-clog
- remove  dev/requirements.txt and move dev related deps to tox.ini


qui:
We reverted this because it caused `Object of type 'bytes' is not JSON serializable` when you try to view an action run (e.g. `tronview MASTER.test.1.run`).

PR #571, opened 2018-10-24
Add get_tron_metrics script and tests

kawaiwanyelp:
### Description
- In #559, @qui added a generic metrics endpoint. This script pulls metrics from that endpoint and sends them to meteorite via its bash interface (so we don't need to import it). 
- The script is intended to be run as a cron job.

### Testing
- manual testing with bogus metrics + dry-runs
- make test

### Notes to reviewers
- Should I be adding docs for this? I'm thinking not since this is rather Yelp specific.
- Too many tests? I just added that amount just in case things change later on.
- pre-commit was pinned because it was, at the time of writing, changed recently to use importlib_resources but there aren't wheels available for them yet.

### DIFF
--- a/tron/bin/get_tron_metrics.py
+++ b/tron/bin/get_tron_metrics.py
@@ -1,0 +1,158 @@
+#!/usr/bin/env python3.6
+#
+# get_tron_metrics.py
+#   This script is designed to retrieve metrics from Tron via its API and send
+#   send them to meteorite.
+import logging
+import pprint
+import subprocess
+import sys
+import textwrap
+
+from tron.commands import cmd_utils
+from tron.commands.client import Client
+
+log = logging.getLogger('get_tron_metrics')
+
+
+def parse_cli():
+    parser = cmd_utils.build_option_parser()
+    parser.description = (
+        "Collects metrics from Tron via its API and forwards them to "
+        "meteorite."
+    )
+    parser.add_argument(
+        "--dry-run",
+        action="store_true",
+        default=False,
+        help="Don't actually send metrics out. Defaults: %(default)s"
+    )
+    args = parser.parse_args()
+    return args
+
+
+def check_bin_exists(bin):
+    """
+    Checks if an executable binary exists
+
+    :param bin: (str) Name of the executable; could be a path to one
+    """
+    try:
+        subprocess.call(['which', bin])
+    except OSError:
+        return False
+    return True
+
+
+def send_data_metric(
+    name, metric_type, value,
+    dimensions={}, dry_run=False, **kwargs
+):
+    """
+    Sends a single data point to meteorite via bash command
+
+    :param name: (str) Name of the metric
+    :param metric_type: (str) Type of the meteorite metric. Must be in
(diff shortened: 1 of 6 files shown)


tron/bin/get_tron_metrics.py
```diff
@@ -0,0 +1,158 @@
...
+    send_data_metric(name, 'gauge', kwargs.pop('value'), **kwargs)
+
+
+def send_meter(name, **kwargs):
+    send_counter(name, **kwargs)  # We ignore mX_rate args
+
+
+def send_histogram(name, **kwargs):
+    for k in ['p50', 'p75', 'p95', 'p99']:  # Only send p50-99
+        gauge_name = f"{name}-{k}"
```
qui: The convention is to use '.' to separate parts of the metric name.

tron/bin/get_tron_metrics.py
```diff
@@ -0,0 +1,158 @@
...
+        log.error(
+            "Meteorite failed with:\n"
+            f"{textwrap.indent(error, '    ')}"
+        )
+    else:
+        log.debug(f"From meteorite: {output}")
+
+
+def send_counter(name, **kwargs):
+    send_data_metric(name, 'counter', kwargs.pop('count'), **kwargs)
```
qui: When you're actually calling send_data_metric, can you specify the kwargs, instead of passing everything you get with **kwargs? Also use the keyword names, like metric_type='counter'.

kawaiwanyelp: Will change.

tron/bin/get_tron_metrics.py
```diff
@@ -0,0 +1,158 @@
...
+    # We mirror the metrics implementation in Tron by splitting timer into a
+    # meter and a histogram
+    send_meter(name, **kwargs)
+    send_histogram(name, **kwargs)
+
+
+_METRIC_SENDERS = {
+    'counter': send_counter,
+    'gauge': send_gauge,
+    'meter': send_meter,
```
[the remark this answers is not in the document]
kawaiwanyelp: I like moving `_METRIC_SENDERS` to the top of the file, but I would prefer not to remove `send_meter` since the script should mirror what happens in the metric endpoint, at the very least from an organizational standpoint. Namely, timers are expressed as a combination of meters and histograms.

tron/bin/get_tron_metrics.py
```diff
@@ -0,0 +1,167 @@
...
+    process = subprocess.Popen(
+        cmd,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+    )
+    output, error = process.communicate()
+    output = output.decode('utf-8').rstrip()
+    error = error.decode('utf-8').rstrip()
+
+    if process.returncode:
```
solarkennedy: can you put `!= 0` here? I know normally this is very pythonic, but with unix return codes I think it is a bit better to say what you mean

kawaiwanyelp: Sure.

PR #572, opened 2018-10-25
Combine DST code into one place

qui:
This changes existing behavior in one way: if you have a timezone set, the trontimespec used to always schedule for the second hour in the ambiguous case. Now, it will schedule for the first.

I tested this pretty extensively on playground by changing the system clock. Now in all cases, whether or not there is a timezone set, jobs will run in the first 1AM hour, and not the second 1AM.

e.g. on November 4th, a job that runs every 30 min would run at:
- [x] 00:30 PDT
- [x] 01:00 PDT
- [x] 01:30 PDT
- [ ] 01:00 PST
- [ ] 01:30 PST
- [x] 02:00 PST

### DIFF
--- a/tron/utils/trontimespec.py
+++ b/tron/utils/trontimespec.py
@@ -85,6 +85,21 @@ def to_timezone(t, tzinfo):
         return t
 
 
+def naive_as_timezone(t, tzinfo):
+    """Interprets the naive datetime with the given time zone."""
+    try:
+        result = tzinfo.localize(t, is_dst=None)
+    except AmbiguousTimeError:
+        # We are in the infamous 1 AM block which happens twice on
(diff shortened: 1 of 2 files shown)


tron/utils/trontimespec.py
```diff
@@ -282,17 +297,7 @@ def get_first_day(month, year):
     # TODO: test
```
solarkennedy: TODONE?

solarkennedy:
@dnephin will be proud! (hopefully 😄)

tron/utils/trontimespec.py
qui: eh, no test actually calls handle_timezone, I'm going to leave it.

dnephin:
Nice!

PR #573, opened 2018-10-25
Allow tron to run after unclean shutdown

kawaiwanyelp:
### Description
- In the event of an unclean shutdown, Tron's pidfile and manhole socket could be left behind in the working directory, preventing another Tron daemon running, despite there being no existing daemon. However, these changes will allow us ignore those orphaned files and start Tron regardless.
- `tron.pid`:
    - If we find a pid file and verify no process is running with that pid, instead of exiting, we now simply clear it out and continue on.
    - We do not delete it, as we have already acquired a lock on it.
- `manhole.sock`:
    - If `manhole.sock.lock` was also left behind after an unclean shutdown, Tron is already able to reclaim the lock and start with no problems.
    - However, if the lock file was removed but the socket file remains, Tron would previously refuse to start. Now, the Tron daemon will simply delete the old manhole socket and continue on.

### Testing
- manual testing with `make dev`
- `make test`

### Notes to reviewers
- I also moved the call to `setup_logging` to the `TronDaemon` initialization because otherwise we wouldn't be able to properly log before actually running the daemon, like, say, during setup.

### DIFF
--- a/debian/tron.service
+++ b/debian/tron.service
@@ -5,7 +5,7 @@ After=network.target
 [Service]
 User=tron
 EnvironmentFile=/etc/default/tron
-ExecStart=/usr/bin/trond --pid-file=${PIDFILE} --working-dir=${WORKINGDIR} -H 0.0.0.0 -P 8089 -l ${LOGCONF}
+ExecStart=/usr/bin/trond --lock-file=${LOCKFILE:-$PIDFILE} --working-dir=${WORKING+DIR} -H 0.0.0.0 -P 8089 -l ${LOGCONF}
 TimeoutStopSec=20
 Restart=on-failure
 
--- a/debian/tron.service
+++ b/debian/tron.service
@@ -5,7 +5,7 @@ After=network.target
 [Service]
 User=tron
 EnvironmentFile=/etc/default/tron
-ExecStart=/usr/bin/trond --lock-file=${LOCKFILE:-$PIDFILE} --working-dir=${WORKING+DIR} -H 0.0.0.0 -P 8089 -l ${LOGCONF}
+ExecStart=/usr/bin/trond --lock-file=${LOCKFILE:-$PIDFILE} --working-dir=${WORKINGDIR} -H 0.0.0.0 -P 8089 -l ${LOGCONF}
 TimeoutStopSec=20
 Restart=on-failure
 
--- a/tests/utils/flockfile_test.py
+++ b/tests/utils/flockfile_test.py
@@ -1,0 +1,68 @@
+from __future__ import absolute_import
+from __future__ import unicode_literals
+
+import os
+import tempfile
+
+import mock
+import pytest
+
+from tron.utils.flockfile import FlockFile
+
+
+@pytest.fixture
+def tmp_lockfile():
+    tmpdir = tempfile.TemporaryDirectory()
+    yield os.path.join(tmpdir.name, "test.lock")
+    tmpdir.cleanup()
+
+
+@pytest.fixture
+def mock_flockfile(tmp_lockfile):
+    return FlockFile(tmp_lockfile)
+
+
+@pytest.mark.parametrize('is_locked', [True, False])
+def test_is_locked(mock_flockfile, is_locked):
+    mock_flockfile._has_lock = is_locked
+
+    assert mock_flockfile.is_locked == is_locked
+
+
+@mock.patch('fcntl.flock', mock.Mock(side_effect=OSError), autospec=None)
+def test_is_locked_by_other(mock_flockfile):
+    mock_flockfile._has_lock = False
+
+    assert mock_flockfile.is_locked
+
+
+def test_acquire(mock_flockfile):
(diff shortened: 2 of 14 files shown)


keymone:
> In the event of an unclean shutdown, Tron's pidfile and manhole socket could be left behind in the working directory, preventing another Tron daemon running, despite there being no existing daemon. However, these changes will allow us ignore those orphaned files and start Tron regardless.

does it make sense having pidfiles at all then?

kawaiwanyelp:
@keymone, we want to keep pidfiles as a way of ensuring only one Tron daemon is running. If one is left behind, we can use it to check if another daemon is running; we only start the daemon once we verify that another is not running.

qui:
I think we should get rid of the PID file entirely, and modify the flock file to implement the behavior we want. Killing whatever process ID is in the PID file is possibly dangerous because PIDs can be reused, and it's unnecessary.

kawaiwanyelp:
I've replaced most the pidfile stuff with lockfile stuff everywhere; that includes the docs. However, I've still kept some pidfile CLI options code around for the purposes of backwards compatibility. Once we've updated our environment variables for Tron (which includes a PIDFILE but not LOCKFILE variable), we can fully remove the pidfile code.

debian/tron.service
```diff
@@ -5,7 +5,7 @@ After=network.target
 [Service]
 User=tron
 EnvironmentFile=/etc/default/tron
-ExecStart=/usr/bin/trond --pid-file=${PIDFILE} --working-dir=${WORKINGDIR} -H 0.0.0.0 -P 8089 -l ${LOGCONF}
+ExecStart=/usr/bin/trond --lock-file=${LOCKFILE:-$PIDFILE} --working-dir=${WORKING+DIR} -H 0.0.0.0 -P 8089 -l ${LOGCONF}
```
keymone: `WORKING+DIR` <- that seems wrong?

tron/trondaemon.py
```diff
@@ -26,65 +26,43 @@
 log = logging.getLogger(__name__)
 
 
-class PIDFile(object):
-    """Create and check for a PID file for the daemon."""
+class LockFile(object):
+    """ Create and maintain a lock for the daemon """
 
     def __init__(self, filename):
         self.lock = flockfile.FlockFile(filename)
```
keymone: i know this can seem like neverending task, but we're depending on 6 year old wrapper of a library deprecated 3 years ago.

can we ditch all of that and just use stdlib's fctnl?

docs: https://docs.python.org/3/library/fcntl.html#fcntl.flock

example: http://tilde.town/~cristo/file-locking-in-python.html

tron/trondaemon.py
```diff
@@ -160,30 +140,36 @@ def __init__(self, options):
...
             self._run_mcp()
             self._run_www_api()
             self._run_manhole()
             self._run_reactor()
 
     def _run_manhole(self):
+        # This condition is made with the assumption that no existing daemon
+        # is running. If there is one, the following code could potentially
+        # cause problems for the other daemon by removing its socket.
+        if (os.path.exists(self.manhole_sock) and
```
qui: Hm, this is confusing. What about getting rid of wantPID below, so no lock file is created? We don't really need one anymore. Then you can always delete the manhole_sock file if it exists.

debian/tron.service
kawaiwanyelp: Oops!

tron/trondaemon.py
kawaiwanyelp: Yeah, I can do that.

tron/trondaemon.py
```diff
@@ -26,65 +26,43 @@
...
-            log.warning("Lockfile was already unlocked.")
+            log.warning(f"Tron lockfile already unlocked: {self.filename}")
 
     def __exit__(self, *args):
-        self._try_unlock()
+        self.release()
         try:
-            os.unlink(self.filename)
-            log.info(f"Removed pidfile: {self.filename}")
+            os.remove(self.filename)
```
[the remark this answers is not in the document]
kawaiwanyelp: Okay.

tron/trondaemon.py
kawaiwanyelp: Ah, `wantPID=1` is what's causing the lock to be generated. I'll remove it to avoid confusion.

tests/utils/flockfile_test.py
```diff
@@ -0,0 +1,68 @@
...
+    mock_flockfile.acquire()  # already tested
+
+    mock_flockfile.release()
+
+    assert not mock_flockfile._has_lock
+
+
+@mock.patch('fcntl.flock', mock.Mock(side_effect=BlockingIOError), autospec=None)
+def test_release_other_process_locked(mock_flockfile):
+    mock_flockfile._has_lock = True  # not possible, but allows us to test below
```
qui: Same here, what about just asserting that an error is raised, if you attempt to release the lock when something else has locked it already?

PR #578, opened 2018-10-30
Trigger timeout attribute

keymone:
- add trigger_timeout attribute to action, defaulting to None
- add trigger_timeout_timestamp attribute to actionrun, initialized in action run factory from job's run_time
- transition  to failed state when trigger_timeout passes and there are still triggers pending

bonus:
- add names to special exit codes, we'll maybe use them in UIs later

### DIFF
--- a/tron/core/actionrun.py
+++ b/tron/core/actionrun.py
@@ -100,6 +100,7 @@ class ActionRunFactory(object):
             'trigger_downstreams': action.trigger_downstreams,
             'triggered_by': action.triggered_by,
             'on_upstream_rerun': action.on_upstream_rerun,
+            'trigger_timeout': action.trigger_timeout,
         }
         if action.executor == ExecutorTypes.mesos:
             return MesosActionRun(**args)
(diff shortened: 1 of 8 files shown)


tron/core/actionrun.py
```diff
@@ -572,15 +591,43 @@ def cleanup(self):
...
-            EventBus.subscribe(trigger, self.__hash__(), self.trigger_notify)
+        if self.triggered_by:
+            if self.remaining_triggers:
+                now = timeutils.current_time().timestamp()
+                delay = max(self.trigger_timeout_timestamp - now, 1)
+                self.trigger_timeout_call = reactor.callLater(
+                    delay, self.trigger_timeout_reached
+                )
+
+            for trigger_pattern in self.triggered_by:
```
qui: Is this the same as subscribing to each item in `self.remaining_triggers`?

tron/core/actionrun.py
```diff
@@ -572,15 +591,43 @@ def cleanup(self):
...
+                    )
+
+    def trigger_timeout_reached(self):
+        if self.remaining_triggers:
+            self.trigger_timeout_call = None
+            log.warning(
+                f"{self} reached timeout waiting for: {self.remaining_triggers}"
+            )
+            self.fail(self.EXIT_TRIGGER_TIMEOUT)
+        else:
```
qui: This shouldn't happen normally, right? (the timeout being called but the triggers have finished)

keymone: it can happen, just very unlikely. decided to put the code in for completeness anyway.

PR #579, opened 2018-10-30
Don't display actions/jobs in sorted order in tronview

solarkennedy:
This is the "easy" part, but I think we need to be careful what we wish for.
This also "unsorts" the big `tronview` list. I think I kinda want that sorted?

But the bigger issue is that "the order on disk" is not what users think it is. This is because tronfig re-serializes the content into yaml *again* before posting. 

This PR is only half of the puzzle, because what the user actually needs is this *plus* us not sorting the yaml when we post. To do that, we'll need a different yaml library or some workarounds:
https://stackoverflow.com/questions/16782112/can-pyyaml-dump-dict-items-in-non-alphabetical-order

I could be convinced in either direction about what we should do (including no-ship).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,8 @@ example-cluster/manhole.sock*
 example-cluster/_events/
 *.stdout
 *.stderr
+dev/manhole.sock.lock
+dev/tron.pid
 
 # Generated debian artifacts
 debian/tron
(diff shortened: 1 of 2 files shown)


qui:
What if you only overrode the rows method for DisplayActionRuns? Then tronview would still be sorted.

I think for dicts, the user shouldn't expect any sort of order. Even if we maintain the order on disk, can we ever guarantee the order that Python will iterate through them?

For lists, the re-serialization shouldn't matter.

solarkennedy:
In python 3.7 dicts *are* ordered. Maybe we can just wait for that? And if users want order, ask them to move to a dict?

PR #580, opened 2018-10-30
modify action_id during job migration

chlgit:
Testing locally at my dev box.


requirements-dev.txt
```diff
@@ -1,6 +1,6 @@
 flake8==3.6.0
 mock==2.0.0
-pre-commit==1.12.0
+pre-commit<1.12.0
```
qui: We're pinning stuff exactly now, so let's put 1.11.2.

PR #581, opened 2018-11-02
Allow TASK_LOST updates to retry automatically

qui:
Before, we had decided that we shouldn't retry if an action becomes unknown, because it means we've lost the connection and we don't know enough to retry. 

With Mesos, actions become unknown if we get a TASK_LOST update for them. In most cases, TASK_LOST means the agent running the task was terminated, and we should retry. It's possible that we get TASK_LOST for a task that is still running, in the case of a network partition, but I think if an action has automatic retries configured, they should be resilient to at-least-once execution anyway. 

This is kind of the same as losing an SSH connection, but... it didn't make sense to me to tell people that they would need to manually retry these cases, if it's going to be common with spot fleet. Thoughts?

### DIFF
--- a/tron/core/actionrun.py
+++ b/tron/core/actionrun.py
@@ -155,7 +155,7 @@ class ActionRun(Observable):
             RUNNING:
                 dict(fail_unknown=UNKNOWN, **default_transitions),
             STARTING:
-                dict(started=RUNNING, fail=FAILED),
+                dict(started=RUNNING, fail=FAILED, fail_unknown=UNKNOWN),
             UNKNOWN:
                 dict(running=RUNNING, **default_transitions),
             QUEUED:
@@ -342,7 +342,7 @@ class ActionRun(Observable):
         if run.is_running:
             run._done('fail_unknown')
         if run.is_starting:
-            run._exit_unsuccessful(None)
+            run.handle(ActionCommand.FAILSTART)
         return run
 
     def start(self):
@@ -444,7 +444,10 @@ class ActionRun(Observable):
                         len(self.exit_statuses),
                     )
                 )
-        return self.fail(exit_status)
+        if exit_status is None:
+            return self._done('fail_unknown', exit_status)
+        else:
+            return self._done('fail', exit_status)
 
     def triggers_to_emit(self) -> List[str]:
         if not self.trigger_downstreams:
@@ -720,7 +723,7 @@ class SSHActionRun(ActionRun, Observer):
             return self.transition_and_notify('started')
 
         if event == ActionCommand.FAILSTART:
-            return self._exit_unsuccessful(None)
+            return self._exit_unsuccessful(-2)
 
         if event == ActionCommand.EXITING:
             if action_command.exit_status is None:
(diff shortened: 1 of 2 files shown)


solarkennedy:
I think this is the same kind of behavior that surprised users when we migrated them to chronos. A non-idempotent job can start, get lost halfway through, then we retry automatically and then surprise, we bill people twice, etc.

Maybe we are past this now. Maybe it could be a toggle that is defaulted to ON, but when we migrate users we comment it out?

```
myaction:
  command: foo
  # Remove this line if it is safe to automatically retry this action in the case of an infrastructure-related issue:
  retry_when_lost: false
  # Uncomment this line if it is safe retry this action if it fails
  # retries: 3
```

I love systems that do their best to keep on going in the fact of failure, but I don't want to surprise our users with new behavior.

Rob-Johnson:
is there anything more we can do to try and assert the state of the mesos agent? do we get any notification from mesos that the agent is disappearing? it'd be nice to confirm.

I think I agree with kyle - maybe we should pass the decision back to the user if we can't be sure that this won't be surprising for people

qui:
@solarkennedy It will only retry if the action has retries configured already. This change makes it so that it will retry the "no exit code" case for Mesos, in addition to existing retries for non-zero exit codes.

@Rob-Johnson Mesos will say that an agent is "removed" if it fails its healthchecks, but there's no way to know if it failed its healthchecks because it was terminated vs because it lost network connectivity to the master.

Adding a couple more changes here. I realized that if there are no retries, my first commit would put the action into the "failed" state, which would be misleading to users. It should still be unknown.

Then I also needed to change the way FAILSTART is handled in SSH vs Mesos. In SSH, we want it to be 'failed' (the SSH connection had an error). In Mesos, the task may have been staging, so it should be unknown.

chlgit:
I like the idea that Tron should try its best to finish the job if it could. The jobs with retry configuration should be safe enough to retry. The jobs which can't survive from losing hosts could run on the stable pool instead.

tron/core/actionrun.py
```diff
@@ -672,7 +675,7 @@ def handle_action_command_state_change(self, action_command, event):
             return self.transition_and_notify('started')
 
         if event == ActionCommand.FAILSTART:
-            return self._exit_unsuccessful(None)
+            return self._exit_unsuccessful(-2)
```
chlgit: Out of curiosity, why -2? Any non-zero integer works, right?

suntzu86:
@qui I'm think expecting jobs with retry to be 'at least once'-safe is reasonable.

But I think that is different than a job that can have multiple instances running simultaneously.

To that end, I'm a fan of @solarkennedy 's suggestion. Unless I'm totally misunderstanding :/

tron/core/actionrun.py
suntzu86: @qui nit: could we define a named constant then?  -2 seems a bit arcane

qui: Yeah, that was done here: https://github.com/Yelp/Tron/pull/578
I think that was after I started this PR, I'll update it in a follow-up - there's a bug elsewhere I need to fix anyway.

PR #583, opened 2018-11-02
Add cluster option to metrics script

kawaiwanyelp:
### Description
- Added a `cluster` option to the `get_tron_metrics` script
- In the script, we forward the `cluster` value as a dimension in all Tron metrics we send to meteorite

### Test
- dry-run for metrics script
- make test


qui:
Can you add dev/tron.lock to gitignore?

PR #589, opened 2018-11-16
[wip] SIGTERM not being handled... sometimes

kawaiwanyelp:
### Description
- Previously, when Tron received a SIGTERM, it would bypass our graceful shutdown procedure and immediately terminate. The problem was that the default handler for SIGTERM (`signal.SIG_DFL`) immediately terminated, not giving a chance for `signal.sigtimedwait` to receive the signal and handle it synchronously.
- I've changed all but SIGINT's handler to be `signal.SIG_IGN`. This 'handler' ignores the signal, which allows `signal.sigtimedwait` to handle it.

### Testing
- make dev, send SIGTERM using `kill`
- mesosstage (**sometimes successful, described below**)

### Notes to reviewers
- The old adage after all is that "Multithreaded programs and signals do not play well each other".
- On my devbox via `make dev`, I've been able to terminate `trond` gracefully; i.e. we sychronously receive the signal and call `_handle_shutdown`.
- However, on mesosstage, it only happens sometimes.
- @solarkennedy's theory:
    - @solarkennedy seems to think that if the signal is received when the main thread is running (i.e. it is currently the thread that acquired the GIL), Tron will properly terminate. However, if some other thread, say our pymesos thread, is the one running instead, even though the Tron process has received the signal, Python will throw it into a blackhole and we will never shutdown.
    - We tested this hypothesis but setting the Tron service shutdown timer to infinity, setting some debug statements that would log when we were on the main thread, attempting to stop the Tron service. We saw that when the systemd "Stopping trond" message was logged amidst some non-main thread logs, Tron never seemed to receive the signal ever. It didn't even register the signal after it switched back to the main thread (indicated by main thread logs time stamped after we saw systemd's "Stopping trond" message).
    - So why does it work on `make dev` (on my devbox) so regularly? I think it's because I only have 1 test job running on Mesos, whereas mesosstage has many more. In other words, mesosstage's Tron spends much more time running the pymesos thread instead of the main thread.
    - If this theory is right, I am not sure how to program around the GIL, nor am I sure if pymesos is to blame. Perhaps this problem exists no matter what Python-Mesos interface we use.
- I don't think @solarkennedy's theory is right because when I set the SIGTERM signal handler to something else (say, something that logs or terminates immediately), whether or not we are in the midst of main thread or non-main thread log statements, the 'custom' signal handler is called without fail.
    - However, @keymone argues that handling signals synchronously like this is really the only correct way to handle signals, and I am inclined to agree.
    - We might need to take a closer look at the interaction between signal handlers and `sigwait`s (including `sigtimedwait`). As discussed in #560, `sigwait`s do not seem to unblock unless signal handlers ignore signals, and there are always signal handlers set (by Python or otherwise).

### DIFF
--- a/tron/trondaemon.py
+++ b/tron/trondaemon.py
@@ -98,6 +98,7 @@ class TronDaemon(object):
         setup_logging(self.options)
 
         self.mcp = None
+        # This must happen before the context is built
         self._sigint_handler = self._make_sigint_handler(
             signal.getsignal(signal.SIGINT)
         )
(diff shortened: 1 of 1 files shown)


keymone:
what happens if we set signal handler in separate thread while `sigtimedwait` is running? my gut feeling is that the issue is interplay between waiting and setting signals, should we try removing signal handlers from nodaemoncontext completely and switching to a fork of pymesos that doesn't touch signals: https://github.com/keymone/pymesos/commit/a011662f1b038c55f7ab84363a03b43ae7b286f2 ?

kawaiwanyelp:
@keymone I'm now in favor of forking.

As for waiting vs setting signals, firstly, there is no way to completely remove signal handlers - there are defaults. Additionally, unless we can figure out the interplay you mentioned, we have to the signal handlers to `SIG_IGN` so that they don't terminate Tron immediately.

kawaiwanyelp:
Closed in favor of #592.

PR #591, opened 2018-11-18
Add signal queue with tests

kawaiwanyelp:
### Description
- Add the SignalQueue class, which makes it so that signal handlers enqueue signals, so that we can dequeue them elsewhere to handle them serially.
- Modify the Tron daemon to use this class to handle signals

### Tests done
- make test
- pymesos interrupt tests (via docker): This change still allows us to detect when interrupts come from pymesos and continue on regardless.
- mesosstage: Manually started and stopped the tron service to ensure SIGTERM's are being properly caught and handled via the new SignalQueue.

### Notes to reviewers
- Solves the issue highlighted in PR #589
- What do I use a non-thread-safe queue?
    - tl;dr: Deadlock caused by enqueuing pending signals while waiting for them
    - I approached this problem by simulating `sigwait`'s behavior; that is, when a signal is received, assuming the signal handler doesn't terminate the program, the signal should be set to pending so that `sigwait` can return them.
    - I initially used Python's `queue.Queue` as my queue implementation. These queues are designed to be used in multithreaded applications and are therefore made thread-safe by way of locks.
    - What I found was that the act of calling a signal handler while the main thread is waiting for a signal can result in deadlock. Observe the following traceback:
        - --- Normal Main Thread Operation ---
        - q.wait()
        - q.get_nowait()
        - q._lock.acquire() (internal)
        - --- Signal Handler Interrupts (always on main thread)---
        - q.handler(signal, frame)
        - q.put_nowait(signal)
        - q._lock.acquire() (internal) <------- DEADLOCK (lock hasn't - been released)
    - As a result, a non-thread-safe queue implementation had to be used. However, given the simplicity of how we used this queue, locks are not needed.

### DIFF
--- a/tests/utils/signalqueue_test.py
+++ b/tests/utils/signalqueue_test.py
@@ -1,0 +1,42 @@
+import signal
+
+import mock
+import pytest
+
+from tron.utils.signalqueue import SignalQueue
+
+
+@pytest.fixture
+def mock_lock():
+    with mock.patch('threading.Lock', autospec=True) as mock_Lock:
+        yield mock_Lock
+
+
+@pytest.fixture
+def mock_sigqueue(mock_lock):
+    return SignalQueue()
+
+
+def test_init(mock_sigqueue, mock_lock):
+    # we don't want any locking in this module, since it'll cause deadlock
+    assert mock_lock.call_count == 0
+
+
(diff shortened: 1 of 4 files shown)


kawaiwanyelp:
Closed in favor of #592.

PR #593, opened 2018-11-20
Add consistency check on Tron startup

kawaiwanyelp:
### Description
- Add a consistency check after loading state during startup to save it immediately after
- Failing to save would indicate corruption

### Testing done
- manual testing
- make test

### Notes to reviewers
- I'm not sure how much this protects us, since I haven't been able to generate a Tron bdb state file that can load but not save.
- Additionally, this check doesn't really protect us while Tron is running, i.e. Tron's running state gets corrupted and doesn't save.

### DIFF
--- a/tron/serialize/runstate/statemanager.py
+++ b/tron/serialize/runstate/statemanager.py
@@ -152,10 +152,13 @@ class PersistentStateManager(object):
 
         jobs = self._restore_dicts(runstate.JOB_STATE, job_names)
         frameworks = self._restore_dicts(runstate.MESOS_STATE, ['frameworks'])
-        return {
+
+        state = {
             runstate.JOB_STATE: jobs,
             runstate.MESOS_STATE: frameworks,
         }
+        self._check_consistency(state)
+        return state
 
     def _restore_metadata(self):
         metadata = self._impl.restore([self.metadata_key])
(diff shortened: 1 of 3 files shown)


solarkennedy:
I think crashing faster is good, also with an "obvious" error.
Any idea how long this function takes on our prod state?

tron/serialize/runstate/statemanager.py
```diff
@@ -201,6 +204,22 @@ def _save_from_buffer(self):
...
+        :param state: A dict of serialized state, like that returned by
+            self.restore
+        """
+        log.info("Running consistency checks")
+        try:
+            # We attempt to save all the state. A failure indicates corruption
+            for type_enum, items in state.items():
+                for name, state_data in items.items():
+                    self.save(type_enum, name, state_data)
+        except PersistenceStoreError as e:
```
solarkennedy: Does this say what name/type/item/etc it couldn't save?

tron/serialize/runstate/statemanager.py
```diff
@@ -201,6 +204,22 @@ def _save_from_buffer(self):
...
+            log.error("Failed consistency check: Could not save serialized state")
+            raise e
```
chlgit: So, Tron would crash if the error is raised here, right?

tron/serialize/runstate/statemanager.py
kawaiwanyelp: I'm not sure. Like I said, I haven't been able to simulate the error. I only know that this exception is thrown based on that we try to handle it elsewhere when saves fail. @keymone do you know?

tron/serialize/runstate/statemanager.py
kawaiwanyelp: Yes, which is exactly what we want. If it fails the consistency check on start up, we should fail immediately.

tron/serialize/runstate/statemanager.py
keymone: the error happens when bdb file on disk is corrupt, you can't easily reproduce that in python. PersistenceStoreError won't have information about which key failed in this case - definitely worth try/except'ing inside the loop and re-raising with key name in the exception.

kawaiwanyelp: We don't actually need to add the key name to the exception as it is already added (and logged) where we throw it (`_save_from_buffer`).

PR #595, opened 2018-11-27
Mesos action run should fail if it gets an offer timeout event

qui:
There was a bug in logging terminal events that don't come from Mesos, so the task didn't exit. (Introduced in https://github.com/Yelp/Tron/pull/590.) I moved the logging code to `log_event_info` (which we wrap with a try-except to prevent logging from messing up state transitions) and check that the event is from Mesos before assuming event.raw will be a dict.

After the task exits with FAILSTART and a non-zero exit code, the action run handler should also fail. I changed exit_unsuccessful to transition to unknown if the exit status is None, instead of always failing, in https://github.com/Yelp/Tron/pull/581.

Manually verified this works for offer timeouts by running an action with an invalid pool, which will always lead to an offer timeout. The action fails.

### DIFF
--- a/tests/mesos_test.py
+++ b/tests/mesos_test.py
@@ -205,13 +205,15 @@ class TestMesosTask(TestCase):
         assert self.task.is_failed
         assert self.task.is_done
 
-    def test_handle_unknown_terminal_event(self):
+    def test_handle_terminal_event_offer_timeout(self):
         self.task.started()
         event = mock_task_event(
             task_id=self.task_id,
             platform_type=None,
             terminal=True,
             success=False,
+            raw='failed due to offer timeout',
+            message='stop',
         )
         self.task.handle_event(event)
         assert self.task.is_failed
--- a/tests/mesos_test.py
+++ b/tests/mesos_test.py
@@ -243,7 +243,13 @@ class TestMesosTask(TestCase):
         assert self.task.is_complete
 
     def test_log_event_error(self):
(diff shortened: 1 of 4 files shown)


tests/mesos_test.py
```diff
@@ -205,13 +205,15 @@ def test_handle_error(self):
...
-    def test_handle_unknown_terminal_event(self):
+    def test_handle_terminal_event_offer_timeout(self):
         self.task.started()
         event = mock_task_event(
             task_id=self.task_id,
             platform_type=None,
             terminal=True,
             success=False,
+            raw='failed due to offer timeout',
+            message='stop',
```
kawaiwanyelp: These two lines will cause the same "AttributeError: 'str' object has no attribute 'get' " error, because we call `get` on `raw` and `message` [here](https://github.com/Yelp/Tron/blob/offer-timeout-bug/tron/mesos.py#L210). However, the tests will still pass because we catch those exceptions [here](https://github.com/Yelp/Tron/blob/offer-timeout-bug/tron/mesos.py#L230). Maybe write a separate test for `log_event_info` so that check that the logs we want are indeed being logged?

qui: It doesn't cause the exception, because now we only call get when platform_type isn't None.

There's a test for logging exceptions [here](https://github.com/Yelp/Tron/blob/offer-timeout-bug/tests/mesos_test.py#L245), I'll update it to include log.warning.

PR #602, opened 2018-12-18
convert volumes to dict when recovering

Rob-Johnson:
this needs to be an ordereddict when it is given to mesos so it can be
merged with default volumes


tron/core/actionrun.py
```diff
@@ -792,7 +792,7 @@ def recover(self):
             docker_image=self.docker_image,
             docker_parameters=self.docker_parameters,
```
qui: Looks like extra_volumes is the only one that really matters, but docker_parameters and constraints are also the incorrect type here (see submit command above). Maybe we can combine this logic a bit into a helper function in MesosActionRun?

PR #606, opened 2019-01-05
Added description to check_tron_jobs for better email titles

solarkennedy:
`description` in theory ends up in the title of these emails.

I haven't actually tested this yet.


qui:
I've suggested that people use "description" before, to add more context to their emails. It doesn't seem to turn up in the emails, but you can see it in Uchiwa: https://uchiwa.yelpcorp.com/#/client/uswest1-prod/tron?check=check_tron_job.admetrics.autobidding_metrics_daily

paasta-contract-monitor also adds a description.

qui:
Sure, part of my point is that it doesn't actually show up in the email subject?

solarkennedy:
I'm going to try to solve this at the sensu_handlers layer instead for all checks.
Also I was wrong about `description` anyway.

PR #607, opened 2019-01-15
Tron 231

EmanekaT:
https://jira.yelpcorp.com/browse/TRON-231
The issue is in function getDuration, the input could be '1 day, 12:02:1.12'  but the getDuration function only handles inputs like '12:02:1.12'.
I tested by setting duration to 1 day, 12:02:1.12, and run make dev, and checked the duration is correct.

### DIFF
--- a/tronweb/coffee/views.coffee
+++ b/tronweb/coffee/views.coffee
@@ -35,8 +35,8 @@ window.dateFromNow = (string, defaultString='never') ->
 
 window.getDuration = (time) ->
     if time.indexOf("day") != -1
-        [day, rest] = time.split(',')
-        [days, b]= day.split(' ')
+        [dayStr, rest] = time.split(',')
+        [days, day]= dayStr.split(' ')
         [time, ms] = rest.split('.')
         [hours, minutes, seconds] = time.split(':')
         moment.duration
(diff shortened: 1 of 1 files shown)


tronweb/coffee/views.coffee
```diff
@@ -34,12 +34,23 @@ window.dateFromNow = (string, defaultString='never') ->
 
 
 window.getDuration = (time) ->
-    [time, ms] = time.split('.')
-    [hours, minutes, seconds] = time.split(':')
-    moment.duration
-        hours: parseInt(hours)
-        minutes: parseInt(minutes)
-        seconds: parseInt(seconds)
+    if time.indexOf("day") != -1
```
chlgit: https://en.wikipedia.org/wiki/Don%27t_repeat_yourself  
There are some duplicated codes here. Could you fix that?

EmanekaT:
> If the duration is 10 hours, would it show "0 day 10 hours" or "10 hours"?

Just 10 hours.

tronweb/coffee/views.coffee
```diff
@@ -34,9 +34,14 @@ window.dateFromNow = (string, defaultString='never') ->
 
 
 window.getDuration = (time) ->
+    days = '0'
+    if time.indexOf("day") != -1 or time.indexOf("days") != -1
```
chlgit: Should `if time.indexOf("day") != -1` be enough here?

EmanekaT: Oh... yes.

PR #610, opened 2019-01-18
Added disk support to tron on mesos

solarkennedy:
This is low priority, but a nice to have.

I didn't make it a required option to make it easier for me to roll this out.


solarkennedy:
Er, I need to get my types right.

kawaiwanyelp:
I added a default of 100mb for an action's disk. That also fixes @solarkennedy type error.

kawaiwanyelp:
I changed the disk space default to 1024mb to match https://github.com/Yelp/paasta/blob/65517582b5282f267a62f5ed402d35ce1e185706/paasta_tools/utils.py#L457.

tron/core/action.py
```diff
@@ -53,6 +54,7 @@ def from_config(cls, config):
             executor=config.executor,
             cpus=config.cpus,
             mem=config.mem,
+            disk=config.disk or 1024.,
```
keymone: this is probably fine, but in python `0.0` is `false`, so user specifying `disk=0` will erroneously get `disk=1024`

PR #611, opened 2019-01-30
Fix job appearing pending when waiting on trigger requirement

keymone:
Jobrun object will be in state running if it has no running Actionruns but some of them are waiting for trigger requirement.

### DIFF
--- a/tron/core/jobrun.py
+++ b/tron/core/jobrun.py
@@ -214,13 +214,8 @@ class JobRun(Observable, Observer):
         ]
 
         if not started_runs:
-            blocked_on_triggers = [
-                r
-                for r in self.action_runs
-                if r.is_blocked_on_trigger
-            ]
-            if blocked_on_triggers:
-                for r in blocked_on_triggers:
+            for r in self.action_runs:
+                if r.is_blocked_on_trigger:
                     log.debug(f"{r} is blocked on triggers: {r.remaining_triggers}")
 
         return started_runs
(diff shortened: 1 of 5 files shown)


tron/core/jobrun.py
```diff
@@ -207,12 +207,24 @@ def _start_action_runs(self):
...
             action_run
             for action_run in self.action_runs.get_startable_action_runs()
             if action_run.start()
         ]
 
+        if not started_runs:
+            blocked_on_triggers = [
+                r
+                for r in self.action_runs
+                if r.is_blocked_on_trigger
```
Rob-Johnson: there is an inconsistency between this and line 223 - is this supposed to be triggers or trigger?

tron/core/jobrun.py
```diff
@@ -207,12 +207,24 @@ def _start_action_runs(self):
...
+            ]
+            if blocked_on_triggers:
```
Rob-Johnson: this is minor, but couldn't you just traverse the list once?

```
for r in self.action_runs:
     if r.is_blocked_on_triggers:
         log.debug(f"{r} is blocked on triggers: {r.remaining_triggers}")
```

tron/core/jobrun.py
```diff
@@ -308,6 +320,8 @@ def state(self):
             return ActionRun.STARTING
         if self.action_runs.is_failed:
             return ActionRun.FAILED
+        if self.seconds_until_run_time() == 0 and self.action_runs.is_blocked_on_trigger:
+            return ActionRun.RUNNING
```
Rob-Johnson: this seems kind of confusing - we want it to be marked running when it isn't?

keymone: a lot of tron machinery depends on having correct states at correct times, i found this to be least confusing and smallest change that makes things work. we don't have "waiting on requirements" state because root actions couldn't have requirements in tron before - they would never start, now this assumption is broken.

PR #613, opened 2019-02-07
Fix inconsistency between JobRun state attribute and is_<state> checks

keymone:
`is_<state>` checks on jobruns are proxied directly  to actionrun collection and are not always consistent with value of jobrun.state. This PR replaces proxying of `is_<state>` checks with a more direct  comparison to `state` property value.

### DIFF
--- a/tron/core/jobrun.py
+++ b/tron/core/jobrun.py
@@ -148,7 +148,7 @@ class JobRun(Observable, Observer):
             action_run.setup_subscriptions()
 
         self.action_runs_proxy = proxy.AttributeProxy(
-            self.action_runs,
+            run_collection,
             [
                 'queue',
                 'cancel',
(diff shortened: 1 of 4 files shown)


tron/core/jobrun.py
```diff
@@ -325,9 +315,14 @@ def state(self):
         return ActionRun.UNKNOWN
 
     def __getattr__(self, name):
-        if self.action_runs_proxy:
+        if name[:3] == 'is_':
```
Rob-Johnson: `name.startswith('is_')`?

tron/core/jobrun.py
```diff
@@ -325,9 +315,14 @@ def state(self):
...
+            if name[3:] not in ActionRun.STATE_MACHINE.states:
```
Rob-Johnson: maybe pull `name[3:]` into a variable to reduce the duplication?

PR #614, opened 2019-02-07
Add cross-job dependency visualization to Tronweb

drmorr0:
### Description of change

1) When Tron initializes, we compute the _entire_ DAG so that we can quickly reference upstream and downstream triggers.  This is stored in the JobGraph class.
2) The JobGraph provides a way to construct an ActionGraph for a particular job: instead of reading the ActionGraph from the config, we look at the jobgraph and get all the actions for that job, along with all of the upstream and downstream triggers.
    - The ActionGraph class has largely been rewritten; there was a bunch of code in there that wasn't being used anywhere, which I stripped out
    - The ActionGraph interface has also changed slightly
3) The ActionGraphAdapter and the ActionRunGraphAdapter are responsible for getting all of the upstream dependencies for each action in the job into the right format.

### Testing done

* So far all the testing I've done has just been on my local devbox.  You can see what tronweb looks like [here](http://dev54-uswest1adevc:8089/web/#home) (requires VPN).

### DIFF
--- a/dev/config/MASTER.yaml
+++ b/dev/config/MASTER.yaml
@@ -1,5 +1,6 @@
+eventbus_enabled: True
 ssh_options:
-   agent: False
+   agent: True
 
 nodes:
   - hostname: localhost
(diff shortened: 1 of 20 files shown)


qui:
Thinking again about what users would need, I think the killer feature is answering "Why did my trigger time out?" For that, it would be great if from an action run, you had links to the immediately upstream action runs that produce the triggers you depend on. 

That being said, it think it's also useful to be aware of all the jobs upstream and downstream, which this does. I'd be okay with merging this and following up with the links, just worth thinking about the links in case getting the exact upstream runs would change the design in a significant way. 

If you decide to go forward with this, let me know - I haven't done a detailed code review just in case it changes.

qui:
Another thing - the tronweb example seems to show some bugs in cross-job dependencies, which may or may not have existed before.

The job run [MASTER.testjob1.7](http://dev54-uswest1adevc:8089/web/#job/MASTER.testjob1/7) is in a bad state - the second action seems to have failed based on trigger timeout immediately, and the first action remains scheduled. I would expect 1) the trigger timeout to take some time and 2) the whole job run to fail if that happens.

The job [MASTER.testjob2](http://dev54-uswest1adevc:8089/web/#job/MASTER.testjob2) depends on testjob1. All of the previous runs are "running" (waiting for a trigger). allow_overlap is false by default, so only one should be allowed to run at a time.

qui:
both bugs mentioned above fixed in #622

PR #617, opened 2019-02-26
Migrating from Berkley DB to DynamoDB (TRON-638)

EmanekaT:
Tests are added and the statestore implementation is basically finshed.
I also added requirements-tools. I limited the version of moto and boto3 version because the latest moto release does not work well with the latest boto3. 
I would test it on mesosstage, and then deploy it after chl is back.


tron/serialize/runstate/DynamoDBStateStore.py
```diff
@@ -0,0 +1,132 @@
...
+        It builds a unique partition key. The key could be objects with __str__ method.
+        """
+        return ShelveKey(type, iden)
+
+    def restore(self, keys: [ShelveKey]) -> dict:
+        """
+        Fetch all under the same parition key(keys).
+        ret: <dict of key to states>
+        """
+        def join(d1, d2):
```
keymone: i think this is the same as `some_dict.update(other_dict)`

tron/serialize/runstate/DynamoDBStateStore.py
```diff
@@ -0,0 +1,132 @@
...
+            if i not in res2.keys():
+                res2[i] = j
+        return res2
+
+    def __getitem__(self, key: ShelveKey) -> object:
+        """
+        It returns an object which is deserialized from binary
+        """
+        val = bytearray()
+        for index in range(self._get_num_of_partitions(key)):
```
keymone: we probably need a single query with consistent read here instead of bunch of get_item calls in a loop

tron/serialize/runstate/DynamoDBStateStore.py
```diff
@@ -0,0 +1,132 @@
...
+        thread2 = Thread(target = self.shelve.save, args=(key_value_pairs,))
+        thread1.start()
+        thread2.start()
+        thread1.join()
+        thread2.join()
+
+    def __setitem__(self, key: ShelveKey, val: bytes) -> None:
+        size = math.ceil(len(val) / OBJECT_SIZE)
+        for index in range(size):
+            self.table.put_item(
```
keymone: use batch writer here: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/dynamodb.html#DynamoDB.Table.batch_writer

tron/serialize/runstate/DynamoDBStateStore.py
```diff
@@ -0,0 +1,132 @@
...
+                    'key': str(key),
+                    'index': index,
+                    'val': val[index * OBJECT_SIZE:min(index * OBJECT_SIZE + OBJECT_SIZE, len(val))],
+                    'size': size,
+                }
+            )
+
+    def _delete_item(self, key: ShelveKey) -> None:
+        for index in range(self._get_num_of_partitions(key)):
+            self.table.delete_item(
```
keymone: again, batch writer

tron/serialize/runstate/DynamoDBStateStore.py
```diff
@@ -0,0 +1,132 @@
...
+
+    def __setitem__(self, key: ShelveKey, val: bytes) -> None:
+        size = math.ceil(len(val) / OBJECT_SIZE)
+        for index in range(size):
+            self.table.put_item(
+                Item={
+                    'key': str(key),
+                    'index': index,
+                    'val': val[index * OBJECT_SIZE:min(index * OBJECT_SIZE + OBJECT_SIZE, len(val))],
+                    'size': size,
```
keymone: rename to num_partitions?

tron/serialize/runstate/DynamoDBStateStore.py
```diff
@@ -0,0 +1,132 @@
...
+
+OBJECT_SIZE = 400000
+
+
+class DynamoDBStateStore(object):
+    def __init__(self, name, dynamodb_region) -> None:
+        self.dynamodb = boto3.resource('dynamodb', region_name=dynamodb_region)
+        self.client = boto3.client('dynamodb', region_name=dynamodb_region)
+        self.name = name
+        self.shelve = ShelveStateStore(name)
```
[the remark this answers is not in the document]
EmanekaT: For migration, we are writing to and reading from both DyanmoDB and BerkleyDB so that in case anything breaks, we could roll back without losing any data.

Rob-Johnson:
looks on the right path - are you planning on following up with schema changes to get us under the size limits?

tron/serialize/runstate/DynamoDBStateStore.py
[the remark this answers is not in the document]
keymone: don't think we should bother. this class is not that complex (especially if you remove all shelve-related stuff), we can test on stages and do a big-bang migration later.

EmanekaT:
@Rob-Johnson 
No. I decide just chop the binary into different partitions. I don't think it's necessary to modify the schema. It takes 20s to read 88M data from berkleyDB and write to dynamoDB. And 88M is all we have in prod right now.

EmanekaT:
> looks on the right path - are you planning on following up with schema changes to get us under the size limits?

tron/serialize/runstate/DynamoDBStateStore.py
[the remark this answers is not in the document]
EmanekaT: Since it's already implemented, I guess I can just use the current code and remove the shelve part once dynamodb works.

tron/serialize/runstate/dynamodb_state_store.py
```diff
@@ -0,0 +1,176 @@
...
+        self.shelve = ShelveStateStore(name)
+        self.table = self.dynamodb.Table(name.replace('/', '-'))
+
+    def build_key(self, type, iden) -> ShelveKey:
+        """
+        It builds a unique partition key. The key could be objects with __str__ method.
+        """
+        return ShelveKey(type, iden)
+
+    def isValid(self, shelve_kv_pairs: dict, dynamo_kv_pairs: dict) -> bool:
```
keymone: rename to `is_valid`

tron/serialize/runstate/dynamodb_state_store.py
```diff
@@ -0,0 +1,176 @@
+import copy
+import math
+import pickle
+from threading import Thread
+
+import boto3
+
+from tron.serialize.runstate.shelvestore import ShelveKey
```
keymone: do we really need to involve shelve keys here? can we just make this an implementation detail of dynamodb store itself? like have `build_key` simply  return `f"{type}:{iden}"`? i'd like this class to not depend on anything but boto.

EmanekaT: It's easier to implement in this way. DynamoDB and ShelveStateStore could thus share the same key. Once thing I can do is to convert ShelveKey to f"{type}:{iden}" before storing to DynamoDB and reading. Then, after migration, I can just use  f"{type}:{iden}".

tron/serialize/runstate/dynamodb_state_store.py
```diff
@@ -0,0 +1,131 @@
...
+        ret: <dict of key to states>
+        """
+        items = zip(
+            keys,
+            (self[key] for key in keys),
+        )
+
+        return {k: v for k, v in items if v}
+
+    def alert(self):
```
keymone: move to mirror store

EmanekaT: I'm thinking about keeping alert here even after mirrorstore is removed.

keymone: we're only alerting on inconsistency between shelve and dynamodb so it only makes sense to have this function in mirror store

EmanekaT: In this case, yes. I mean I'm thinking about adding checks for dynamodb. So I keep it here for the future.

tron/serialize/runstate/mirror_state_store.py
```diff
@@ -0,0 +1,73 @@
...
+
+class MirrorStateStore():
+    def __init__(self, shelve_store, dynamodb_store) -> None:
+        self.shelve_store = shelve_store
+        self.dynamodb_store = dynamodb_store
+
+    def is_valid(self, shelve_kv_pairs: dict, dynamo_kv_pairs: dict) -> bool:
+        """
+        It checks if all keys in DynamoDB are in shelveStateStore and have the same values.
+        It is possible that some keys are in shelveStateStore but is not in DynamoDB when the
```
qui: How will you know when you're ready to stop using shelve, then? (when all keys in shelve are in dynamo)

tron/serialize/runstate/dynamodb_state_store.py
```diff
@@ -0,0 +1,223 @@
...
+        """
+        Remove all previous data with the same partition key, examine the size of state_data,
+        and splice it into different parts under 400KB with different sort keys,
+        and save them under the same partition key built.
+        """
+        # try:
+        for key, val in key_value_pairs:
+            self._delete_item(key)
+            self[key] = pickle.dumps(val)
+        # except Exception as e:
```
qui: Maybe you should log the key values and exception, at least.

tron/serialize/runstate/dynamodb_state_store.py
```diff
@@ -0,0 +1,223 @@
...
+                        'index': index,
+                    }
+                )
+
+    def _get_num_of_partitions(self, key: ShelveKey) -> int:
+        """
+        Return how many parts is the item partitioned into
+        """
+        try:
+            partition = self.table.get_item(
```
qui: Fetching the whole key seems unnecessary - how about ProjectionExpression?

.travis.yml
```diff
@@ -9,6 +9,9 @@ python:
 
 before_install:
   - sudo apt-get install -y libdb5.3-dev
+  - sudo rm -f /etc/boto.cfg
+  - export AWS_SECRET_ACCESS_KEY=foobar_secret
```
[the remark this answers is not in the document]
EmanekaT: This is irrelevant to how tron in real situation. It's just a bug in moto. I need these two lines only in travis environment.

tron/serialize/runstate/mirror_state_store.py
EmanekaT: I assume unless dynamodb read/write fails, where an exception is thrown and I'll be alerted, all keys in shelve are in dynamodb if they are ever written.

tron/serialize/runstate/dynamodb_state_store.py
EmanekaT: Oh right. I forgot that.

tron/config/schema.py
```diff
@@ -95,6 +95,7 @@ def config_object_factory(name, required=None, optional=None):
     optional=[
         'connection_details',
         'buffer_size',
+        'dynamodb_region',
```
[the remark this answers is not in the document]
EmanekaT: But it's used for sql right?

tron/serialize/runstate/dynamodb_state_store.py
qui: Okay, works for now.

PR #619, opened 2019-02-26
Don't upgrade all the things when building a package

solarkennedy:
My mom always told me to never use the latest version of everything.

### DIFF
--- a/itest.sh
+++ b/itest.sh
@@ -52,6 +52,12 @@ tronfig -n MASTER /work/example-cluster/tronfig/MASTER.yaml
 tronfig /work/example-cluster/tronfig/MASTER.yaml
 cat /work/example-cluster/tronfig/MASTER.yaml | tronfig -n MASTER -
 
+if test -L /opt/venvs/tron/lib/python3.6/encodings/punycode.py; then
+    echo "Whoa, the tron package shouldn't have an encoding symlink!"
+    echo "Check out https://github.com/spotify/dh-virtualenv/issues/272"
+    exit 1
+fi
+
 kill -SIGTERM $TRON_PID
 wait $TRON_PID || true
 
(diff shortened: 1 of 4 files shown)


itest.sh
```diff
@@ -52,6 +52,11 @@ tronfig -n MASTER /work/example-cluster/tronfig/MASTER.yaml
 tronfig /work/example-cluster/tronfig/MASTER.yaml
 cat /work/example-cluster/tronfig/MASTER.yaml | tronfig -n MASTER -
 
+if test -L /opt/venvs/tron/lib/python3.6/encodings/punycode.py; then
+    echo "Whoa, the tron package shouldn't have an encoding symlink!"
+    exit 1
```
chriskuehl: Can you link to the relevant ticket? I have a feeling the reasoning behind this might get lost otherwise.

yelp_package/trusty/Dockerfile
```diff
@@ -24,20 +24,14 @@ RUN apt-get -q update && \
...
     && apt-get -q clean
 
-RUN wget https://bootstrap.pypa.io/get-pip.py -O /tmp/get-pip.py
-RUN python3.6 /tmp/get-pip.py
-RUN pip3.6 install -U tox wheel setuptools PyYAML
-
 RUN cd /tmp && \
     wget http://mirrors.kernel.org/ubuntu/pool/universe/d/dh-virtualenv/dh-virtualenv_1.0-1_all.deb && \
     gdebi -n dh-virtualenv*.deb && \
     rm dh-virtualenv_*.deb
```
chriskuehl: Probably could install dh-virtualenv via apt too?

solarkennedy:
This protects us against https://github.com/spotify/dh-virtualenv/issues/272

yelp_package/trusty/Dockerfile
solarkennedy: I could, but it is a super old version

itest.sh
solarkennedy: fixed

PR #621, opened 2019-02-27
Add docs for monitoring defaults

kawaiwanyelp:
### Description
- A lot of devs still expect the old behavior, so I'm updating the docs to say that any fields that aren't included in their job monitoring will default to those in `monitoring.yaml`.

### DIFF
--- a/docs/jobs.rst
+++ b/docs/jobs.rst
@@ -33,6 +33,13 @@ Optional Fields
     Dictionary of key: value pairs to inform the monitoring framework on how to
     alert teams for job failures.
 
+    If you're using PaaSTA, for any monitoring fields not specified for a job,
+    Tron will default to those set in `monitoring.yaml` in your soa-configs.
+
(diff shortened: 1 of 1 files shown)


docs/jobs.rst
```diff
@@ -33,6 +33,13 @@ Optional Fields
     Dictionary of key: value pairs to inform the monitoring framework on how to
     alert teams for job failures.
 
+    If you're using PaaSTA, for any monitoring fields not specified for a job,
+    Tron will default to those set in `monitoring.yaml` in your soa-configs.
+
+    You can see more about this behavior in the `PaaSTA docs`_.
+
+    .. _PaaSTA docs: https://paasta.readthedocs.io/en/latest/yelpsoa_configs.html#tron-tron-clustername-yaml
```
solarkennedy: I think 
https://paasta.readthedocs.io/en/latest/yelpsoa_configs.html#monitoring-yaml

is a better link

PR #622, opened 2019-02-27
Waiting state if an action is waiting for normal or cross-job dependencies

qui:
We need to differentiate between "waiting for dependencies" and "waiting for the previous job run to finish". QUEUED was being used for both, but now I'm proposing WAITING for the first state and QUEUED for the second. The key change is in L177 of tron/core/actionrun.py.

https://github.com/Yelp/Tron/pull/611 attempted to make "waiting for dependencies" into the RUNNING state, but that makes it seem like it's running as soon as its scheduled time has passed. That causes these issues:
* In the run_job callback that gets called when the scheduled time comes, it won't start the run if it's not "SCHEDULED". With #611, it already appeared running, so job_run.start() never got called. https://github.com/Yelp/Tron/blob/f115e9502153de09dc36f93d2bbbf303a35bb606/tron/core/job_scheduler.py#L130 
* Jobs don't get queued correctly if the next scheduled time comes and the previous run is waiting for dependencies, because no action runs are active if it's waiting: https://github.com/Yelp/Tron/blob/f115e9502153de09dc36f93d2bbbf303a35bb606/tron/core/job_scheduler.py#L139

### DIFF
--- a/tron/core/actionrun.py
+++ b/tron/core/actionrun.py
@@ -142,6 +142,7 @@ class ActionRun(Observable):
     SKIPPED = 'skipped'
     STARTING = 'starting'
     SUCCEEDED = 'succeeded'
+    WAITING = 'waiting'
     UNKNOWN = 'unknown'
 
     default_transitions = dict(fail=FAILED, success=SUCCEEDED)
@@ -165,9 +166,15 @@ class ActionRun(Observable):
                     schedule=SCHEDULED,
                     **default_transitions,
                 ),
+            WAITING:
+                dict(
+                    cancel=CANCELLED,
+                    start=STARTING,
+                    **default_transitions,
+                ),
             SCHEDULED:
                 dict(
-                    ready=QUEUED,
+                    ready=WAITING,
                     queue=QUEUED,
                     cancel=CANCELLED,
                     start=STARTING,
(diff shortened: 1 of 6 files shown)


qui:
I tested this on stage. This screenshot shows that a run is "waiting", that the next one was still scheduled before the waiting one completed (issue in TRON-620), and the scheduled one became queued when the waiting one went overtime (issue in TRON-648). 

```
qui@tronplayground-uswest1adevc:~$ tronview MASTER.qui_test_overlap
Job                 : MASTER.qui_test_overlap
State               : running
Scheduler           : cron */3 * * * *
Max runtime         : None
Node Pool           : localhost (1 node(s))
Run on all nodes    : False
Allow overlapping   : False
Queue overlapping   : True

List of Actions:
one

Job runs:
Run ID    State       Node                          Scheduled Time           
.324      queued      batch@localhost               2019-02-26 15:54:00      
          Start: -  End: -  (-)
.323      waiting     batch@localhost               2019-02-26 15:51:00      
          Start: -  End: -  (-)
.322      cancelled   batch@localhost               2019-02-22 10:54:00      
          Start: -  End: -  (-)
```

tron/core/actionrun.py
```diff
@@ -918,6 +925,7 @@ def __init__(self, action_graph, run_map):
                 proxy.attr_proxy('is_scheduled', any),
                 proxy.attr_proxy('is_cancelled', any),
                 proxy.attr_proxy('is_active', any),
+                proxy.attr_proxy('is_waiting', all),
```
keymone: this is true only if all runs are in `waiting`, is this the logic we need?

qui: Right, changed to `any`. That way it's "waiting" in the following case:
  action1: succeeded
  action2: waiting

qui:
I'm pretty sure this is ready to go now, but would like to release https://github.com/Yelp/Tron/pull/624 first so it's easier to rollback and recover the state.

PR #631, opened 2019-03-15
fixed data validation of dynamodb migration 

EmanekaT:
I forgot that the value returned by restore has already been deserialized.
Also, I'm adding dynamodb to terraform so the autoscaling part is unnecessary.


tron/serialize/runstate/mirror_state_store.py
```diff
@@ -63,9 +66,12 @@ def dynamodb_restore(keys, res):
         thread1.join()
         thread2.join()
 
-        if not self.is_valid(shelve_kv_pairs, dynamo_kv_pairs):
-            self.dynamodb_store.alert('Data in dynamoDB is not synced to BerkleyDB. This is not critical \
-                                          since only BerkleyDB is used to restore states right now')
+        diffs = []
```
qui: In general it's better to not mutate arguments, and return this from the method (is_valid) instead.

EmanekaT: You are right. But I want to the function to only return true/false and save diffs in the meantime. I think it's fine rn as I will remove this file soon after migration.

PR #633, opened 2019-03-23
Make expected_runtime alerts page

solarkennedy:


tron/bin/check_tron_jobs.py
```diff
@@ -123,7 +123,7 @@ def compute_check_result_for_job_runs(client, job, job_content):
         status = 0
     elif last_state == State.STUCK:
         prefix = "WARN: Job exceeded expected runtime or still running when next job is scheduled"
```
chlgit: so it's not "WARN" anymore, right?

solarkennedy:
I would like to re-open this.

qui:
I agree that this is an important feature, but I am concerned about creating pages that are not actionable, especially because false positives can also hide real issues later.
* e.g. job is stuck, someone resolves the page. The sensu check is still failing for that job. The job fails later, but the team doesn't get paged immediately because the sensu check is already failing and realert_every is set to alert later.

Here are some reasons why this would page now:
* The job or action takes more than 24 hours normally and expected_runtime is not set. expected_runtime defaults to 24 hours.
* There is a run in progress when the next one is scheduled to run (overlapping). This can happen if things are running off-schedule for whatever reason, like a backfill or a retry after a job has failed the first time.

The first one isn't a super big deal - one solution is that those job owners add a real expected_runtime. Or we could take out the default in Tron and only page if someone has explicitly set an expected_runtime.

The second - perhaps we don't need alerting on overlapping runs?

Or perhaps there's a way I haven't thought of yet to address the issues around false positives - what are your thoughts?

solarkennedy:
> I agree that this is an important feature, but I am concerned about creating pages that are not actionable, especially because false positives can also hide real issues later.
> 
> * e.g. job is stuck, someone resolves the page. The sensu check is still failing for that job. The job fails later, but the team doesn't get paged immediately because the sensu check is already failing and realert_every is set to alert later.

What if we used different alert names for these?

> 
> Here are some reasons why this would page now:
> 
> * The job or action takes more than 24 hours normally and expected_runtime is not set. expected_runtime defaults to 24 hours.

I think this is a sane default, if users expect their runtime to be longer than this, then they can set their expected runtime. I would volunteer to do a mass edit to add expected_runtime to all the jobs that currently have a runtime > 24h.

> * There is a run in progress when the next one is scheduled to run (overlapping). This can happen if things are running off-schedule for whatever reason, like a backfill or a retry after a job has failed the first time.
> 
> The second - perhaps we don't need alerting on overlapping runs?

Would this page? I think the alert is currently designed to first look at the most relevant run, and *then* inspect it's runtime.  I agree though that a retry would extend the runtime, potentially past the expected runtime, and hence alert. I think that is "OK".

qui:
Sure, I think different check names could work. 

> I would volunteer to do a mass edit to add expected_runtime to all the jobs that currently have a runtime > 24h.
OK, sounds good.

> Would this page? I think the alert is currently designed to first look at the most relevant run, and then inspect it's runtime. I agree though that a retry would extend the runtime, potentially past the expected runtime, and hence alert. I think that is "OK".

Yes, it looks at the most relevant run, and starts by looking at the runtime. I'm referring to another check we have, after expected runtime, for deciding if a run is "stuck". I am not convinced we need it and would like to remove it: https://github.com/Yelp/Tron/blob/eaed662aa27555f488de334f59c4c26122d90dc3/tron/bin/check_tron_jobs.py#L238

I agree that it's OK that a retry that extends the runtime would alert.

PR #636, opened 2019-03-28
Make check_tron_jobs look at entire history by removing count arg

kawaiwanyelp:
### Description
- `check_tron_jobs` previously only checked the last 20 runs, using the min and max dates of those runs when the precious flag is set. However, if runs are out of order and a successful run for a checked date is older than the 20, the script will throw a false error.
- This change makes it so that `check_tron_jobs` looks at a job's entire history to fix the above problem. Given that we have a default maximum of 50 historical job, there shouldn't be a huge performance hit (it's just bucketing a few more things).

### Testing
make test

### DIFF
--- a/tests/bin/check_tron_jobs_test.py
+++ b/tests/bin/check_tron_jobs_test.py
@@ -1455,33 +1455,41 @@ class TestCheckPreciousJobs(TestCase):
         assert '2018.10.14' in run_buckets
         assert run_buckets['2018.10.14'] == []
 
-    @patch('check_tron_jobs.Client', autospec=True)
-    def test_compute_check_result_for_job_not_precious(self, mock_client):
+    @patch('tron.bin.check_tron_jobs.guess_realert_every', mock.Mock(return_value=1), autospec=None)
+    @patch('tron.bin.check_tron_jobs.Client', autospec=True)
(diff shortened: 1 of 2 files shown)


qui:
Increasing the history is fine (someone else reported that they had gotten paged for NOT_SCHEDULED because their scheduled job had fallen out of the top 20), but I don't think it solves the issue from the ticket about false alerts. The team will still get paged for the same reason ~30 runs from now, when we don't have history for the runs before the backfill anymore.

What if we didn't alert if there are no runs in a bucket? That would mean we only alert if there's a failed run for some day. It's unlikely that there would be no Tron run for a day (only if there's a widespread Tron bug, or if it was intentionally cancelled).

solarkennedy:
You are right @qui, it doesn't really fix the issue 100%.
I don't like the idea of not alerting for no runs in a bucket though. 

What if we were more explicit about how many days the script cared about? Right now we look at the buckets as if we don't even "know" what days to look out for.  We could instead explicitly look for N days in the past, and a green run for each of those days. If that number was more like 14, then the bug wouldn't happen as much, and for bonus we could make the output of this thing be much more obvious for oncall people:

```
CRIT: This job is missing a green run for $N-3
[N] -

kawaiwanyelp:
I'm going to merge this since it makes the script better. However, I'll still be considering a more permanent solution - @qui is right, this doesn't solve the issue 100%.

PR #640, opened 2019-04-04
Improve dynamodb read/write speed

EmanekaT:
1. Instead of reading num_partitions and then read contents one by one, it just fetch all partitions and read 100 at a time (limit of batch_get_item method).
Test done:
  Tested on mesosstage.

### DIFF
--- a/tron/serialize/runstate/dynamodb_state_store.py
+++ b/tron/serialize/runstate/dynamodb_state_store.py
@@ -35,6 +35,7 @@ class DynamoDBStateStore(object):
             first_items = self._get_first_items(keys)
             remaining_items = self._get_remaining_items(first_items)
             items = self._merge_items(first_items, remaining_items)
+            #TODO: remove this after berkleyDB is removed.
             for key in keys:
                 if str(key) in items:
                     tranlated_items[key] = items[str(key)]
--- a/tron/serialize/runstate/dynamodb_state_store.py
+++ b/tron/serialize/runstate/dynamodb_state_store.py
@@ -119,6 +119,19 @@ class DynamoDBStateStore(object):
         except Exception as e:
             self.alert(str(e))
 
+    def __setitem__(self, key: ShelveKey, val: bytes) -> None:
+        num_partitions = math.ceil(len(val) / OBJECT_SIZE)
+        with self.table.batch_writer() as batch:
+            for index in range(num_partitions):
+                batch.put_item(
+                    Item={
+                        'key': str(key),
+                        'index': index,
+                        'val': val[index * OBJECT_SIZE:min(index * OBJECT_SIZE + OBJECT_SIZE, len(val))],
+                        'num_partitions': num_partitions,
+                    }
+                )
+
     def _delete_item(self, key: ShelveKey) -> None:
         with self.table.batch_writer() as batch:
             for index in range(self._get_num_of_partitions(key)):
@@ -145,19 +158,6 @@ class DynamoDBStateStore(object):
             return int(partition.get('Item', {}).get('num_partitions', 0))
         except self.client.exceptions.ResourceNotFoundException:
             return 0
-
-    def __setitem__(self, key: ShelveKey, val: bytes) -> None:
-        num_partitions = math.ceil(len(val) / OBJECT_SIZE)
-        with self.table.batch_writer() as batch:
-            for index in range(num_partitions):
-                batch.put_item(
-                    Item={
-                        'key': str(key),
-                        'index': index,
-                        'val': val[index * OBJECT_SIZE:min(index * OBJECT_SIZE + OBJECT_SIZE, len(val))],
-                        'num_partitions': num_partitions,
-                    }
-                )
 
     def cleanup(self) -> None:
         return
--- a/tron/serialize/runstate/dynamodb_state_store.py
+++ b/tron/serialize/runstate/dynamodb_state_store.py
@@ -1,6 +1,7 @@
 import logging
 import math
 import pickle
+from collections import defaultdict
 
 import boto3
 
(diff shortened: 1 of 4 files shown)


tron/serialize/runstate/dynamodb_state_store.py
```diff
@@ -29,14 +30,18 @@ def restore(self, keys) -> dict:
...
         try:
-            items = zip(
-                keys,
-                (self[key] for key in keys),
-            )
+            first_items = self._get_first_items(keys)
+            remaining_items = self._get_remaining_items(first_items)
+            items = self._merge_items(first_items, remaining_items)
+            #TODO: remove this after berkleyDB is removed.
+            for key in keys:
```
qui: Why do we need this? Are the keys passed in not strings?

tron/serialize/runstate/dynamodb_state_store.py
```diff
@@ -29,14 +30,18 @@ def restore(self, keys) -> dict:
         Fetch all under the same parition key(keys).
         ret: <dict of key to states>
         """
+        tranlated_items = {}
```
qui: nit: translated

tron/serialize/runstate/dynamodb_state_store.py
```diff
@@ -58,28 +63,46 @@ def alert(self, msg: str):
...
+                    table_name: {
+                        'Keys': keys[i:min(len(keys), i + 100)],
+                        'ConsistentRead': True
+                    },
+                }
+            )['Responses'][table_name]
+            items.extend(vals)
+        return items
+
+    def _get_first_items(self, keys: list):
```
qui: How about renaming these _get_first_partitions, _get_remaining_partitions, and _merge_partitions?

tron/serialize/runstate/dynamodb_state_store.py
```diff
@@ -58,28 +63,46 @@ def alert(self, msg: str):
...
+        new_keys = [{'key': {'S': str(key)}, 'index': {'N': '0'}} for key in keys]
+        return self._get_items(new_keys)
+
+    def _get_remaining_items(self, items: list):
+        keys_for_remaining_items = []
+        for item in items:
+            keys_for_remaining_items.extend([{'key': {'S': str(item['key']['S'])}, 'index': {'N': str(i)}} for i in range(1, int(item['num_partitions']['N']))])
```
qui: nit: can you add some line breaks to make this more readable?

tron/serialize/runstate/dynamodb_state_store.py
```diff
@@ -89,7 +112,6 @@ def save(self, key_value_pairs) -> None:
         """
         try:
             for key, val in key_value_pairs:
-                self._delete_item(key)
```
qui: I don't think it's safe to assume that the data size will never go down below a partition boundary. How about checking if the number of partitions has changed and only deleting the extra ones if it has gone down, in __setitem__?

EmanekaT: It's ok if data size goes down. I have a num_partition in the database. So if data size goes down, num_partition is smaller and there would be no errors. The only issue is that the extra data are not deleted and left in the database. But I think that's not a big deal for us.

tron/serialize/runstate/dynamodb_state_store.py
EmanekaT: No. Keys passed in are ShelveKeys used by shelvestore. I use them in mirror_state_store. I'll change everything to string once we get rid of shelvestore.

tron/serialize/runstate/dynamodb_state_store.py
keymone: we definitely want to cleanup old data. either leave the _delete_item call or figure out the difference between partition numbers and remove the stale ones.

EmanekaT: I'm thinking about adding a batch/script to clean them up to make writing faster. We have a limit of 20s right now.

qui: How much slower would it be to delete the items?

If deleting is really a blocker, an alternative could be adding a TTL to the items in dynamodb, so they eventually get cleaned up if they haven't been updated in a long time.

EmanekaT: I believe it could take up to a few seconds. The main reason is that I don't think it's gonna save us much space anyway. Deleting extra partitions here don't really delete removed jobs. I think it's better to delete removed jobs together with a script when DynamoDB is not busy.

keymone: i'm not a fan of batch script to cleanup those - it's better to be correct at  the spot than relying on somebody cleaning up later.
i also wouldn't worry about 20s window - we only need to flush the buffer during that time, not write all of the state (it's already written). i can't imagine flushing the buffer taking anywhere near 20s (i might be wrong, worth  checking that).

EmanekaT: That makes sense as long as someone could clean up old jobs. I'll add delete back.

keymone: yeah, cleanup of old jobs is a separate issue.

tron/serialize/runstate/dynamodb_state_store.py
```diff
@@ -134,5 +146,18 @@ def _get_num_of_partitions(self, key: ShelveKey) -> int:
         except self.client.exceptions.ResourceNotFoundException:
             return 0
 
+    def __setitem__(self, key: ShelveKey, val: bytes) -> None:
```
keymone: can you move this above _delete_item to reduce the diff and avoid missing unintentional changes?

PR #641, opened 2019-04-05
check_tron_jobs no longer alerts if job is not scheduled

acoover:
### Changes
- `get_relevant_action` will never hit the last point where we need to return the last element from the `action_runs` - this only occurred when we would pass "SUCEEDED" as the value for `last_state`
- Removed logic from `get_relevant_run_and_state` which checked whether a job was scheduled and, if not, returned the "NOT_SCHEDULED" state

### Tests
- Removed tests invoking `get_relevant_run_and_state` since function no longer depends on `PRECIOUS_JOB_ATTR`
- `test_no_job_scheduled_or_queuing` checks that we consider that job in the "SUCEEDED" state

P.S. This is my first official github pull request, I'm so proud 😁

### DIFF
--- a/tron/bin/check_tron_jobs.py
+++ b/tron/bin/check_tron_jobs.py
@@ -29,7 +29,6 @@ class State(Enum):
     FAILED = "failed"
     STUCK = "stuck"
     NO_RUN_YET = "no_run_yet"
-    NOT_SCHEDULED = "not_scheduled"
     NO_RUNS_TO_CHECK = "no_runs_to_check"
     UNKNOWN = "unknown"
     SKIPPED = "skipped"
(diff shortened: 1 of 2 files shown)


tron/bin/check_tron_jobs.py
```diff
@@ -293,7 +276,7 @@ def get_relevant_action(*, action_runs, last_state, actions_expected_runtime):
                     return action_run
                 if action_state == 'running':
                     stuck_action_run_candidate = action_run
-    return stuck_action_run_candidate or action_runs[-1]
+    return stuck_action_run_candidate
```
qui: Why take out action_runs[-1]?

I think another part of the code assumes that the relevant action is not None: https://github.com/Yelp/Tron/blob/ddb1524315ea99c895d774844434249c1e6fc76f/tron/bin/check_tron_jobs.py#L108

chlgit: I can't come out with any example in which we should return action_runs[-1] except non-schedule case. I think we should have enough confidence as long as unit tests passing.

acoover: My interpretation of the code is similar to the one suggested by @chlgit - that `action_runs[-1]` exists purely to catch the case where `last_state=NOT_SCHEDULED` - however, it could also be the case that this clause is triggered when we pass `last_state=NO_RUNS_TO_CHECK` since I imagine that doesn't map to an actual `action_state`

I'll go ahead and leave this part in since I'm unfamiliar with this code and am choosing to be safe.

PR #643, opened 2019-04-15
set boto3 logging level to warning

EmanekaT:
All loggers 
https://fluffy.yelpcorp.com/i/kBxWnJxJ2NDk6G2d3t9LhvHQVNDw13cC.html


qui:
We should be handling all log configuration in https://github.com/Yelp/Tron/blob/master/tron/logging.conf

There are a couple issues:
1) we don't have logging configured for boto3 here, so it's being handled by the root handler
2) we don't use this file directly on servers, puppet creates a different one. And that one has the root handler set to DEBUG

We should update root handler to WARNING in puppet. Then, if we're ok with WARNING for boto messages, we don't have to make any changes here. If not, then we can add loggers for boto3 and botocore in this file and update settings as desired.

PR #645, opened 2019-04-17
Workaround for trond segfaults

vkhromov:
`trond` is rarely segfaulting at random moments when running under heavy
load with Python 3.6.0.  It seems that the crash is caused by
`task_processing.plugins.mesos.logging_executor` writes into a closed
filehandler inside `FileHandleWrapper`, triggering
https://bugs.python.org/issue31976.

The bug is fixed in Python 3.7.x, but since we cannot migrate from 3.6.x
at the moment, let's add a workaround by protesting all filehandler
operations by a lock, avoiding the bug as the result.

### DIFF
--- a/tron/serialize/filehandler.py
+++ b/tron/serialize/filehandler.py
@@ -12,7 +12,7 @@ import time
 from collections import OrderedDict
 from subprocess import PIPE
 from subprocess import Popen
-from threading import Lock
+from threading import RLock
 
 from tron.utils import maybe_encode
 
@@ -43,7 +43,7 @@ class FileHandleWrapper(object):
         self.manager = manager
         self.name = name
         self.last_accessed = time.time()
-        self._fh_lock = Lock()
+        self._fh_lock = RLock()
         self._fh = NullFileHandle
 
     def close(self):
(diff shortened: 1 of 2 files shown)


tron/serialize/filehandler.py
```diff
@@ -50,21 +52,23 @@ def close(self):
...
-                log.error("Failed to open %s: %s", self.name, e)
-                return
-
-        self.last_accessed = time.time()
-        self._fh.write(maybe_encode(content))
-        self.manager.update(self)
+        with self._fh_lock:
+            if self._fh == NullFileHandle:
+                try:
+                    self._fh = open(self.name, 'ab')
```
qui: If write gets called after close, the file handle will not be explicitly closed. Is that okay?

vkhromov: Probably not.  Let's print a warning in that case.

qui: OK. Another possibility would be to just open the file handle once, in init, and not re-open and write if the file has already been closed.

vkhromov: It seems that the idea of this class and `FileHandleManager` below is to close files when they are idle and to reopen them when they're accessed again. If we'll just open a file in `__init__` and don't close it from the manager, we probably could just remove both classes.

qui: Yeah, the reason I don't think we need it in this class is that the FileHandleManager just calls `close()` on FileHandleWrappers, which removes the FileHandleWrapper from its cache. So the FileHandleManager re-creates the FileHandleWrapper when a file is accessed again. No one is calling close_wrapped, which leaves the closed wrapper around.

I think that means we can let the FileHandleManager take care of the re-opening logic, and remove it from here. Up to you - we probably won't have too many open file handles around, but it would be nice to keep it clean if it makes sense.

qui:
Ah, I see the latest change. That works too! Thanks.

PR #646, opened 2019-04-17
Log state changes

qui:
Once we have this, we can set up a log handler to pull these into a separate log and ingest them wherever we want.

Some example lines:
```
{"id": "MASTER.testjob1.8.first", "type": "SSHActionRun", "state": "running", "timestamp": 1555533995.7740123}
{"id": "MASTER.testjob0.8.zeroth", "type": "SSHActionRun", "state": "unknown", "timestamp": 1555533996.7385292}
{"id": "MASTER.testjob0.8", "type": "JobRun", "state": "unknown", "timestamp": 1555533996.7392828}
```

### DIFF
--- a/tron/core/jobrun.py
+++ b/tron/core/jobrun.py
@@ -1,7 +1,9 @@
 """
  Classes to manage job runs.
 """
+import json
 import logging
+import time
 from collections import deque
 
 import tron.metrics as metrics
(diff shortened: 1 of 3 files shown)


tron/core/jobrun.py
```diff
@@ -231,6 +245,7 @@ def handle_action_run_state_change(self, action_run: ActionRun, event):
 
         # propagate all state changes (from action runs) up to state serializer
         self.notify(self.NOTIFY_STATE_CHANGED)
```
solarkennedy: Would it be easier to put this in the `notify` method? (the log is a type of notification?)

tron/core/jobrun.py
```diff
@@ -179,6 +192,7 @@ def seconds_until_run_time(self):
 
     def start(self):
         """Start this JobRun as a scheduled run (not a manual run)."""
```
solarkennedy: What method does a manual run invoke? Will it get logged?

qui: Yes, manual runs invoke this method too.

solarkennedy: Can you correct the docstring then?

tron/core/jobrun.py
qui: We won't have access to the action name at that point.

PR #647, opened 2019-04-17
"Remove" interval scheduler from documentation

EmanekaT:
I marked interval scheduler as removed and change all examples using interval scheduler to use cron. I used version 0.9.12.6 which would be the version of the next release.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -31,7 +31,7 @@ Basic Example
     jobs:
       - name: "getting_node_info"
         node: local
-        schedule: "interval 10 mins"
+        schedule: "cron */10 * * * *"
         actions:
           - name: "uname"
             command: "uname -a"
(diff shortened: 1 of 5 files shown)


solarkennedy:
I would rather you just remove all use of it and then remove the code entirely.

EmanekaT:
> I would rather you just remove all use of it and then remove the code entirely.

I think it's better to mark if as removed instead. In that way, if the user find interval scheduler somewhere in the wiki, they could check the doc and see why it is removed. Otherwise they might think our doc is incomplete until their changes failed check in yelpsoa_configs.

solarkennedy:
I promise, it is better to just remove it entirely.

We've removed all usage of it at Yelp. I don't want stale docs lying around referencing features that don't exist.

PR #649, opened 2019-04-19
duration field added to tronweb

drmorr0:
see the results at http://dev54-uswest1adevc:8089/web/#job/MASTER.testjob0


solarkennedy:
Can you do this for actions too to be consistent?

qui:
I agree it would be nice to have the duration of each action in the table of action runs, on the job run page.

PR #650, opened 2019-04-19
Use correct parser for months in cron

qui:


EmanekaT:
This could be in another ticket but it's better if we can add more tests and perhaps more function. Since people are using cron more heavily, I'm sure they will assume our cron is as powerful as the real one.

qui:
Our cron parser does support all standard cron formats - we were just running into errors because it wasn't parsing months correctly.

PR #651, opened 2019-04-22
reload old config if new config fails

drmorr0:
tested on my devbox and on mesosstage

One thing I noticed is that if the config file in /nail/tron/config/whatever.yaml has a bad value, tron can't restart.  That's probably a different issue, but maybe something worth fixing?

### DIFF
--- a/tron/api/controller.py
+++ b/tron/api/controller.py
@@ -2,7 +2,6 @@
 Web Controllers for the API.
 """
 import logging
-from copy import copy
 
 from tron import yaml
 from tron.eventbus import EventBus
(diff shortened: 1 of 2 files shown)


tron/api/controller.py
```diff
@@ -174,11 +175,19 @@ def update_config(self, name, content, config_hash):
         if self.config_manager.get_hash(name) != config_hash:
             return "Configuration has changed. Please try again."
 
+        old_config = copy(self.read_config(name)['config'])
```
qui: Do we need to copy, if we're reading from the file?

drmorr0: Oh, yea, I was being dumb here.

PR #652, opened 2019-04-22
Fixed guess_realert_every

EmanekaT:
Tron does not realert as expected after Tron jobs fail when no run ever starts. Check TRON-1104 for detail.
I updated tron such that it checks run_time when no start_time exists. No start_time exists when the job never runs. run_time is the scheduled time of the job.

### DIFF
--- a/tests/bin/check_tron_jobs_test.py
+++ b/tests/bin/check_tron_jobs_test.py
@@ -1337,11 +1337,15 @@ class TestCheckJobs(TestCase):
                         'MASTER.test.2',
                     'state':
                         'failed',
-                    'start_time': None,
-                    'run_time':
+                    'start_time':
                         time.strftime(
                             '%Y-%m-%d %H:%M:%S',
                             time.localtime(time.time() - 600),
+                        ),
+                    'run_time':
+                        time.strftime(
+                            '%Y-%m-%d %H:%M:%S',
+                            time.localtime(time.time() - 200),
                         ),
                 },
                 {
--- a/tests/bin/check_tron_jobs_test.py
+++ b/tests/bin/check_tron_jobs_test.py
@@ -1328,7 +1328,7 @@ class TestCheckJobs(TestCase):
                 ),
             'runs': [
                 {
-                    'id': 'MASTER.test.3',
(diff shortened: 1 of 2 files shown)


tron/bin/check_tron_jobs.py
```diff
@@ -286,14 +286,17 @@ def guess_realert_every(job):
             return -1
         job_runs = job.get('runs', [])
         job_runs_started = [
-            run for run in job_runs if run['start_time'] is not None
+            run['start_time'] for run in job_runs if run['start_time'] is not None
         ]
+        if len(job_runs_started) == 0 and job.get('run_time', None):
+            job_runs_started = [
+                job['run_time']
```
qui: run_time does not exist for a job, it's on each job_run

EmanekaT: Fixed it. I'm manually testing it rn. Will let you know once it passed.

tron/bin/check_tron_jobs.py
```diff
@@ -286,14 +286,17 @@ def guess_realert_every(job):
             return -1
         job_runs = job.get('runs', [])
         job_runs_started = [
-            run for run in job_runs if run['start_time'] is not None
+            run['start_time'] for run in job_runs if run['start_time'] is not None
```
qui: I think we should use `run['start_time'] or run.get('run_time')`.

With the current logic, if the run history looks like:
```
5 scheduled
4 failed with trigger timeout (no start time)
3 succeeded (has start time)
```
I think this script will return a realert_every interval based on the difference between 3 and 5, which is incorrect. The interval should be the same as the interval between each run.

tests/bin/check_tron_jobs_test.py
```diff
@@ -1317,6 +1317,54 @@ def test_guess_realert_every(self):
         realert_every = check_tron_jobs.guess_realert_every(job_runs)
         assert_equal(realert_every, 4)
 
+    def test_guess_realert_every_no_action_run_starts(self):
```
qui: This test doesn't test the new behavior you've added, it's logically the same as the test above (test_guess_realert_every) because the latest run has a valid start_time.

PR #655, opened 2019-04-26
Prevent a connection error in the recovery action from failing the action

qui:
If the recovery action fails, then the state of the action should be unknown, not failed, because we don't know what happened to the real action command.

Internal ticket TRON-1116. 

This just affects the logic in the handler: https://github.com/Yelp/Tron/blob/a068fe84474cf2a51b72ad31b98c1df419ea6d98/tron/core/actionrun.py#L742
Before this change, if the recovery command failed to connect, the ActionCommand would emit the FAILSTART event, which Tron thinks is safe to retry because nothing has happened yet. After this change, if it fails to connect, it emits a EXITING event with `exit_status=None`, which triggers `fail_unknown` as desired.

I tested this manually:
1. let a run start
2. shut down Tron
3. edit the node config for localhost with a different user (e.g. `user: batch`) so SSH authentication would fail
4. restart Tron and let it start recovery

Before this change, the affected action would get failed and retry. After, it stays unknown.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,6 +36,7 @@ example-cluster/_events/
 *.stderr
 dev/manhole.sock.lock
 dev/tron.pid
+dev/_events/
 
 # Generated debian artifacts
 debian/tron
--- a/Makefile
+++ b/Makefile
@@ -59,7 +59,7 @@ cluster_itests:
 	tox -e cluster_itests
 
 dev:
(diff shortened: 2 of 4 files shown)


qui:
Oh yeah, I also tested recovery in the normal case (without changing the login user) to make sure it works.

PR #658, opened 2019-05-02
fix guess_realert when next_run is the same as previous_run

EmanekaT:
I tested with real data as well but didn't use it here since that is too long. 
The bug happens when run_time is the same as next_run, 1 is returned.

### DIFF
--- a/tests/bin/check_tron_jobs_test.py
+++ b/tests/bin/check_tron_jobs_test.py
@@ -1289,6 +1289,11 @@ class TestCheckJobs(TestCase):
                     'id': 'MASTER.test.3',
                     'state': 'scheduled',
                     'start_time': None,
+                    'run_time':
+                        time.strftime(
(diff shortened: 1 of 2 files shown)


tests/bin/check_tron_jobs_test.py
```diff
@@ -1317,6 +1317,44 @@ def test_guess_realert_every(self):
...
+                time.strftime(
+                    '%Y-%m-%d %H:%M:%S',
+                    time.localtime(time.time() + 600),
+                ),
+            'runs': [
+                {
+                    'id':
+                        'MASTER.test.2',
+                    'state':
+                        'failed',
```
qui: Instead of making a separate test, can you add 'run_time' to `'MASTER.test.3'` in the test above? That's the usual case for this error, the first run in the list is the scheduled run, which has the same run_time as next_run.

PR #661, opened 2019-05-07
Bump dependencies

keymone:
- address GH security alerts
- remove sqlalchemy dependency and related code
- import tron.yaml instead of yaml everywhere


qui:
failing test :/ did the API of sqlalchemy change? we could also just delete that state store, unless you think it will be useful for local testing once we get rid of berkeleydb.

keymone:
i'm in favor of removing sqlalchemy store

requirements.txt
```diff
@@ -69,9 +69,6 @@ rsa==4.0
 s3transfer==0.2.0
 setuptools==41.0.1
 six==1.12.0
-SQLAlchemy==1.3.3
-task-processing==0.1.5
```
qui: Don't we still need this?

keymone: thanks!

PR #663, opened 2019-05-08
Build for bionic

jvperrin:
- Added `debian/.debhelper/` to `.gitignore` (and sorted the things under `debian/` for good measure)
- Start building for bionic (same `Dockerfile` as for xenial, but I removed the deadsnakes stuff since [bionic ships with python 3.6](https://packages.ubuntu.com/bionic/python3.6) and also removed the dh-virtualenv wget installation since [bionic ships with the same version, 1.0-1](https://packages.ubuntu.com/bionic/dh-virtualenv))

Testing:
- Ran `make itest_bionic` and it passed successfully

Is there anything else I should update? I didn't touch some of the stuff like [this](https://github.com/Yelp/Tron/blob/ba6eee0094101e0a8b26dd6e66f5ca6ccb0ca3f7/Makefile#L67-L78) since it doesn't really affect building an actual tron bionic package.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,12 +39,13 @@ dev/tron.pid
 dev/_events/
 
 # Generated debian artifacts
-debian/tron
-debian/tron.preinst.debhelper
+debian/.debhelper/
+debian/debhelper-build-stamp
 debian/files
-debian/tron.substvars
+debian/tron
 debian/tron.debhelper.log
 debian/tron.postinst.debhelper
 debian/tron.postrm.debhelper
+debian/tron.preinst.debhelper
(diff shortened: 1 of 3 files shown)


solarkennedy:
Green checkmark means it probably works fine!

EmanekaT:
I agree green checks mean it probably works. But maybe it's worth checking if it works by running on jenkins pipeline after it's merged? I remember we had issue building trusty images on jenkins before even thought travis builds are successful.

jvperrin:
Agreed, I'll keep an eye on the jenkins pipeline and see how it does. Thanks for the reviews!

PR #666, opened 2019-05-10
Make recovery batch script also check if action runner suddenly goes away

kawaiwanyelp:
### Description
If the action runner suddenly goes away, the status file might not receive a return code. In addition to watching for a return code, I've made the recovery script also check if the action runner is still around. If it isn't, we exit with a code 1.

### Testing
make test (made a `recover_batch_test.py` file)

### DIFF
--- a/tron/bin/recover_batch.py
+++ b/tron/bin/recover_batch.py
@@ -35,12 +35,36 @@ def parse_args():
 
 def notify(notify_queue, ignored, filepath, mask):
     with open(filepath.path) as f:
-        last_entry = f.readlines()[-1]
-        x = yaml.load(last_entry)
-        return_code = x.get('return_code')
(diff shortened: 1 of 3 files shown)


tron/bin/recover_batch.py
```diff
@@ -35,12 +35,36 @@ def parse_args():
...
+                    f'{signal.Signals(exit_code).name}'
+                )
+            else:
+                exit_code = return_code
+
+        elif not psutil.pid_exists(pid):
+            exit_code = 1
+            error_msg = (
+                f'Action runner pid {pid} no longer running; '
+                'unable to recover batch'
```
solarkennedy: I don't like the word 'batch' in this context. Can you just say 'action run' ?

PR #669, opened 2019-05-16
Enable page for dynamodb alerts

EmanekaT:
As title. I'll remove the sensu alerts in the end and only have the error logs + an external alert. But before that, I'll use these sensu alerts instead. I'll not deploy dynamodb on prod before I finish and tested my dynamodb backup service and the new stateless alert.


qui:
Can you document what the oncall should do if they receive this page?

EmanekaT:
> Can you document what the oncall should do if they receive this page?

I added documents on tron runbook. I added a link to tron runbook. I also changed dev/config/master.yaml to use shelve so make dev works.

PR #671, opened 2019-05-21
Set up periodic backup for DynamoDB

EmanekaT:
This is for TRON-1130. 
The puppet side is here https://reviewboard.yelpcorp.com/r/407414/
I tested this on mesosstage.  
I didn't add any error checking and assume cron will notify us if that happens. And it's not critical even if it fails.

image

### DIFF
--- a/tron/bin/ddbbackup.py
+++ b/tron/bin/ddbbackup.py
@@ -98,7 +98,7 @@ def main():
         max_backups = args.max_backups
         ddb = boto3.client('dynamodb', region_name=region_name)
         #create backup and remove old ones
-        create_backup(ddb, table_name, max_backups)
+        create_backup(ddb, table_name)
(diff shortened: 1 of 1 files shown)


tron/bin/ddbbackup.py
```diff
@@ -0,0 +1,106 @@
...
+    )
+    return args
+
+
+def read_config(args):
+    return manager.ConfigManager(args.config_path).load().get_master().state_persistence
+
+
+def main():
+    # Fetch configs. Arguments are passed from systemctl. You can find them in loggging.conf
```
qui: Is it still true that the arguments come from systemctl / logging.conf?
Seems like they just come from the command in puppet, in which case, I don't think you need to add an extra explanation.

EmanekaT: Oh thx. I forgot to remove the comments.

PR #672, opened 2019-05-24
add the name of the job used by tron_stateless_alert as timestamp in yelpsoa_config

EmanekaT:
Then, i'll add a cron job that reads the job from dynamodb and check timestamp.  I might add more options or move this to another namespace later. But for now, I just want to update gitolite so I can test my script on mesosstage easier.


EmanekaT:
We decide to put it to hierra

PR #673, opened 2019-05-28
Add tron_stateless_alert for dynamodb

EmanekaT:
Tested on mesosstage. This will be a cron job defined on puppet.
The job runs every 30 mins and checks if the last timestamp of the last job run is more than 30 mins (default) or not. It does not check/alert when there is no job run.

### DIFF
--- a/tron/bin/check_tron_datastore_staleness.py
+++ b/tron/bin/check_tron_datastore_staleness.py
@@ -5,6 +5,8 @@ import os
 import sys
 import time
 
+import pytz
+
 from tron.config import manager
 from tron.config import schema
 from tron.serialize.runstate.dynamodb_state_store import DynamoDBStateStore
@@ -52,7 +54,7 @@ def parse_cli():
     parser.add_argument(
         "--staleness-threshold",
         default=DEFAULT_STALENESS_THRESHOLD,
-        help="how long (in mins) to wait to alert after the last timestamp of the job",
+        help="how long (in seconds) to wait to alert after the last timestamp of the job",
     )
     args = parser.parse_args()
     args.working_dir = os.path.abspath(args.working_dir)
@@ -80,11 +82,11 @@ def main():
         dynamodb_region = persistence_config.dynamodb_region
         table_name = persistence_config.table_name
         store = DynamoDBStateStore(table_name, dynamodb_region)
-        key = store.build_key('job_state', '{}'.format(job_for_stateless_alert))
+        key = store.build_key('job_state', job_for_stateless_alert)
         try:
             job = store.restore([key])[key]
         except Exception as e:
-            logging.error(f'Failed to retreive status for job {job_for_stateless_alert} due to {str(e)}')
+            logging.error(f'Failed to retreive status for job {job_for_stateless_alert} due to {e}')
             sys.exit(1)
 
         # Exit if the job never runs.
@@ -94,15 +96,15 @@ def main():
             sys.exit(1)
 
         # Alert if timestamp is not updated after staleness_threshold
-        stateless_for_secs = time.time() - last_run_time.timestamp()
+        stateless_for_secs = time.time() - last_run_time.astimezone(pytz.utc).timestamp()
         if stateless_for_secs > args.staleness_threshold:
-            logging.error(f'DynamoDB has not been updated for {stateless_for_secs} mins')
+            logging.error(f'{key} has not been updated in DynamoDB for {stateless_for_secs} seconds')
             sys.exit(1)
+        logging.info(f"DynamoDB is up to date. It's last updated at {str(last_run_time)}")
     # Alert for BerkeleyDB
     elif store_type == schema.StatePersistenceTypes.shelve:
         os.execl('/usr/lib/nagios/plugins/check_file_age', '/nail/tron/tron_state', '-w', '1800', '-c', '1800')
 
-    logging.info(f"DynamoDB is up to date. It's last updated at {str(last_run_time)}")
     sys.exit(0)
 
 
--- a/tron/bin/check_tron_datastore_staleness.py
+++ b/tron/bin/check_tron_datastore_staleness.py
@@ -14,8 +14,8 @@ from tron.serialize.runstate.dynamodb_state_store import DynamoDBStateStore
 # Default values for arguments
 DEFAULT_WORKING_DIR = '/var/lib/tron/'
 DEFAULT_CONF_PATH = 'config/'
-DEFAULT_STALENESS_THRESHOLD = 30
-log = logging.getLogger('tron_stateless_alert')
+DEFAULT_STALENESS_THRESHOLD = 1800
+log = logging.getLogger('check_tron_datastore_staleness')
 
 
 def get_last_run_time(job):
@@ -28,8 +28,7 @@ def get_last_run_time(job):
         for action in run['runs']:
             if action.get('start_time') and action.get('state') != 'scheduled':
                 timestamps.append(action.get('start_time'))
-    timestamps.sort()
-    return timestamps[-1] if timestamps else None
+    return max(timestamps) if timestamps else None
 
 
 def parse_cli():
@@ -47,7 +46,7 @@ def parse_cli():
         help="File path to the Tron configuration file",
     )
     parser.add_argument(
-        "--job-for-stateless-alert",
+        "--job-name",
         required=True,
         help="The job name to read timestamp from",
     )
@@ -74,7 +73,7 @@ def main():
     args = parse_cli()
     persistence_config = read_config(args)
     store_type = schema.StatePersistenceTypes(persistence_config.store_type)
-    job_for_stateless_alert = args.job_for_stateless_alert
+    job_name = args.job_name
 
     # Alert for DynamoDB
     if store_type == schema.StatePersistenceTypes.dynamodb:
@@ -82,11 +81,11 @@ def main():
         dynamodb_region = persistence_config.dynamodb_region
         table_name = persistence_config.table_name
         store = DynamoDBStateStore(table_name, dynamodb_region)
-        key = store.build_key('job_state', job_for_stateless_alert)
+        key = store.build_key('job_state', job_name)
         try:
             job = store.restore([key])[key]
         except Exception as e:
-            logging.error(f'Failed to retreive status for job {job_for_stateless_alert} due to {e}')
+            logging.exception(f'Failed to retreive status for job {job_name} due to {e}')
             sys.exit(1)
 
         # Exit if the job never runs.
@@ -100,10 +99,10 @@ def main():
         if stateless_for_secs > args.staleness_threshold:
             logging.error(f'{key} has not been updated in DynamoDB for {stateless_for_secs} seconds')
             sys.exit(1)
-        logging.info(f"DynamoDB is up to date. It's last updated at {str(last_run_time)}")
+        logging.info(f"DynamoDB is up to date. It's last updated at {last_run_time}")
     # Alert for BerkeleyDB
     elif store_type == schema.StatePersistenceTypes.shelve:
-        os.execl('/usr/lib/nagios/plugins/check_file_age', '/nail/tron/tron_state', '-w', '1800', '-c', '1800')
+        os.execl('/usr/lib/nagios/plugins/check_file_age', '/nail/tron/tron_state', '-w', str(args.staleness_threshold), '-c', str(args.staleness_threshold))
 
     sys.exit(0)
 
--- a/tron/bin/check_tron_dynamodb_staleness.py
+++ b/tron/bin/check_tron_dynamodb_staleness.py
@@ -2,6 +2,7 @@
 import argparse
 import logging
 import os
+import sys
 import time
 
 from tron.config import manager
@@ -11,7 +12,7 @@ from tron.serialize.runstate.dynamodb_state_store import DynamoDBStateStore
 # Default values for arguments
 DEFAULT_WORKING_DIR = '/var/lib/tron/'
 DEFAULT_CONF_PATH = 'config/'
-DEFAULT_ALERT_AFTER = 30
+DEFAULT_STALENESS_THRESHOLD = 30
 log = logging.getLogger('tron_stateless_alert')
 
 
@@ -49,8 +50,8 @@ def parse_cli():
         help="The job name to read timestamp from",
     )
     parser.add_argument(
-        "--alert-after",
-        default=DEFAULT_ALERT_AFTER,
+        "--staleness-threshold",
+        default=DEFAULT_STALENESS_THRESHOLD,
         help="how long (in mins) to wait to alert after the last timestamp of the job",
     )
     args = parser.parse_args()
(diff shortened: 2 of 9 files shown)


tron/bin/tron_stateless_alert.py
```diff
@@ -0,0 +1,141 @@
+#!/usr/bin/env python3.6
```
solarkennedy: Can you rename this command to `check_tron_dynamodb_staleness` ?

tron/bin/tron_stateless_alert.py
```diff
@@ -0,0 +1,141 @@
...
+
+    time_diff = (
+        now -
+        time.mktime(_timestamp_to_timeobj(trim_nanoseconds(str(last_run_time))))
+    )
+    # secs to mins
+    return time_diff / 60
+
+
+def alert(check_every, msg):
```
solarkennedy: There is no need to make this check send an alert itself, just have it return a 0 or 1 in sys.exit()

tron/bin/tron_stateless_alert.py
```diff
@@ -0,0 +1,141 @@
...
+
+def main():
+    # Fetch configs. You can find the arguments in puppet.
+    args = parse_cli()
+    persistence_config = read_config(args)
+    store_type = schema.StatePersistenceTypes(persistence_config.store_type)
+    job_for_stateless_alert = args.job_for_stateless_alert
+
+    # Make sure datastore is dynamodb
+    if store_type == schema.StatePersistenceTypes.dynamodb:
```
solarkennedy: It would be kinda cool if this check could just automatically check the staleness for any type of backend.
That would simplify the puppet code

tron/bin/tron_stateless_alert.py
```diff
@@ -0,0 +1,141 @@
...
+
+    def _timestamp_to_timeobj(timestamp):
+        return time.strptime(timestamp, '%Y-%m-%d %H:%M:%S')
+
+    time_diff = (
+        now -
+        time.mktime(_timestamp_to_timeobj(trim_nanoseconds(str(last_run_time))))
+    )
+    # secs to mins
+    return time_diff / 60
```
solarkennedy: What if this function was just called `calculate_seconds_since_last_run_time` ?
Then it would be easier to understand how this function relates to the rest of the code in `main`

tron/bin/tron_stateless_alert.py
```diff
@@ -0,0 +1,141 @@
...
+        default=DEFAULT_CONF_PATH,
+        help="File path to the Tron configuration file",
+    )
+    parser.add_argument(
+        "--job-for-stateless-alert",
+        required=True,
+        help="The job name to read timestamp from",
+    )
+    parser.add_argument(
+        "--alert-after",
```
solarkennedy: Don't use `alert_after` for this.
How about `staleness_threshold` and make it in seconds?

tron/bin/tron_stateless_alert.py
```diff
@@ -0,0 +1,141 @@
...
+        default=DEFAULT_WORKING_DIR,
+        help="Working directory for the Tron daemon, default %(default)s",
+    )
+    parser.add_argument(
+        "-c",
+        "--config-path",
+        default=DEFAULT_CONF_PATH,
+        help="File path to the Tron configuration file",
+    )
+    parser.add_argument(
```
solarkennedy: Do need a sentinel job?
Is there a generic way to measure the last updated timestamp without looking at a particular job, analogous to how on shelve we look at the timestamp fo the file?

tron/bin/tron_stateless_alert.py
EmanekaT: Sure, I could import berkleydb here.

tron/bin/tron_stateless_alert.py
EmanekaT: Unfortunately, DynamoDB does not provide any generic timestamp. While I could scan the table to find the latest timestamp, that slows things down especially in norcal-prod where we have tons of data.

tron/bin/tron_stateless_alert.py
```diff
@@ -0,0 +1,141 @@
...
+        args.config_path,
+    )
+    return args
+
+
+def read_config(args):
+    return manager.ConfigManager(args.config_path).load().get_master().state_persistence
+
+
+def calc_time_diff(last_run_time, now):
```
solarkennedy: I think this is way overkill.

I think it could just be one line?

   return (datetime.utcnow() - dateutil.parser.parse(last_run_time)).seconds

Can you confirm this timestamp is in utc?

tron/bin/tron_stateless_alert.py
```diff
@@ -0,0 +1,141 @@
...
+        dynamodb_region = persistence_config.dynamodb_region
+        table_name = persistence_config.table_name
+        # Fetch job state from dynamodb
+        store = DynamoDBStateStore(table_name, dynamodb_region)
+        key = store.build_key('job_state', '{}'.format(job_for_stateless_alert))
+        try:
+            job = store.restore([key])
+        except Exception as e:
+            alert(
+                args.alert_after,
```
solarkennedy: Just use `f` strings and print to stderr

tron/bin/tron_stateless_alert.py
```diff
@@ -0,0 +1,141 @@
...
+                'Failed to retreive status for job {} due to {}'.format(job_for_stateless_alert, str(e))
+            )
+
+        # Alert if timestamp is not updated after alert_after mins
+        # Skip the check if the job never runs. This is because the intended job used as
+        # timestamp is paasta-contract-monitor, which alerts if it does not run
+        last_run_time = get_last_run_time(job)
+        if last_run_time:
```
solarkennedy: I think it is ok to crash if we don't have a last runtime

tron/bin/tron_stateless_alert.py
solarkennedy: Just check the timestamp on the file, just like we do now.
Heck you could literally call `os.system` with the exact same command as we run now.

tron/bin/check_tron_dynamodb_staleness.py
```diff
@@ -0,0 +1,110 @@
...
+        # Alert if timestamp is not updated after staleness_threshold
+        stateless_for_secs = time.time() - last_run_time.timestamp()
+        if stateless_for_secs > args.staleness_threshold:
+            logging.error(f'DynamoDB has not been updated for {stateless_for_secs} mins')
+            sys.exit(1)
+    # Alert for BerkeleyDB
+    elif store_type == schema.StatePersistenceTypes.shelve:
+        os.execl('/usr/lib/nagios/plugins/check_file_age', '/nail/tron/tron_state', '-w', '1800', '-c', '1800')
+
+    logging.info(f"DynamoDB is up to date. It's last updated at {str(last_run_time)}")
```
solarkennedy: this can go under the dynamo case

tron/bin/check_tron_dynamodb_staleness.py
```diff
@@ -0,0 +1,110 @@
...
+            sys.exit(1)
+
+        # Exit if the job never runs.
+        last_run_time = get_last_run_time(job)
+        if not last_run_time:
+            logging.error(f'No last run for {key} found. If the job was just added, it might take some time for it to run')
+            sys.exit(1)
+
+        # Alert if timestamp is not updated after staleness_threshold
+        stateless_for_secs = time.time() - last_run_time.timestamp()
```
solarkennedy: can you confirm that the `last_run_time.timestamp()` is in UTC?

tron/bin/check_tron_dynamodb_staleness.py
```diff
@@ -0,0 +1,110 @@
...
+
+    # Alert for DynamoDB
+    if store_type == schema.StatePersistenceTypes.dynamodb:
+        # Fetch job state from dynamodb
+        dynamodb_region = persistence_config.dynamodb_region
+        table_name = persistence_config.table_name
+        store = DynamoDBStateStore(table_name, dynamodb_region)
+        key = store.build_key('job_state', '{}'.format(job_for_stateless_alert))
+        try:
+            job = store.restore([key])[key]
```
solarkennedy: Can you confirm the syntax here is `.restore([key])[key]` ?
The restore argument takes a list and returns a dictionary that needs a key lookup again?

tron/bin/check_tron_dynamodb_staleness.py
```diff
@@ -0,0 +1,110 @@
...
+    store_type = schema.StatePersistenceTypes(persistence_config.store_type)
+    job_for_stateless_alert = args.job_for_stateless_alert
+
+    # Alert for DynamoDB
+    if store_type == schema.StatePersistenceTypes.dynamodb:
+        # Fetch job state from dynamodb
+        dynamodb_region = persistence_config.dynamodb_region
+        table_name = persistence_config.table_name
+        store = DynamoDBStateStore(table_name, dynamodb_region)
+        key = store.build_key('job_state', '{}'.format(job_for_stateless_alert))
```
solarkennedy: You can remove `'{}.format(..` and just give the variable name

tron/bin/check_tron_dynamodb_staleness.py
```diff
@@ -0,0 +1,110 @@
...
+        # Exit if the job never runs.
+        last_run_time = get_last_run_time(job)
+        if not last_run_time:
+            logging.error(f'No last run for {key} found. If the job was just added, it might take some time for it to run')
+            sys.exit(1)
+
+        # Alert if timestamp is not updated after staleness_threshold
+        stateless_for_secs = time.time() - last_run_time.timestamp()
+        if stateless_for_secs > args.staleness_threshold:
+            logging.error(f'DynamoDB has not been updated for {stateless_for_secs} mins')
```
solarkennedy: can you include the job name key here too, to remind the responder exactly what it was looking for?

tron/bin/check_tron_dynamodb_staleness.py
```diff
@@ -0,0 +1,110 @@
+#!/usr/bin/env python3.6
```
solarkennedy: The new name of this check should just be `check_tron_datastore_staleness`

tron/bin/check_tron_dynamodb_staleness.py
EmanekaT: The store returns the dictionary, and the key is to retrieve the job from the dictionary. I tested on mesosstage.

tron/bin/check_tron_dynamodb_staleness.py
EmanekaT: It's not in UTC. last_run_time is a datetime object with tzinfo. And timestamp returns epoch time based on that. I tested on mesosstage to make sure the time diff is correct.

EmanekaT: hmm, it's not. I need to use pytz.

tron/bin/check_tron_datastore_staleness.py
```diff
@@ -0,0 +1,112 @@
...
+        help="Working directory for the Tron daemon, default %(default)s",
+    )
+    parser.add_argument(
+        "-c",
+        "--config-path",
+        default=DEFAULT_CONF_PATH,
+        help="File path to the Tron configuration file",
+    )
+    parser.add_argument(
+        "--job-for-stateless-alert",
```
qui: nit: staleness alert, not stateless
Might be able to just shorten it to `--job`? up to you

tron/bin/check_tron_datastore_staleness.py
```diff
@@ -0,0 +1,112 @@
...
+import pytz
+
+from tron.config import manager
+from tron.config import schema
+from tron.serialize.runstate.dynamodb_state_store import DynamoDBStateStore
+
+# Default values for arguments
+DEFAULT_WORKING_DIR = '/var/lib/tron/'
+DEFAULT_CONF_PATH = 'config/'
+DEFAULT_STALENESS_THRESHOLD = 30
```
qui: The default is 30 seconds? Do you mean minutes?

tron/bin/check_tron_datastore_staleness.py
```diff
@@ -0,0 +1,112 @@
...
+
+        # Alert if timestamp is not updated after staleness_threshold
+        stateless_for_secs = time.time() - last_run_time.astimezone(pytz.utc).timestamp()
+        if stateless_for_secs > args.staleness_threshold:
+            logging.error(f'{key} has not been updated in DynamoDB for {stateless_for_secs} seconds')
+            sys.exit(1)
+        logging.info(f"DynamoDB is up to date. It's last updated at {str(last_run_time)}")
+    # Alert for BerkeleyDB
+    elif store_type == schema.StatePersistenceTypes.shelve:
+        os.execl('/usr/lib/nagios/plugins/check_file_age', '/nail/tron/tron_state', '-w', '1800', '-c', '1800')
```
qui: Should this also use the `--staleness-threshold` for consistency? (instead of hardcoding 1800 seconds)
also `--working-dir`, and the state file name

tron/bin/check_tron_datastore_staleness.py
```diff
@@ -0,0 +1,112 @@
...
+    if store_type == schema.StatePersistenceTypes.dynamodb:
+        # Fetch job state from dynamodb
+        dynamodb_region = persistence_config.dynamodb_region
+        table_name = persistence_config.table_name
+        store = DynamoDBStateStore(table_name, dynamodb_region)
+        key = store.build_key('job_state', job_for_stateless_alert)
+        try:
+            job = store.restore([key])[key]
+        except Exception as e:
+            logging.error(f'Failed to retreive status for job {job_for_stateless_alert} due to {e}')
```
qui: It's usually better to print exceptions with repr, e.g `{e!r}`, because that provides more information. For example, for `e = KeyError('a_key')`, str(e) is `'a_key'` but repr(e) is `KeyError('a_key')`

You could also just use `logging.exception` to show the traceback.

tron/bin/check_tron_datastore_staleness.py
```diff
@@ -0,0 +1,112 @@
...
+
+from tron.config import manager
+from tron.config import schema
+from tron.serialize.runstate.dynamodb_state_store import DynamoDBStateStore
+
+# Default values for arguments
+DEFAULT_WORKING_DIR = '/var/lib/tron/'
+DEFAULT_CONF_PATH = 'config/'
+DEFAULT_STALENESS_THRESHOLD = 30
+log = logging.getLogger('tron_stateless_alert')
```
qui: Why not use `__name__`?
Otherwise you should update this now, since the name of the script has changed.

tron/bin/check_tron_datastore_staleness.py
```diff
@@ -0,0 +1,112 @@
...
+    '''
+    Get all sorted timestamps, and only count the actions that actually ran
+    '''
+    timestamps = []
+    job_runs = job['runs']
+    for run in job_runs:
+        for action in run['runs']:
+            if action.get('start_time') and action.get('state') != 'scheduled':
+                timestamps.append(action.get('start_time'))
+    timestamps.sort()
```
qui: You shouldn't need to sort this, can't you just find the max?

tron/bin/check_tron_datastore_staleness.py
EmanekaT: Shouldn't __name__ be '__main'__'? Thx for pointing it out, I shall update it.

tron/bin/check_tron_datastore_staleness.py
EmanekaT: Oh right, I forgot python has this function.

PR #674, opened 2019-05-29
Preserve last run's scheduled time when reconfiguring

kawaiwanyelp:
### Description
- When a job run is cancelled, Tron will schedule the next run after the cancelled run. However, if the  job is reconfigured , there is a bug that makes it so that the recreated scheduled run is instead scheduled to run at the same time as the cancelled job.
- To fix this bug, I have made it so that when recreating the scheduled run, Tron copies the existing the scheduled run's run time to the new scheduled run, thus preserving the correct time.

### Tests
manual testing with `make dev`
`make test`

### DIFF
--- a/tests/core/job_test.py
+++ b/tests/core/job_test.py
@@ -272,14 +272,48 @@ class TestJobScheduler(TestCase):
             calls = [mock.call(mock_runs[i]) for i in range(0, len(mock_runs))]
             self.job.watch.assert_has_calls(calls)
 
+    def test_create_and_schedule_runs_specific_time(self):
+        self.job_scheduler.get_runs_to_schedule = mock.Mock(return_value=[mock.Mock()])
+        self.job_scheduler._set_callback = mock.Mock()
+        self.job_scheduler.create_and_schedule_runs(next_run_time='a_datetime')
+        assert self.job_scheduler.get_runs_to_schedule.call_args_list == [mock.call('a_datetime')]
+
+    def test_create_and_schedule_runs_guess(self):
+        self.job_scheduler.get_runs_to_schedule = mock.Mock(return_value=[mock.Mock()])
+        self.job_scheduler._set_callback = mock.Mock()
+        self.job_scheduler.create_and_schedule_runs(next_run_time=None)
+        assert self.job_scheduler.get_runs_to_schedule.call_args_list == [mock.call(None)]
+
     def test_disable(self):
         self.job_scheduler.disable()
         assert self.job_scheduler.job.enabled is False
         self.job_scheduler.job.runs.cancel_pending.assert_called_once()
 
     def test_schedule_reconfigured(self):
+        job_run = mock.Mock()
+        job_run.run_time = 'a_run_time'
+        self.job.runs.get_pending.return_value = [job_run]
+        self.job_scheduler.create_and_schedule_runs = mock.Mock()
+
         self.job_scheduler.schedule_reconfigured()
-        self.job.runs.remove_pending.assert_called_once()
+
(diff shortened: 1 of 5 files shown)


solarkennedy:
Can you check out the failing tests?

kawaiwanyelp:
> Can you check out the failing tests?

tron/bin/recover_batch.py
```diff
@@ -10,6 +10,10 @@
 from twisted.internet import inotify
 from twisted.internet import reactor
 from twisted.python import filepath
+try:
+    from yaml import CSafeLoader as SafeLoader
+except ImportError:
```
solarkennedy: I don't really want this here.

Can you use `tron.yaml` instead like we do elsewhere in the codebase?

tron/core/job_scheduler.py
```diff
@@ -49,11 +49,18 @@ def enable(self):
             return
 
         self.job.enabled = True
-        self.create_and_schedule_runs(ignore_last_run_time=True)
+        self.create_and_schedule_runs(next_run_time='current')
 
-    def create_and_schedule_runs(self, ignore_last_run_time=False):
-        for job_run in self.get_runs_to_schedule(ignore_last_run_time):
-            self._set_callback(job_run)
+    def create_and_schedule_runs(self, next_run_time=None):
```
qui: Feels like a bad practice to accept different types of arguments here: the string 'current', None (which means last run time), or a datetime... What if it just accepted a datetime or None? 

Debatable what makes sense as a default if the next_run_time is None, but I think "current" makes sense because that's the default if you pass in None to `self.job.scheduler.next_run_time` (in `get_runs_to_schedule`). 

Then you'd have to move the logic about getting the last_run_time to the `schedule` method.

tron/core/job_scheduler.py
```diff
@@ -73,9 +80,17 @@ def manual_start(self, run_time=None):
...
+        if not self.job.enabled:
+            return
+
+        # when reconfiguring, preserve the latest scheduled run's time
+        pending_run_times = [j.run_time for j in list(self.job.runs.get_pending())]
+        if len(pending_run_times) != 1:
+            log.warning(f'{self.job} has {len(pending_run_times)} pending runs, not 1')
+        next_run_time = 'current' if len(pending_run_times) == 0 else pending_run_times[0]
+
+        self.job.runs.remove_pending()
```
qui: Maybe this could return the runs it removes?

tron/bin/recover_batch.py
kawaiwanyelp: Sure.

tron/core/job_scheduler.py
kawaiwanyelp: Done.

tron/core/job_scheduler.py
kawaiwanyelp: That wasn't the case before the changes. Why do you want this?

qui: Oh, I was suggesting you could use the run time from the removed pending runs, instead of having to find it separately.

tests/core/job_test.py
```diff
@@ -272,14 +272,48 @@ def test_restore_state_sets_job_runs(self):
             calls = [mock.call(mock_runs[i]) for i in range(0, len(mock_runs))]
             self.job.watch.assert_has_calls(calls)
 
+    def test_create_and_schedule_runs_specific_time(self):
```
qui: I think this whole test case class belongs in job_scheduler_test.py? If you feel like moving it.

PR #680, opened 2019-07-09
Sort action_runs by start_time in tronview

solarkennedy:


tron/commands/display.py
```diff
@@ -384,9 +385,6 @@ def store_data(self, data):
         self.data = data['runs']
         self.job_run = data
 
-    def rows(self):
```
qui: why remove this?

solarkennedy: by removing that rows method, the actionruns object can use the rows method of the super class, which has sorting support.

PR #682, opened 2019-07-12
Fix stderr output on check_tron_jobs

solarkennedy:


tron/bin/check_tron_jobs.py
```diff
@@ -126,10 +126,11 @@ def compute_check_result_for_job_runs(client, job, job_content):
     elif last_state == State.STUCK:
         prefix = "WARN: Job exceeded expected runtime or still running when next job is scheduled"
         status = 1
-        stderr = relevant_job_run['stderr']
+        stderr = relevant_job_run.get('stderr', "(No stderr available)")
```
qui: I think you need to get this from the `relevant_action`.

PR #685, opened 2019-07-16
Build the debian package with systemd

solarkennedy:
I have no idea what I'm doing, but stack overflow says this is the thing to do.
Needs more testing.

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -4,7 +4,7 @@
 PIP_INDEX_URL ?= https://pypi.python.org/simple
 
 %:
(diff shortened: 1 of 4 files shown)


solarkennedy:
Gosh, I can't remember why I opened this...

qui:
It was an attempt to resolve the message about needing to run "systemctl daemon-reload" when you upgrade the package.

solarkennedy:
Hmm.

solarkennedy:
discarding for now

PR #689, opened 2019-07-19
Provide better user errors when a RequestError happens

solarkennedy:
Users don't need to see a stack trace

### DIFF
--- a/bin/tronctl
+++ b/bin/tronctl
@@ -15,6 +15,7 @@ import argcomplete
 from tron import __version_info__
 from tron.commands import client
 from tron.commands import cmd_utils
+from tron.commands.client import RequestError
 from tron.commands.cmd_utils import ExitCode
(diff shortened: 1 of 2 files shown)


bin/tronctl
```diff
@@ -227,9 +228,14 @@ def main():
...
     cmd = COMMANDS[args.command]
-    for ret in cmd(args):
-        if not ret:
-            sys.exit(ExitCode.fail)
+    try:
+        for ret in cmd(args):
+            if not ret:
+                sys.exit(ExitCode.fail)
+    except RequestError as err:
+        print(f"Error connecting to tron: {err}", file=sys.stderr)
```
qui: What if you just print the server it's trying to connect to, from the args?

solarkennedy: Can do

solarkennedy:
@qui can you look at this again? In the end `args.server` is always set to something.

PR #690, opened 2019-07-19
Released 1.0.0

solarkennedy:
This is kind of a symbolic thing.
I think we've "made it" with Tron, now that we have gotten to the point where we can make a release and not have to babysit it as it auto-upgrades.

Here's to looking forward to 2.0 with k8s support!?


ssk2:
:shipit: fantastic work!

PR #695, opened 2019-08-20
WIP Make recover_batch tolerate empty files

solarkennedy:


qui:
Does this mean the recovery will exit 1 immediately if the file doesn't exist? Doesn't it already do that?

If we want it to be backwards compatible, we should wait for another notification if the file exists but it is empty.

solarkennedy:
I'll do this in multiple parts to reduce the chance of breakage.

PR #699, opened 2019-08-23
Added better ssh connection messages

solarkennedy:


solarkennedy:
Example output:
```
INFO Started SSH connection from 127.0.0.1:40736 to 127.0.0.1:22
...
INFO Stopped SSH connection from 127.0.0.1:40736 to 127.0.0.1:22
```

PR #700, opened 2019-08-29
Make batch recovery error messages easier to understand

solarkennedy:


qui:
Why don't we keep waiting for updates on the file if it is empty, instead of failing? The file should eventually be written at the end, even with the bug where we weren't closing the file. I think that was how it was working before we started looking at PIDs (https://github.com/Yelp/Tron/pull/666/ and https://github.com/Yelp/Tron/pull/683).

solarkennedy:
@qui @kawaiwanyelp ok, I think I've:

* DRY'd it out
* Made the logging more honest
* Handled the situation where we don't have a PID and need to just wait (hopefully that won't introduce "hung" jobs?)

tron/bin/recover_batch.py
```diff
@@ -53,48 +54,41 @@ def notify(notify_queue, ignored, filepath, mask):
...
-                'Action run killed by signal '
-                f'{signal.Signals(exit_code).name}'
-            )
+            error_msg = f'Action run killed by signal {signal.Signals(exit_code).name}'
         else:
             exit_code = return_code
-
-    elif pid is None or not psutil.pid_exists(pid):
+    elif pid is None:
+        log.warning(f"Status file {filepath.path} didn't have a PID. Will watch the file for updates.")
```
kawaiwanyelp: I am a little afraid this will hang, but since @qui has suggested otherwise, I am willing to try it out.

tron/bin/recover_batch.py
```diff
@@ -53,48 +54,41 @@ def notify(notify_queue, ignored, filepath, mask):
...
-        )
+        error_msg = f'Action runner pid {pid} no longer running. Assuming an exit of 1.'
 
     if exit_code is not None:
         reactor.stop()
         notify_queue.put((exit_code, error_msg))
 
 
+def file_is_empty(filepath):
+    return os.stat("file").st_size == 0
```
kawaiwanyelp: s/"file"/filepath/g

tron/bin/recover_batch.py
solarkennedy: As discussed, we kind want it to hang by design

tron/bin/recover_batch.py
solarkennedy: omg

kawaiwanyelp:
Looks like the failing Travis check is due to the same timeouts plaguing PaaSTA :P

PR #701, opened 2019-09-09
Always reset the restart delay when manually retrying an action

solarkennedy:
Due to the way the "returns" are setup here, the return of a delayed action is `None`, which gives a weird message to the user. (cc @smoy )

### DIFF
--- a/tron/core/actionrun.py
+++ b/tron/core/actionrun.py
@@ -421,8 +421,14 @@ class ActionRun(Observable):
         """Invoked externally (via API) when action needs to be re-tried
         manually.
         """
+
+        # Manually retrying means we force the retries to be 1 and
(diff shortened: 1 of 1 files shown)


qui:
We end up calling `restart` later, which creates a new delay: https://github.com/Yelp/Tron/blob/b6cde2e6c5e627518beb0c2478ca7e8c49ee3e90/tron/core/actionrun.py#L442

The lack of exit code is probably because that conditional returns None. I think you need to not create that delay if it's a manual retry.

keymone:
`cancel_delay` is only for `stop` or `kill` as it finalizes the run, i did bad job naming it. you need to manually cancel the promise like it's done in `start`: https://github.com/Yelp/Tron/blob/master/tron/core/actionrun.py#L371

PR #702, opened 2019-09-10
Added new namespace and manual config context variables

solarkennedy:
I'm trying to get a bit warmed up to this codebase in preparation to make it possible to embed graphs into tronview/tronweb for users.

### DIFF
--- a/docs/command_context.rst
+++ b/docs/command_context.rst
@@ -5,7 +5,22 @@ Built-In Command Context Variables
 ==================================
 
 Tron includes some built in command context variables that can be used in
-command configuration.
+command configuration for actions.
+
+These variables can be used in the command of an action, using Python's format syntax (``{}``).
+
+For example::
+
+    # myservice.yaml
+    myjob:
+      node: localhost
+      actions:
+        myaction1:
+          command: "Hello world! I'm {action} for job {name} running on {node}"
+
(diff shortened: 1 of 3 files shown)


docs/command_context.rst
```diff
@@ -72,3 +68,10 @@ Context variables only available to Jobs
     The last successful run date (defaults to current date if there was no
     previous successful run). Supports date arithmetic using the form
     ``{last_success#shortdate-1}``.
+
+**manual**
+    ``true`` if the job was run manually. ``false`` otherwise.
```
kawaiwanyelp: I was a little confused by this explanation. I suggest giving an example of "manual" usage, in this case `tronctl start`, and an example of how it could be used to enable feature like verbose logging.

docs/command_context.rst
```diff
@@ -72,3 +68,10 @@ Context variables only available to Jobs
...
+
+**namespace**
+    The namespace of the config where the job comes from. Often ``MASTER`` or ``servicename``.
```
kawaiwanyelp: Please give an example here, so if a job "bar" is defined in for the foo service in foo/tron-norcal-prod.yaml, then the job would appear in Tron as "foo.bar", where `foo` is the namespace.

PR #704, opened 2019-09-17
Don't do a retry_delay when manually retrying a failed command.

solarkennedy:
I don't know if this is the best solution, but it does seem to do the trick. (tested with `make dev`)

### DIFF
--- a/dev/config/MASTER.yaml
+++ b/dev/config/MASTER.yaml
@@ -20,42 +20,54 @@ mesos_options:
   dockercfg_location: file:///root/.dockercfg
 
 jobs:
-    - name: testjob0
-      node: localhost
-      schedule: "cron * * * * *"
-      actions:
-        - name: zeroth
-          command: "sleep 5"
-          trigger_downstreams:
-            minutely: "{ymdhm}"
-          cpus: 1
-          mem: 100
+  testjob0:
+    enabled: false
+    node: localhost
+    schedule: "cron * * * * *"
+    actions:
+      zeroth:
+        command: "sleep 5"
+        trigger_downstreams:
+          minutely: "{ymdhm}"
+        cpus: 1
+        mem: 100
 
-    - name: "testjob1"
-      node: localhost
-      schedule: "cron * * * * *"
-      actions:
-        - name: "first"
-          command: "sleep 5"
-          cpus: 1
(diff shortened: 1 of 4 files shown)


dev/config/MASTER.yaml
```diff
@@ -20,42 +20,54 @@ mesos_options:
   dockercfg_location: file:///root/.dockercfg
 
 jobs:
-    - name: testjob0
```
keymone: instead of tracking changes to this file, i think we should rename it to MASTER.yaml.example and gitignore MASTER.yaml

tron/core/actionrun.py
```diff
@@ -429,6 +429,7 @@ def retry(self):
         if self.in_delay is not None:
             self.in_delay.cancel()
             self.in_delay = None
+        self.retries_delay = None
```
keymone: i think this is going to affect subsequent automatic retries. do we care? if yes - we should instead have something like `self.skip_next_retries_delay = True` and relevant logic.

qui: Hmmm I think this solution is good enough (even though it did irritate my idealism a little bit at first :) )

If users are manually retrying an action, then there probably aren't any automatic retries remaining, so the retry delay doesn't really matter anymore.

example-cluster/tronfig/MASTER.yaml
```diff
@@ -27,7 +27,9 @@ jobs:
     schedule: "cron */5 * * * *"
     actions:
       one:
-        command: sleep 10 && date
+        command: exit 1
+        retries: 3
+        reties_delay: 1m
```
qui: typo - this is causing Travis to fail

tron/core/actionrun.py
solarkennedy: Yea I didn't think we cared at this point if a human is running the `retry` command.
y/fixit

dev/config/MASTER.yaml
solarkennedy: I can do that!

solarkennedy: Actually.. idk, I think we should just not care. I'll remove my comments, but I would like to make `make dev` always work out of the box with a sane config

PR #705, opened 2019-09-18
Speed up check tron jobs

stug:
3 changes to speed things up:
1. Don't called `client.index()` over and over again
2. Don't make a 2nd call to `client.job()` to get the specific job run -- we already have that info from what I can tell (but someone should check me on that)
3. Only get `client.action_runs()` if the job failed (this the is the big one)

On my dev box this takes running `check_tron_jobs` from ~4.5min to ~45 seconds.

See TRON-1385 for details

### DIFF
--- a/tron/bin/check_tron_jobs.py
+++ b/tron/bin/check_tron_jobs.py
@@ -107,7 +107,11 @@ def compute_check_result_for_job_runs(client, job, job_content, url_index):
         url_index,
         relevant_action['id'],
     )
-    action_run_details = client.action_runs(action_run_id.url, num_lines=10)
+
+    if last_state in (State.STUCK, State.FAILED, State.UNKNOWN):
+        action_run_details = client.action_runs(action_run_id.url, num_lines=10)
+    else:
+        action_run_details = {}
 
     if last_state == State.SUCCEEDED:
         prefix = f"OK: The last job ({job_run_id}) run succeeded on {cluster}. Will watch future or in progress runs for the next failure"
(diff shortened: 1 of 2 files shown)


tron/bin/check_tron_jobs.py
```diff
@@ -96,19 +95,23 @@ def compute_check_result_for_job_runs(client, job, job_content):
...
         last_state=last_state,
         actions_expected_runtime=actions_expected_runtime
     )
     action_run_id = get_object_type_from_identifier(
         url_index,
         relevant_action['id'],
     )
-    action_run_details = client.action_runs(action_run_id.url, num_lines=10)
+
+    if last_state in (State.STUCK, State.FAILED):
```
stug: is this the right set of states in which to fetch the action runs?

qui: You should also include UNKNOWN.

tron/bin/check_tron_jobs.py
```diff
@@ -96,19 +95,23 @@ def compute_check_result_for_job_runs(client, job, job_content):
...
+        action_run_details = client.action_runs(action_run_id.url, num_lines=10)
+    else:
+        action_run_details = None
```
qui: Maybe this should be an empty dict just for consistency's sake? We do `action_run_details.get` below - only on failed/stuck actions now (where you are fetching action_run_details anyway), but thinking it might be easy to mess up in the future.

stug: Yep, that's smart

tron/bin/check_tron_jobs.py
stug: Done

PR #706, opened 2019-09-19
fix job run ids reported by check_tron_jobs

stug:
The previous method was returning a `TronObjectIdentifier` namedtuple, resulting in alerts referencing jobs called things like `TronObjectIdentifier(type='JOB_RUN', url='/api/jobs/jolt.gc_zookeeper/32780')`.

For comparison:
```
(Pdb) relevant_job_run['id']
'ahab.update_docker_repositories.40088'
```


stug:
> How did this work before?

I think it's new since 1.0.1.  In that version `job_run_id` was never referenced in the output.

PR #707, opened 2019-09-20
Fix YAML doc loading

qui:
While testing recovery, I found that recovery was hanging forever. There are a few bugs, which I'll do in separate PRs.

First, both read_last_yaml_entries and get_key_from_last_line were only loading part of the status entry, e.g. `{'timestamp': 1569013680.5727654}`. From pyyaml 3.11 to 5.1, they changed `default_flow_style` from True to False. With True, each status entry is in one line, while with False, each key is in a separate line:
```
>>> yaml.safe_dump({'a': 1, 'b': 2}, explicit_start=True, default_flow_style=True)
'--- {a: 1, b: 2}\n'
>>> yaml.safe_dump({'a': 1, 'b': 2}, explicit_start=True, default_flow_style=False)
'---\na: 1\nb: 2\n'
```

Now I'm using yaml.load_all to parse it for us instead of assuming things about the lines. That should work with either flow style.

### DIFF
--- a/tests/bin/recover_batch_test.py
+++ b/tests/bin/recover_batch_test.py
@@ -1,3 +1,4 @@
+import tempfile
 from queue import Queue
 
 import mock
@@ -5,6 +6,14 @@ import pytest
 import yaml
 
 from tron.bin import recover_batch
+from tron.bin.action_runner import StatusFile
+
+
+@pytest.fixture
+def mock_file():
+    f = tempfile.NamedTemporaryFile()
+    yield f.name
+    f.close()
 
 
 @mock.patch('tron.bin.recover_batch.read_last_yaml_entries', autospec=True)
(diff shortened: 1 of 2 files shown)


solarkennedy:
Whoa! Jeeze. I did find it a little funny that we were looking a "line" in the thing instead of looking at it from a structured point of view.

tests/bin/recover_batch_test.py
```diff
@@ -35,3 +44,28 @@ def test_notify(mock_reactor, mock_read_last_yaml_entries, line, exit_code, erro
     assert actual_exit_code == exit_code
     assert actual_error_msg == error_msg
     assert mock_reactor.stop.call_count == 1
+
+
+def test_read_last_yaml_roundtrip(mock_file):
+    """Check that read_last_yaml_entries returns the same thing that the action
+    runner wrote."""
+    status = StatusFile(mock_file)
+    expected_content = {'return_code': 1, 'pid': 10345, 'command': 'foo'}
```
kawaiwanyelp: nit: can you provide an example where there are multiple lines, and have the test expect only the last?

qui: This test actually does write multiple lines, updated the test to check both lines so it's more clear.

PR #708, opened 2019-09-23
Put back check for action runners that have completed in recovery

qui:
This is the second (and last) bug I found in recovery so far. We mistakenly removed the part of `run` that checks if the status file already has an exit code in https://github.com/Yelp/Tron/pull/700, while trying to DRY it out. If the action_runner completes while Tron is down, the recovery action will hang forever waiting for the status file to be updated again.

I pulled most of the logic into `get_exit_code` to reduce repetition and added test coverage. Based off of https://github.com/Yelp/Tron/pull/707.

### DIFF
--- a/tron/bin/recover_batch.py
+++ b/tron/bin/recover_batch.py
@@ -1,7 +1,6 @@
 #!/usr/bin/env python3.6
 import argparse
 import logging
-import os
 import signal
 import sys
 from queue import Queue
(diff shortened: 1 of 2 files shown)


tron/bin/recover_batch.py
```diff
@@ -46,57 +45,53 @@ def read_last_yaml_entries(filename):
...
-        content = read_last_yaml_entries(filepath)
-        return content.get(key)
-    except Exception as e:
-        log.warning(f"Exception encountered when inspecting {filepath}: {e}")
-        return None
+    return exit_code, error_message
 
 
 def run(fpath):
+    # Check if the process has already completed.
```
kawaiwanyelp: nit: Since we accidentally removed it last time, I would add a note here saying that this prevents hanging.

PR #711, opened 2019-09-27
Move SSH action recovery logic into the class to enable retries

qui:
I tested this with `make dev` locally.

I (or someone else) can follow up with the actual retries later, wanted to keep the diff small to reduce the chance of mistakes 😬


qui:
^ That change was just filling out the TODO in test_recover_action_run_action_runner.

PR #715, opened 2019-10-04
Automatically try to recover on unknown ssh actions

qui:


tests/core/actionrun_test.py
```diff
@@ -738,24 +740,37 @@ def test_handler_unhandled(self):
...
         assert not self.action_run.recover()
 
-    def test_recover_action_run_incorrect_state(self):
-        # Should return falsy if not UNKNOWN.
-        self.action_run.action_runner = SubprocessActionRunnerFactory(
+
+class TestSSHActionRunRecover:
+    @pytest.fixture(autouse=True)
+    def setup_action_run(self):
+        self.output_path = filehandler.OutputPath(tempfile.mkdtemp())
```
Rob-Johnson: can you wrap this in another decorator so we know for sure that the tmp dir will get cleaned up?

qui: I pulled output_path out into a separate fixture - let me know if that's not what you were thinking.

qui: merging this by popular demand, but I can always update it later.

PR #717, opened 2019-10-09
Try to handle backfills better in monitoring precious runs

qui:
I think this handles the issues we've been seeing, where a backfill causes erroneous alerts.

I reversed the order of the runs in the test to match the Tron API - it returns newest runs first.

### DIFF
--- a/tests/bin/check_tron_jobs_test.py
+++ b/tests/bin/check_tron_jobs_test.py
@@ -1,14 +1,20 @@
 import time
 
 import mock
+import pytest
 from mock import patch
 from mock import PropertyMock
 
 from testifycompat import assert_equal
-from testifycompat import setup
 from testifycompat import TestCase
 from tron.bin import check_tron_jobs
 from tron.bin.check_tron_jobs import State
+
+
+@pytest.fixture(autouse=True)
+def mock_run_interval():
+    with patch.object(check_tron_jobs, '_run_interval', 300):
+        yield
 
 
 class TestCheckJobs(TestCase):
@@ -24,7 +30,6 @@ class TestCheckJobs(TestCase):
         mock_check_job_result,
     ):
         type(mock_args.return_value).job = PropertyMock(return_value=None)
-        type(mock_args.return_value).run_interval = 300
         mock_client.return_value.jobs.return_value = [
             {
                 'name': 'job1',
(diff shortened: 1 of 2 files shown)


qui:
Updated to use a simpler approach - just check the last 7 days. If a day has been alerting for more than 7 days already, the alert will stay open, it just won't get auto-resolved by this script.

tron/bin/check_tron_jobs.py
```diff
@@ -337,6 +338,10 @@ def sort_runs_by_interval(job_content, interval='day', until=None):
                 time.mktime(_timestamp_to_timeobj(run['run_time']))
                 for run in job_content['runs']
             ])
+            earliest_run_time = max(
+                earliest_run_time,
```
solarkennedy: This logic is about 12 lines long. What do you think a function called 'calculate_earliest_run_time' that took `job_content` and `until` ?

tests/bin/check_tron_jobs_test.py
```diff
@@ -1554,22 +1559,44 @@ def test_sort_runs_by_interval_day(self):
...
         run_buckets = check_tron_jobs.sort_runs_by_interval(self.job, 'day')
 
         assert '2018.10.14' in run_buckets
         assert run_buckets['2018.10.14'] == []
 
+    @patch('time.time', mock.Mock(return_value=1539633600.0), autospec=None)
+    def test_sort_runs_by_interval_day_old_empty_buckets(self):
+        # If the newest run is a backfill for an older date, shouldn't be
+        # included for buckets
+        self.job['runs'] = [{
```
solarkennedy: This doesn't pollute other tests because fixtures are recreated... right?

tests/bin/check_tron_jobs_test.py
```diff
@@ -1554,22 +1559,44 @@ def test_sort_runs_by_interval_day(self):
...
+            'run_num': 16,
+            'run_time': '2018-10-01 12:00:00',
+            'start_time': '2018-10-01 12:00:00',
+            'end_time': '2018-10-01 12:30:00',
+            'state': 'succeeded',
+            'exit_status': 0,
+        }] + self.job['runs']
+
+        run_buckets = check_tron_jobs.sort_runs_by_interval(self.job, 'day')
+        assert '2018.10.02' not in run_buckets
```
solarkennedy: Can you add a comment to say why 10.02 is special here? Because it is the day after right? so therefore we are proving that 10.01 was not included in the earliest_until, right? (essentially ignored in the earliest_run_time calculation

tests/bin/check_tron_jobs_test.py
qui: Yeah, the default scope for fixtures is per-function.

tests/bin/check_tron_jobs_test.py
qui: Yeah, that's right. I just updated it to 10.01 to make it more obvious, and added a comment too.

PR #718, opened 2019-10-11
tron writes environment variables for the action_runner

drmorr0:
we export four variables, parsed from the run_id

TRON_JOB_NAMESPACE
TRON_JOB_NAME
TRON_RUN_NUM
TRON_ACTION

### DIFF
--- a/tron/bin/action_runner.py
+++ b/tron/bin/action_runner.py
@@ -73,6 +73,24 @@ def validate_output_dir(path):
             raise OSError("Could not create output dir %s" % path)
 
 
+def build_environment(run_id):
+    try:
+        namespace, job, run_num, action = run_id.split('.', maxsplit=3)
+    except ValueError:
+        # if we can't parse the run_id, we don't want to abort, so just
+        # set these semi-arbitrarily
+        namespace, job, run_num, action = ['UNKNOWN'] * 4
+
+    new_env = dict(os.environ)
+    new_env['TRON_JOB_NAMESPACE'] = namespace
+    new_env['TRON_JOB_NAME'] = job
+    new_env['TRON_RUN_NUM'] = run_num
+    new_env['TRON_ACTION'] = action
+
+    logging.debug(new_env)
+    return new_env
+
+
 def run_proc(output_path, command, run_id, proc):
     logging.warning(f'{run_id} running as pid {proc.pid}')
     status_file = StatusFile(os.path.join(output_path, STATUS_FILE))
@@ -103,12 +121,13 @@ def parse_args():
     return parser.parse_args()
 
 
-def run_command(command):
+def run_command(command, run_id):
     return subprocess.Popen(
         command,
         shell=True,
         stdout=subprocess.PIPE,
         stderr=subprocess.PIPE,
+        env=build_environment(run_id=run_id),
     )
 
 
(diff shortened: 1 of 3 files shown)


tron/bin/action_runner.py
```diff
@@ -103,12 +103,19 @@ def parse_args():
...
 
-def run_command(command):
+def run_command(command, run_id):
+    namespace, job, run_num, action = run_id.split('.')
     return subprocess.Popen(
         command,
         shell=True,
         stdout=subprocess.PIPE,
         stderr=subprocess.PIPE,
+        env=dict(
```
solarkennedy: I'm afraid that this is a breaking change compared to the previous which inherited the original env, including PATH, etc.

Can you make that not the case?

tron/bin/action_runner.py
```diff
@@ -103,12 +103,19 @@ def parse_args():
     return parser.parse_args()
 
 
-def run_command(command):
+def run_command(command, run_id):
+    namespace, job, run_num, action = run_id.split('.')
```
solarkennedy: Can you turn this into a function that returns the env dictionary?

That will be easier to test and change in the future.

solarkennedy: This is too scary also. You are assuming there will only ever be 4 periods. Please make the function more defensive against this.

Also it needs to be in a big try/except. I don't want this split to cause a job not to run.

tron/bin/action_runner.py
drmorr0: Done.

drmorr0:
```
drmorr@dev54-uswest1adevc ~tron > tronview --server=http://dev54-uswest1adevc.dev.yelpcorp.com:8089 MASTER.testjob0.1.zeroth
Action Run          : MASTER.testjob0.1.zeroth
State               : succeeded
Node                : drmorr@localhost
Command             : env
Bare command        : env
Start time          : 2019-10-11 15:44:00
End time            : 2019-10-11 15:44:00
Final exit status   : 0
Retry exit statuses : []
Waits for triggers  :
Publishes triggers  : minutely.2019-10-11T15:44
Requirements:

Stdout:
XDG_RUNTIME_DIR=/run/user/4048
LANG=en_US.UTF-8
TRON_ACTION=zeroth
TRON_JOB_NAME=testjob0
SHELL=/bin/zsh
GIT_COMMITTER_EMAIL=drmorr@yelp.com
PWD=/nail/home/drmorr
SSH_CONNECTION=127.0.0.1 39642 127.0.0.1 22
EDITOR=vim
TRON_JOB_NAMESPACE=MASTER

Stderr:

```

tron/bin/action_runner.py
drmorr0: Done.

tron/bin/action_runner.py
```diff
@@ -73,6 +73,24 @@ def validate_output_dir(path):
             raise OSError("Could not create output dir %s" % path)
 
 
+def build_environment(run_id):
+    try:
+        namespace, job, run_num, action = run_id.split('.', maxsplit=3)
+    except ValueError:
+        # if we can't parse the run_id, we don't want to abort, so just
+        # set these semi-arbitrarily
+        namespace, job, run_num, action = '', '', '', run_id
```
solarkennedy: `UNKOWN` would be ok I think. That would make it more clear that there was a bug in parsing.

Setting the action to the `run_id` sound bug-prone.

tron/bin/action_runner.py
```diff
@@ -143,7 +162,7 @@ def main():
     args = parse_args()
     validate_output_dir(args.output_dir)
     configure_logging(run_id=args.run_id, output_dir=args.output_dir)
-    proc = run_command(args.command)
+    proc = run_command(args.command, args.run_id)
```
solarkennedy: kwargs please here

PR #719, opened 2019-10-22
Address GH alerts

keymone:
bump twisted, werkzeug and ecdsa


keymone:
weekly email, it's sent to people with admin permissions on repositories

PR #720, opened 2019-10-24
Extend tron environment variables to mesos-based actions too. Also Docs.

solarkennedy:


tron/core/actionrun.py
```diff
@@ -884,7 +885,7 @@ def _create_mesos_task(self, mesos_cluster, serializer, task_id=None):
                          for c in self.constraints],
             docker_image=self.docker_image,
             docker_parameters=[e._asdict() for e in self.docker_parameters],
-            env=self.env,
+            env=compose_environment_variables(input_env=self.env, run_id=self.id),
```
drmorr0: Why do you have this extra function here?  isn't it sufficient to do something like
```
env=build_environment(run_id=self.id, original_env=self.env)
```

All `compute_environment_variables` appears to be doing is creating a copy of a dict that's already a copy of the original env, it doesn't really seem necessary?

tron/bin/action_runner.py
```diff
@@ -73,15 +73,18 @@ def validate_output_dir(path):
...
+        original_env = dict(os.environ)
     try:
         namespace, job, run_num, action = run_id.split('.', maxsplit=3)
     except ValueError:
         # if we can't parse the run_id, we don't want to abort, so just
         # set these semi-arbitrarily
         namespace, job, run_num, action = ['UNKNOWN'] * 4
 
-    new_env = dict(os.environ)
+    new_env = dict()
```
drmorr0: Can't we just do `new_env = dict(original_env)`?

tron/core/actionrun.py
solarkennedy: We can. Originally I had more going on in that function, then I trimmed it down.

Maybe YAGNI, but I wanted room for other mesos-specific env variables, more than the ones provided by the action runner.

drmorr0: My personal preference is to add it later if we need it, but I don't much care either way.

PR #721, opened 2019-10-31
Added an option to hide stderr from checks

Qmando:
The stderr in the output of Tron checks can be obnoxiously big and usually does not actually contain output from the task itself, but rather Mesos. This adds an setting to turn it off.


qui:
@Qmando thanks for taking the initiative here! FYI it won't take effect until the next release - we should do one soon for a task_processing bump, so I'd estimate <1 week.

We could probably also work on making the messages from Mesos less verbose / put them somewhere else, so when people look at the stderr it's less noisy.

PR #722, opened 2019-11-01
documented new tronctl commands and cross job dependency

EmanekaT:


docs/jobs.rst
```diff
@@ -450,3 +458,26 @@ Job Run States
...
+    details.
+
+.. _shortdate:
+**What does shortdate in triggers mean?**
+    There are two concepts of shortdate here.
+
+    shortdate in triggered_by: this shortdate is technically the run_date,
+    indicating when the tron job runs
+    shortdate in command: this shortdate is used by batch jobs to specify which s3
+    dir it is writing to or polling.
```
solarkennedy: I think these are better placed in y/tron-faq than here. I think they will be more discoverable and easier for others to edit. They are also very yelpy

docs/jobs.rst
```diff
@@ -219,16 +219,24 @@ Optional Fields
     have been emitted by upstream actions. Unlike with ``requires`` attribute,
     dependent actions don't have to belong to the same job. ``triggered_by``
     template may contain any pattern allowed in ``command`` attribute.
+    See :ref:`shortdate` for an explantion of shortdate
```
solarkennedy: Can you put the explanation right below the `shortdate` example here?

docs/jobs.rst
EmanekaT: They already exists on yelp wiki the cross-job dependency page. I move them here mostly because I don't think everyone knows all the different y/links or wiki pages. At least I expect everything to be on Tron doc.

docs/jobs.rst
EmanekaT: I'm a bit hesitant because shortdate is not only used by triggered_by but also other fields. I can move the link below the example though.

PR #727, opened 2019-12-12
refactor in preparation for adding the SparkActionRun class

drmorr0:
As we consider adding more and more executor types to Tron, I don't want to end up in this situation where all of them get dumped in one giant file, so I'm just going to split them out right now.  The only annoying thing here is the circular import, which means we have to import the specific ActionRun classes inside the Factory instead of at the top of the file.

### DIFF
--- a/tron/config/schema.py
+++ b/tron/config/schema.py
@@ -226,7 +226,7 @@ StatePersistenceTypes = Enum(
     'StatePersistenceTypes', dict(shelve='shelve', yaml='yaml', dynamodb='dynamodb')
 )
 
-ExecutorTypes = Enum('ExecutorTypes', dict(ssh='ssh', mesos='mesos'))
+ExecutorTypes = Enum('ExecutorTypes', dict(ssh='ssh', mesos='mesos', spark='spark'))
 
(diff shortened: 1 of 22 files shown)


tron/config/schema.py
```diff
@@ -226,7 +226,7 @@ def config_object_factory(name, required=None, optional=None):
     'StatePersistenceTypes', dict(shelve='shelve', yaml='yaml', dynamodb='dynamodb')
 )
 
-ExecutorTypes = Enum('ExecutorTypes', dict(ssh='ssh', mesos='mesos'))
+ExecutorTypes = Enum('ExecutorTypes', dict(ssh='ssh', mesos='mesos', spark='spark'))
```
solarkennedy: My personal opinion is that tron shouldn't know about a special spark thing. I was hoping it could launch it directly on mesos, so it could just "be" a mesos executor.

solarkennedy:
I was hoping we could remove the ssh executor sometime in 2020 after all spark jobs are launched on mesos. In practice we will be adding a k8s executor too, so I guess we need room for that.

drmorr0:
closing because i didn't understand what I was doing

PR #728, opened 2019-12-19
Try to improve log readability a bit in mesos actions

solarkennedy:
Looking at a lot of tronview, it is hard to tell when a retry stops and starts.
I'm hoping that these changes will make it easier for users to see individual retries and stuff.

### DIFF
--- a/tron/mesos.py
+++ b/tron/mesos.py
@@ -196,7 +196,7 @@ class MesosTask(ActionCommand):
             )
         elif mesos_type == 'running':
             agent = event.raw.get('agent_id', {}).get('value')
-            self.log.info(f'Running on agent {agent}')
+            self.log.info(f'Remote job started running on agent {agent}:')
         elif mesos_type == 'finished':
             pass
         elif mesos_type in self.ERROR_STATES:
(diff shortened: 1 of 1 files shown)


qui:
I know we're deprecating SSH actions soon, but for principle's sake, all actions would benefit from spaces between logs on retries, so I think the newline/### part of this would be better placed around here: https://github.com/Yelp/Tron/blob/master/tron/core/actionrun.py#L426

Then you'd capture manual retries too.

The updates to the logging text here seems helpful either way.

solarkennedy:
@qui I considered that, but the `log` methods there go to the std python logger, and not the logger that goes to the action stdout. The only logger I know of that you can log to and have it show up in tronview is the one we "extract" in the mesos.py thing.

Would you like me to move that `self.log` functionality into the base actionrun class and work off that?

qui:
Hmm... The stdout/stderr belongs to the MesosTask. Luckily MesosTask inherits from ActionCommand, and ssh actions also use ActionCommands, so I think you can add some logging here so there's space after the end of a command: https://github.com/Yelp/Tron/blob/d53afcbe68b8e487675b95f7a69d91e3b586a691/tron/actioncommand.py#L82

PR #729, opened 2020-01-07
Enforce job graph dependency

jangsutsr:
This makes sure user cannot change the name of a job/action which is
triggered/required by other actions and have things silently fail.

### DIFF
--- a/tron/core/jobgraph.py
+++ b/tron/core/jobgraph.py
@@ -13,7 +13,7 @@ class JobGraph(object):
     cross-job dependencies (aka triggers)
     """
 
-    def __init__(self, config_container):
+    def __init__(self, config_container, should_validate_missing_dependency=False):
         """ Build an adjacency list and a reverse adjacency list for the graph,
         and store all the actions as well as which actions belong to which job
         """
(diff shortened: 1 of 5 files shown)


tron/core/jobgraph.py
```diff
@@ -44,6 +52,31 @@ def __init__(self, config_container):
...
+                    )
+            for action in dependent_actions:
+                if action not in all_actions:
+                    error_messages.append(
+                        'Action {0} is dependency of actions: {1}.'.format(
+                            action,
+                            ', '.join(dependent_actions[action]),
+                        )
+                    )
+            if error_messages:
```
drmorr0: Discussed on slack a slightly more clear format:

```The following actions are dependencies of other actions:
job1.action is dependeny of:
  - action 1
  - action 4
job2.action is dependeny of:
  - action 3```

File: tron/core/jobgraph.py [hunk 2]
```diff
@@ -22,20 +22,28 @@ def __init__(self, config_container):
         self._adj_list = defaultdict(list)
         self._rev_adj_list = defaultdict(list)
 
+        dependent_jobs = defaultdict(list)
```
drmorr0: I think we don't need to track _both_ dependent jobs and dependent actions, since we can compute/track the jobs in the action names.

PR #730, opened 2020-01-08
show the relative datetime for runs correctly

Rob-Johnson:
this fixes the longstanding issue where the references to runtimes are
incorrect for users outside of PST.

credit to @danshearmur who actually did this work


solarkennedy:
Ship it!

tronweb/coffee/views.coffee
```diff
@@ -21,9 +21,10 @@ window.dateFromNow = (string, defaultString='never') ->
         """
 
     if string
-        formatted = moment(string).format('MMM, Do YYYY, h:mm:ss a')
+        date = moment.tz string, 'America/Los_Angeles'
```
solarkennedy: If we are going to hard code this, is it easy to say 

    `(All times in A/L_A time)`

in the html view?

PR #731, opened 2020-01-14
Tron should never fail when trying to recover

drmorr0:
Tron actually crashes twice with long filenames, once when it shuts down and once when it restarts.  This change handles both of these cases.  It doesn't resolve the broader problem with long filenames, but it does ensure that Tron will shut down and start up cleanly.

**Testing Done**
* `make test` passes, added some new tests
* manual testing with long filenamews; I reproduced the error before this change and confirms that this change fixes it (specifically I confirmed that even if one job fails to recover, other jobs can still be recovered).

### DIFF
--- a/tron/core/actionrun.py
+++ b/tron/core/actionrun.py
@@ -776,7 +776,11 @@ class SSHActionRun(ActionRun, Observer):
         desired_delay = INITIAL_RECOVER_DELAY * (3 ** self.recover_tries)
         self.recover_tries += 1
         log.info(f'Starting try #{self.recover_tries} to recover {self.id}, waiting {desired_delay}')
-        return self.do_recover(delay=desired_delay)
+        try:
+            return self.do_recover(delay=desired_delay)
+        except Exception as e:
+            log.error(f'Could not recover {self.id}: {str(e)}')
+            return None
 
     def recover(self):
         log.info(f"Creating recovery run for actionrun {self.id}")
@@ -795,7 +799,11 @@ class SSHActionRun(ActionRun, Observer):
             )
             return None
 
-        return self.do_recover(delay=0)
+        try:
+            return self.do_recover(delay=0)
+        except Exception as e:
+            log.error(f'Could not recover {self.id}: {str(e)}')
+            return None
 
     def do_recover(self, delay):
         recovery_command = f"{self.action_runner.exec_path}/recover_batch.py {self.action_runner.status_path}/{self.id}/status"
--- a/tests/core/actionrun_test.py
+++ b/tests/core/actionrun_test.py
@@ -782,6 +782,11 @@ class TestSSHActionRunRecover:
         self.action_run.machine.state = ActionRun.FAILED
         assert not self.action_run.recover()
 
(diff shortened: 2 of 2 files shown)


tron/core/actionrun.py
```diff
@@ -776,7 +776,11 @@ def handle_unknown(self):
         desired_delay = INITIAL_RECOVER_DELAY * (3 ** self.recover_tries)
         self.recover_tries += 1
         log.info(f'Starting try #{self.recover_tries} to recover {self.id}, waiting {desired_delay}')
-        return self.do_recover(delay=desired_delay)
+        try:
+            return self.do_recover(delay=desired_delay)
+        except Exception as e:
+            log.error(f'Could not recover {self.id}: {str(e)}')
```
vkhromov: Here and everywhere else: `str(e)` strips a lot of information from the exception, including the exception type and the stack trace.
Could you please fix that.

Rob-Johnson:
> but it does ensure that Tron will shut down and start up cleanly.
I don't agree with this - the error still occurs, we're just ignoring it. is ignoring this error actually the right thing to do? without some effect coming from this, it's just going to mean that Tron behaves in unexpected ways when it encounters a failure, and worse, it'll do so silently.

Though it's kind of painful for these things to happen, isn't it better to fail 'hard' than for the failure to manifest in some other way down the line and us to have to spend the time parsing logs to figure out what the cause was?

drmorr0:
> > but it does ensure that Tron will shut down and start up cleanly.
> 
> I don't agree with this - the error still occurs, we're just ignoring it. is ignoring this error actually the right thing to do? without some effect coming from this, it's just going to mean that Tron behaves in unexpected ways when it encounters a failure, and worse, it'll do so silently.
> 
> Though it's kind of painful for these things to happen, isn't it better to fail 'hard' than for the failure to manifest in some other way down the line and us to have to spend the time parsing logs to figure out what the cause was?

I think in this case the outcome of "one job can't recover but the rest of Tron can safely restart" is a better output than "All of Tron is down and no jobs can run" -- this potentially could have been a much-worse outage if it had happened at the end of the month when revenue-critical jobs needed to run.

What if I set up a Splunk alert that let us know when this was happening?

Rob-Johnson:
```
I think in this case the outcome of "one job can't recover but the rest of Tron can safely restart" is a better output than "All of Tron is down and no jobs can run" -- this potentially could have been a much-worse outage if it had happened at the end of the month when revenue-critical jobs needed to run.
```

hmm, yeah, it's a hard call to make. if it fails, what state does it leave the job in? if we encounter a failure that means that *all* jobs fail to recover, I don't want us limping along in some weird state where no jobs can run anyway because they're UNKNOWN without a recovery?

I'm just concerned that we don't really know what the effect of this might be, whereas we'll atleast be able to guarantee that no new jobs will run whilst Tron is down. I'd err on the side of not running them at all, rather than running in an undefined way. Also, this is super rare, and an indicator of a real (serious) problem, so I'm ok with this being treated as an exceptional problem if it does happen.

PR #732, opened 2020-01-15
read environment variables for configuring Clusterman

drmorr0:
Anything that wants to can now set the following environment variables:

`CLUSTERMAN_RESOURCES`: an _exact_ string to be passed into the clusterman metrics writer
`EXECUTOR_CLUSTER`: name of the paasta cluster we want to scale up
`EXECUTOR_POOL`: name of the paasta pool we want to scale up

(if the last two aren't set, it will fall back to `PAASTA_CLUSTER` and `PAASTA_POOL`, that is, the cluster where the Tron master is running)


tron/mesos.py
```diff
@@ -13,10 +16,28 @@
...
 
+def get_clusterman_metrics():
+    try:
+        import clusterman_metrics
+        import clusterman_metrics.util.costs
+
+        clusterman_yaml = CLUSTERMAN_YAML_FILE_PATH
+        staticconf.YamlConfiguration(
+            CLUSTERMAN_METRICS_YAML_FILE_PATH, namespace="clusterman_metrics"
+        )
```
stug: I am probably forgetting, but how are callers going to set this?  I am wondering if it would make sense to do something similar to what we do in spark-run where we look up the AWS region from the cluster (https://github.com/Yelp/paasta/blob/799cc1da422c6c7f53fe1d91e0eb9e0dc6d6ec3f/paasta_tools/cli/cmds/spark_run.py#L578)

stug: Err I am sorry, this comment is in the wrong place.

tron/mesos.py
```diff
@@ -362,6 +383,23 @@ def submit(self, task):
 
         mesos_task_id = task.get_mesos_id()
         self.tasks[mesos_task_id] = task
+        env = task.get_config()['environment']
+        clusterman_resource_str = env.get('CLUSTERMAN_RESOURCES')
+        clusterman_metrics, _ = get_clusterman_metrics()
+        if clusterman_resource_str and clusterman_metrics:
+            clusterman_resources = json.loads(clusterman_resource_str)
+            aws_region = env.get('EXECUTOR_AWS_REGION')
```
stug: I am probably forgetting the details of a previous review, but how are callers going to set this? I am wondering if it would make sense to do something similar to what we do in spark-run where we look up the AWS region from the cluster (https://github.com/Yelp/paasta/blob/799cc1da422c6c7f53fe1d91e0eb9e0dc6d6ec3f/paasta_tools/cli/cmds/spark_run.py#L578)

tron/mesos.py
```diff
@@ -362,6 +383,23 @@ def submit(self, task):
...
+            cluster = env.get('EXECUTOR_CLUSTER', env.get('PAASTA_CLUSTER'))
+            pool = env.get('EXECUTOR_POOL', env.get('PAASTA_POOL'))
+            metrics_client = clusterman_metrics.ClustermanMetricsBotoClient(
+                region_name=aws_region,
+                app_identifier=f'tron_{cluster}_{pool}',
```
stug: I _think_ you need to set `app_identifier` to just `pool`?  I think otherwise clusterman won't be able to find the metrics.

tron/mesos.py
```diff
@@ -362,6 +383,23 @@ def submit(self, task):
...
+            )
+            with metrics_client.get_writer(
+                clusterman_metrics.APP_METRICS, aggregate_meteorite_dims=True
+            ) as writer:
+                for metric_key, desired_quantity in clusterman_resources.items():
+                    writer.send((metric_key, int(time.time()), desired_quantity))
```
stug: ultra-mega-nit: the metric values can be anything -- doesn't have to be desired quantity

tron/mesos.py
drmorr0: Fixed

tron/mesos.py
drmorr0: oh you're right.  that feels like an odd design decision... what happens when two clusters have the same pool name?

tron/mesos.py
drmorr0: fixed

tron/mesos.py
stug: I filed CLUSTERMAN-327 about this long ago, but we never did anything about it.  I guess it hasn't hurt anything yet, but seems like it could cause confusing (and maybe even hard-to-notice) issues.

tron/mesos.py
```diff
@@ -362,6 +383,23 @@ def submit(self, task):
...
+        clusterman_resource_str = env.get('CLUSTERMAN_RESOURCES')
+        clusterman_metrics, _ = get_clusterman_metrics()
+        if clusterman_resource_str and clusterman_metrics:
+            clusterman_resources = json.loads(clusterman_resource_str)
+            cluster = env.get('EXECUTOR_CLUSTER', env.get('PAASTA_CLUSTER'))
+            pool = env.get('EXECUTOR_POOL', env.get('PAASTA_POOL'))
+            aws_region = staticconf.read(f'clusters.{cluster}.aws_region', namespace='clusterman_metrics')
+            metrics_client = clusterman_metrics.ClustermanMetricsBotoClient(
+                region_name=aws_region,
+                app_identifier=f'{pool}',
```
stug: does this really need an f-string?

PR #735, opened 2020-01-22
tron output dir names shortened

drmorr0:
This will prevent issues like DAR-755 from happening again.  In summary, instead of storing things in `/nail/tron/namespace.jobname/namespace.jobname.run_num/namespace.jobname.run_num.actionname/.stdout` (and, when we have to recover an "unknown" action, `/nail/tron/namespace.jobname/namespace.jobname.run_num/namespace.jobname.run_num.actionname/namespace.jobname.run_num.recovery-namespace.jobname.run_num.actionname/.stdout`), we instead store things in `/nail/tron/namespace/jobname/run_num/action_name/.stdout` or `/nail/tron/namespace/jobname/run_num/action_name-recovery/.stdout`.  Aside from keeping us from running into the 255-char pathname limit, this also will help organize things on our Tron masters a little more sanely.

**Testing Done**
* `make test` passes
* manual testing with `make dev`: I confirmed that when we "change over" from the old directory names to the new directory names, we can still recover jobs successfully, and I confirmed that we can still recover successfully once we're totally on the "new way"

### DIFF
--- a/tron/core/actionrun.py
+++ b/tron/core/actionrun.py
@@ -3,6 +3,7 @@
 """
 import datetime
 import logging
+import os
 from typing import List
 
 from twisted.internet import reactor
@@ -271,7 +272,7 @@ class ActionRun(Observable):
         self.extra_volumes = extra_volumes
         self.mesos_task_id = mesos_task_id
         self.output_path = output_path or filehandler.OutputPath()
-        self.output_path.append(self.id)
+        self.output_path.append(self.action_name)
         self.context = command_context.build_context(self, parent_context)
         self.retries_remaining = retries_remaining
         self.retries_delay = retries_delay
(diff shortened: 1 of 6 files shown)


tron/core/actionrun.py
```diff
@@ -800,14 +801,18 @@ def recover(self):
...
+        # Put the "recovery" output at the same directory level as the original action_run's output
+        recovery_output_path = self.output_path.clone()
+        recovery_output_path.base = os.path.dirname(recovery_output_path.base)
+
         # Might not need a separate action run
         # Using for the separate name
         recovery_run = SSHActionRun(
-            job_run_id=self.job_run_id,
-            name=f"recovery-{self.id}",
+            job_run_id=f"{self.job_run_id}-recovery",
```
Rob-Johnson: hm, I don't know whether there are any side effects to this? is it used anywhere to link the recovery action with the original job run

drmorr0: Not as far as I can tell.  I tested by:

1) starting a long-running job on `master`
2) killing tron
3) switching to this branch
4) restarting tron

and it recovered the job just fine.

qui: Why change the job_run_id from `self.job_run_id` to `self.job_run_id-recovery` anyway? It doesn't change the output path.

drmorr0: I don't know, I'm reverting this.

drmorr0:
OK, this new version saves output from before the change to after the change; as a nice side effect, it also ensures that "recovery" output still appears after the job completed, which was not true in the past.  It's a bit hacky, but I'm not too sure there's a better way to do it...

I confirmed that:
* stdout/stderr from master is displayed after restarting Tron on this branch
* stdout/stderr from any jobs that needed to be recovered during the switch appear after restarting Tron on this branch
* stdout/stderr from jobs that run on this branch appear
* stdout/stderr from jobs that are recovered while on this branch appear correctly.

qui:
Don't we also want to be able to recover stdout/stderr for Mesos tasks? If it makes it easier, I think the output from the recovery action is less important - the recovery action just polls the status file for updates. Otherwise this approach makes sense to me.

drmorr0:
This should be fixed so it shows stdout/err from Mesos too.

PR #737, opened 2020-02-06
Skip job graph validation in write_config

jangsutsr:
Tron does not natively support cross-cluster triggers right now. The temporary workaround is using `tronctl publish` to expose cross-cluster triggers. Such dependency is hard to capture and can break tron config writes at pre-write validation step if existing configuration already contains such dependency.
This PR is a short-term fix which turns off job graph validation for config writes. A longer-term solution is adding a flag to rule out certain actions from job graph validation. Eventually we will turn job graph validation back up after tron natively support cross-cluster triggers.

### DIFF
--- a/tests/config/manager_test.py
+++ b/tests/config/manager_test.py
@@ -151,6 +151,7 @@ class TestConfigManager(TestCase):
         self.manager.validate_with_fragment.assert_called_with(
             name,
             self.content,
+            should_validate_missing_dependency=False,
         )
 
     def test_write_config_new_name(self):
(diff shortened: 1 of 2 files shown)


qui:
Please add a traceback and/or description of the error that occurred.

qui:
It would be ideal if we could detect trigger dependency issues in pre-commit or paasta validate rather than (or in addition to) when Tron applies the config. 

Otherwise, if anyone forgets to add `validate_dependencies: ignore` to a new cross-cluster action, tronfig application will start failing again (and someone will get paged).

Either way, we should add a note about `validate_dependencies: ignore` to the wiki FAQ page about cross-cluster triggers.

qui:
On second thought, if we turn the dependency validation back on, we'll get paged every time someone breaks a "normal" trigger dependency, too. We should really prevent users from committing the bad config before it gets deployed.

PR #742, opened 2020-03-06
Added new page_for_expected_runtime option

solarkennedy:
This is just a first pass at this.

I would like to not change any other behavior in the monitoring script yet, just add the additional option.

### DIFF
--- a/tron/bin/check_tron_jobs.py
+++ b/tron/bin/check_tron_jobs.py
@@ -124,8 +124,13 @@ def compute_check_result_for_job_runs(client, job, job_content, url_index, hide_
         status = 0
         stderr = ""
     elif last_state == State.STUCK:
-        prefix = f"WARN: Job {job_run_id} exceeded expected runtime or still running when next job is scheduled on {cluster}"
-        status = 1
+        if job['monitoring'].get("page_for_expected_runtime", False):
+            level = "CRIT"
+            status = 2
+        else:
+            level = "WARN"
+            status = 1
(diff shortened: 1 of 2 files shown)


qui:
Do you need to do this, too? 

https://github.com/Yelp/Tron/blob/0ada41a2cd2d4051ae25b1b5d92f2de978bbc92b/tron/bin/check_tron_jobs.py#L396

solarkennedy:
> Do you need to discard the arg, like this?
> 
> https://github.com/Yelp/Tron/blob/0ada41a2cd2d4051ae25b1b5d92f2de978bbc92b/tron/bin/check_tron_jobs.py#L396-L398

Oh right, ty

tron/bin/check_tron_jobs.py
```diff
@@ -124,8 +124,13 @@ def compute_check_result_for_job_runs(client, job, job_content, url_index, hide_
         status = 0
         stderr = ""
     elif last_state == State.STUCK:
-        prefix = f"WARN: Job {job_run_id} exceeded expected runtime or still running when next job is scheduled on {cluster}"
-        status = 1
+        if job['monitoring'].get("page_for_expected_runtime", False):
```
qui: I think this conditional is the opposite of what you want?

solarkennedy: 😬

qui:
er I guess I should be asking for tests... I'll do that on the next review

PR #744, opened 2020-03-07
Add tron version and uptime to frontend navbar

askedrelic:
Should always be shown, even when changing tabs.

### DIFF
--- a/tronweb/coffee/models.coffee
+++ b/tronweb/coffee/models.coffee
@@ -60,7 +60,7 @@ class window.StatusModel extends Backbone.Model
         "/status/"
 
     parse: (resp) =>
-        booted = moment.unix(resp['boot_time']).format('YY-MM-DD HH:mm ZZ')
+        booted = moment.unix(resp['boot_time']).format('YYYY-MM-DD HH:mm ZZ')
         uptime = moment.duration(moment() - booted).minutes()
         $('#version').html('<b>Tron:</b> v' + resp['version'] + ' <b>Boot:</b> ' + booted + '</b>')
         resp
--- a/tronweb/coffee/models.coffee
+++ b/tronweb/coffee/models.coffee
@@ -54,6 +54,20 @@ class window.RefreshModel extends Backbone.Model
             console.log("scheduled with " + @interval)
             @timeout = setTimeout(@doRefresh, @interval)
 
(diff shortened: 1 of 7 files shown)


solarkennedy:
Could it be in the footer instead?

tronweb/coffee/models.coffee
```diff
@@ -54,6 +54,20 @@ class window.RefreshModel extends Backbone.Model
...
+        super options
+        options = options || {}
+
+    urlRoot: ->
+        "/status/"
+
+    parse: (resp, options) =>
+        booted = moment.unix(resp['boot_time']).format()
+        uptime = moment.duration(moment() - booted).minutes()
+        $('#version').html('Tron: <b>v' + resp['version'] + '</b>  Uptime: <b>' + uptime + 'm</b> Boot: <b>' + booted + '</b>')
```
qui: I think it would be easier to read if you bold the keys (Tron, Boot) instead of the values.
Also, I wrote uptime in the ticket, but I think boot time is easier to use, and we probably don't need both?

askedrelic: Sure, I removed uptime and tried simplifying the format of the boot time.

askedrelic:
Screen Shot 2020-03-12 at 4 11 16 PM

New version for review

tronweb/coffee/models.coffee
```diff
@@ -54,6 +54,16 @@ class window.RefreshModel extends Backbone.Model
             console.log("scheduled with " + @interval)
             @timeout = setTimeout(@doRefresh, @interval)
 
+class window.StatusModel extends Backbone.Model
+
+    urlRoot: ->
+        "/status/"
+
+    parse: (resp) =>
+        booted = moment.unix(resp['boot_time']).format('YY-MM-DD HH:mm ZZ')
```
qui: I think YYYY would make the date easier to read. Otherwise, the formatting looks good to me.

solarkennedy: https://www.reddit.com/r/ISO8601/

askedrelic: Sure, I was thinking about trying to shorten this since I'd assume there is no tron server booted in 19XX, but keeping consistent formatting is probably more valuable.

PR #747, opened 2020-03-16
Pinning nodes in tronweb dependency graph to make it look sane

dubeyrahul:
This change allows users to pin a node, move it around, and unpin the node using double-click. It does so by making the node fixed on `dragstart` event, and making it un-fixed on `dblclick`
It allows users to arrange the tronweb dependency graph to stay **static** once organized, instead of moving to a random state even when users try to organize nodes.
Manually tested by adding more nodes and dependencies in the tronfig and then arranging the tronweb dependency graph.
Note: This change however does not automatically organize the graph to a left-to-right looking DAG, that can be an additional improvement later.

### DIFF
--- a/tronweb/coffee/graph.coffee
+++ b/tronweb/coffee/graph.coffee
@@ -50,7 +50,8 @@ class window.GraphView extends Backbone.View
         @node = @svg.selectAll(".node")
             .data(data)
             .enter().append("svg:g")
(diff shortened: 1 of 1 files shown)


dubeyrahul:
Looks like the integration fails while importing `Feature`. Is this a known issue?

PR #752, opened 2020-04-14
Save state to dynamodb in thread to avoid blocking reactor loop

keymone:


tron/serialize/runstate/dynamodb_state_store.py
```diff
@@ -88,23 +92,36 @@ def _merge_items(self, first_items, remaining_items) -> dict:
...
+                    # Remove all previous data with the same partition key
+                    # TODO: only remove excess partitions if new data has fewer
+                    self._delete_item(key)
+                    self[key] = pickle.dumps(val)
+                except Exception as e:
+                    error = (
+                        'tron_dynamodb_save_failure: failed to save key'
+                        f'"{key}" to dynamodb:\n{repr(e)}'
+                    )
+                    log.error(error)
```
Rob-Johnson, line 114: are you sure this is all we want to do? I'm kind of concerned we'll get into the sort of situation we do in taskproc where threads die (and dump an exception along the way) without us noticing. might it be worth us raising the exception here?

keymone: well in this case we're not killing the thread, i just rely on aws dynamodb client to continue working i guess, not sure if it's a good assumption. there are multiple reasons we can get exception here - throttling, network timeout, etc, not all of them are terminal. in fact i think i should probably add a retry here for that key/value.

how about i add an error counter and throw a system kill after say 100 errors?

PR #755, opened 2020-05-06
Bump requirements to pick up new task_proc version

drmorr0:
(COMPINFRA-333) The task_processing library handled LIKE constraints
using a "match" expression instead of a "fullmatch" expression, which
meant that "pool: stable" jobs would get scheduled on the stable_batch
pool.  This version bump fixes that behaviour.


drmorr0:
This also hypothetically will fix the security notice we're getting about Twisted.

PR #759, opened 2020-07-24
Tron 1527: save job runs separately in state store from job state

qui:
Here's the part that actually saves the job run state separately, based on the events added in https://github.com/Yelp/Tron/pull/758/files.

The diff is big because I needed to add support for delete to the state stores, otherwise old job run state would never get deleted.

This is not backwards compatible, so I wrote a script that can go back and forth. My upgrade plan is:
1) stop Tron, install new version
2) run the script
3) start Tron. if this doesn't work for whatever reason, run the script to migrate state back, then install the old version and restart.

I've tested this a bunch of times back in forth with a local dev cluster.

### DIFF
--- a/tron/serialize/runstate/dynamodb_state_store.py
+++ b/tron/serialize/runstate/dynamodb_state_store.py
@@ -106,7 +106,7 @@ class DynamoDBStateStore(object):
                     log.info(f"save queue size {qlen} > {MAX_SAVE_QUEUE}, sleeping 5s")
                     time.sleep(5)
                     continue
-                self.save_queue[key] = pickle.dumps(val)
+                self.save_queue[key] = val
                 break
 
     def _consume_save_queue(self):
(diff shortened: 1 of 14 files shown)


tron/serialize/runstate/dynamodb_state_store.py
```diff
@@ -106,7 +106,7 @@ def save(self, key_value_pairs) -> None:
                     log.info(f"save queue size {qlen} > {MAX_SAVE_QUEUE}, sleeping 5s")
                     time.sleep(5)
                     continue
-                self.save_queue[key] = pickle.dumps(val)
+                self.save_queue[key] = val
```
keymone, line 109: just realized this, i don't think ordereddict is thread-safe like dequeue was. we may need to add locking here and when popping.

qui: ah, good point. here's a PR for that: https://github.com/Yelp/Tron/pull/760

PR #765, opened 2020-08-04
don't use public pypi for internal builds

drmorr0:


debian/rules
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/make -f
 # -*- makefile -*-
 
-PIP_INDEX_URL ?= https://pypi.python.org/simple
```
kawaiwanyelp, line 4: Given that this is an open source project, should we do an `if` here to check if we're in a Yelp ecosystem before we set this like in Tron's `Makefile`?

drmorr0: no longer relevant :)

PR #774, opened 2020-09-09
redact AWS credentials from the logs

drmorr0:
We were leaking AWS creds in the stderr Mesos logs, this change redacts them.

### DIFF
--- a/tron/mesos.py
+++ b/tron/mesos.py
@@ -1,5 +1,6 @@
 import json
 import logging
+import re
 import socket
 import time
 from urllib.parse import urlparse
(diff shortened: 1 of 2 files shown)


tron/mesos.py
```diff
@@ -164,10 +165,14 @@ def __init__(self, id, task_config, serializer=None):
         self.log = self.get_event_logger()
         self.setup_output_logging()
 
+        config_str = str(self.get_config())
+        # AWS_SECRET_ACCESS_KEYs are Base64-encoded
```
qui, line 169: Out of curiosity, how does the base64 encoding affect this code? should the test have base64 encoding too?

qui (approved):

tron/mesos.py
drmorr0: It just affects the regex.  Lemme make that comment a bit clearer.

PR #775, opened 2020-09-10
TRON-1563: save state and configs for retries independently

qui:
This just changes the state we store - I'll follow up with changes to update the UI to include this information (TRON-630) and allow the config to change between retries (TRON-1161). The only visible difference is that exit_statuses now includes the last exit status as well.


tron/core/actionrun.py
```diff
@@ -450,6 +525,8 @@ def _done(self, target, exit_status=0):
             self.clear_trigger_timeout()
             self.exit_status = exit_status
             self.end_time = timeutils.current_time()
+            if self.last_attempt is not None and self.last_attempt.end_time is None:
```
keymone, line 528: here and in other places: is it possible to have `.stop` called when there is no `last_attempt`? maybe worth just bailing at the start of the method with an error?

keymone (approved):
lgtm

tron/core/actionrun.py
qui: Yeah, it's possible if someone uses `tronctl fail` on a run that hasn't started yet. I think it's safer to just handle it.

PR #776, opened 2020-09-11
Fix up job pages on tronweb2 and add searching

qui:
Added searching by job name, both from the navbar and the homepage. Also made some cosmetic fixes based on what I saw on production clusters - long job names, and trying to make the action graph more legible.

### DIFF
--- a/tronweb2/src/components/ActionGraph/ActionGraph.js
+++ b/tronweb2/src/components/ActionGraph/ActionGraph.js
@@ -31,13 +31,6 @@ function setNodeCol(action, actionData, nextColPerLevel) {
   });
 }
 
-function shortenName(name) {
-  if (name.length > 20) {
-    return `${name.substring(0, 20)}...`;
-  }
-  return name;
-}
-
 function extendActionData(actionList) {
   const actionMap = {};
   actionList.forEach((action) => {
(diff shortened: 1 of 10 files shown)


tronweb2/src/components/ActionGraph/ActionGraph.js
```diff
@@ -79,21 +72,62 @@ function extendActionData(actionList) {
...
-        <circle r="6" cx={nodeXScale(d.col)} cy={nodeYScale(d.level)} />
-        <text dx={nodeXScale(d.col) + 10} dy={nodeYScale(d.level)} className={d.type}>
+        <rect
+          x={nodeXScale(d.col) - rectSidePadding}
+          y={nodeYScale(d.level) - rectHeight / 2}
+          rx="5"
+          ry="5"
+          width={(shortenName(d.name).length * 7) + (2 * rectSidePadding) + 1}
+          height={rectHeight}
+          stroke="black"
```
drmorr0, line 125: I feel like the black is a bit harsh here.  Could we lighten it a little bit, or use a drop-shadow instead?

I also don't care that much so feel free to drop.

tronweb2/src/components/JobsDashboard/JobsDashboard.js
```diff
@@ -49,6 +48,14 @@ function JobsDashboard() {
       <span className="sr-only">Loading...</span>
     </div>
   );
+
+  function filterJobs(string) {
+    const trimmed = string.trim().toLowerCase();
+    const { length } = string;
+    return length === 0 ? [] : jobData.jobs
+      .filter((job) => job.name.toLowerCase().slice(0, length) === trimmed);
```
drmorr0, line 56: Maybe just use `match` instead of `slice`?

tronweb2/src/components/JobsDashboard/JobsDashboard.js
```diff
@@ -58,7 +65,18 @@ function JobsDashboard() {
...
     } else {
-      jobContent = buildJobTable(jobData);
+      let jobsListToShow = jobData.jobs;
+      if (inputValue !== '') {
+        jobsListToShow = filterJobs(inputValue);
+      }
+      jobContent = (
+        <div>
+          <div className="mb-3">
+            <input type="text" className="form-control" placeholder="Search jobs" onInput={(e) => setInputValue(e.target.value)} />
```
drmorr0, line 75: If we're going to keep it a prefix match can we change the placeholder?

tronweb2/src/components/NavBar/NavBar.js
```diff
@@ -1,10 +1,46 @@
...
+  function getSuggestedJobNames(string) {
+    const trimmed = string.trim().toLowerCase();
+    const { length } = string;
+    return length === 0 ? [] : jobNames
+      .filter((jobName) => jobName.toLowerCase().slice(0, length) === trimmed)
+      .slice(0, maxSuggestions);
+  }
+
+  const inputProps = {
+    placeholder: 'Search jobs',
```
drmorr0, line 37: I sorta wonder if there's value in having the search bar in both the navbar and the body of the "Scheduled Jobs" page, but either way.

tronweb2/src/components/ActionGraph/ActionGraph.js
qui: I'd like a drop shadow too, but it's pretty complex to define in svg, so I don't want to spend too much time getting it right for now.

tronweb2/src/components/NavBar/NavBar.js
qui: The Scheduled Jobs page has more info on it, so you can look at the state of several jobs by filtering, as opposed to the navbar for just navigating to a single job.

tronweb2/src/components/JobsDashboard/JobsDashboard.js
qui: I'm using the Fuse library now, which should be much more powerful.

PR #777, opened 2020-09-30
Only reconfigure jobs in that namespace when a namespace is updated

qui:
map_to_job_and_schedule does all the updating - this change reduces the number of jobs that it gets called on.

This should make setup_tron_namespace faster (TRON-1523), and reduce API slowness overall.

### DIFF
--- a/tron/api/controller.py
+++ b/tron/api/controller.py
@@ -190,14 +190,15 @@ class ConfigController(object):
 
         old_config = self.read_config(name)['config']
         try:
+            log.info(f"Reconfiguring namespace {name}")
             self.config_manager.write_config(name, content)
-            self.mcp.reconfigure()
+            self.mcp.reconfigure(namespace=name)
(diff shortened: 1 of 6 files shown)


tron/api/controller.py
```diff
@@ -190,14 +190,15 @@ def update_config(self, name, content, config_hash):
 
         old_config = self.read_config(name)['config']
         try:
+            log.info(f"Reconfiguring namespace {name}")
             self.config_manager.write_config(name, content)
-            self.mcp.reconfigure()
+            self.mcp.reconfigure(namespace=name)
         except Exception as e:
-            log.error("Configuration update failed: %s" % e)
+            log.error(f"Configuration for {name} update failed: %s" % e)
```
keymone, line 197: nit: pull `e` into the f-string

keymone (approved):
nice, ship!

PR #778, opened 2020-10-05
TRON-1161: update configs for retries

qui:
## expected behavior
All retries (both manual and automatic) of an action run will use the latest values in tronfig for any field in ActionCommandConfig: https://github.com/Yelp/Tron/blob/master/tron/core/action.py#L15-L25

There are two exceptions:
1) retries use the same command as before, by default. Manual retries have an option to use the latest command with `tronctl retry {action_name} --use-latest-command`
2) This uses the action name as a key, so it only works if the name of the action run being retried is still in the current tronfig.

## notes

The first commit does the main work here, by changing the update_from_job method: https://github.com/Yelp/Tron/commit/cdf07f1712563641fea63531ea02e6005fd38181#diff-3e3afbf3f0f967c512b5f8c0a05c8b85R151

That will update job run state for every job run in a job when that job gets reconfigured. However, I don't think we really need to keep current command config in state, which is what "Use current command config instead of state on restore" does. This works in most cases - I think the only time we'll end up with a [missing command config](https://github.com/Yelp/Tron/compare/TRON-1161-retry-configs?expand=1#diff-4ec93da02092ad4b6d3ac74e2ea31871R485) is if the following things happen:
1) action is renamed or removed in config
2) Tron is restarted (so it loses the old config in memory)
3) One of the old renamed/removed actions is started or retried

I think this is rare enough that we don't have to worry about it, but I'm open to argument on it. Since we have to keep the state around for a little bit for rollback safety anyway, we could save this commit for a later version and see how bad the reconfiguration time gets.

Tested with `make dev` with different commands and everything seems to work as expected.


tron/api/controller.py
```diff
@@ -86,12 +87,12 @@ def handle_termination(self, command):
...
             return msg % (command, e)
 
-    def handle_retry(self):
+    def handle_retry(self, use_original_command):
         cleanup_run = self.job_run.action_runs.cleanup_action_run
         if cleanup_run and cleanup_run.is_done:
             return "JobRun has run a cleanup action, use rerun instead"
 
-        if self.action_run.retry():
+        if self.action_run.retry(original_command=use_original_command):
```
keymone, line 95: rename the argument for consistency?

tron/core/action.py
```diff
@@ -31,6 +31,13 @@ def state_data(self):
             for field in fields(self)
         }
 
+    def copy(self):
+        kwargs = {
+            field.name: getattr(self, field.name)
+            for field in fields(self)
+        }
+        return ActionCommandConfig(**kwargs)
```
keymone, line 39: i think this should work: `return ActionCommandConfig(**self.__dict__)` or use the property defined just above: `return ActionCommandConfig(**self.state_data)`

keymone (approved):
lgtm, only minor nitpicks

PR #788, opened 2021-04-13
Tronctl success no longer emits trigger when the transition is invalid

lanceleih:
### Description
Previously, we used to emit triggers before we transitioned to the next state. However, this caused some confusion since the transition could be invalid (i.e. no next state to transition to). This change ensures that the transition is first valid before we can emit triggers.

### Tests
`make test` - PASSED

### DIFF
--- a/tron/core/actionrun.py
+++ b/tron/core/actionrun.py
@@ -646,9 +646,12 @@ class ActionRun(Observable):
         ]
 
     def success(self):
-        if self.trigger_downstreams:
-            self.emit_triggers()
-        return self._done('success')
+        transitionvalid = self._done('success')
+        if transitionvalid:
+            if self.trigger_downstreams:
+                self.emit_triggers()
+
+        return transitionvalid
 
     def fail_unknown(self):
         """Failed with unknown reason."""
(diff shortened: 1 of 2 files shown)


tron/core/actionrun.py
```diff
@@ -646,9 +646,12 @@ def remaining_triggers(self):
...
     def success(self):
-        if self.trigger_downstreams:
-            self.emit_triggers()
-        return self._done('success')
+        transitionvalid = self._done('success')
+        if transitionvalid:
+            if self.trigger_downstreams:
+                self.emit_triggers()
+
+        return transitionvalid
```
nemacysts, line 654: super-duper-nit: we should use `snake_case` for variables
```suggestion
        transition_valid = self._done('success')
        if transition_valid:
            if self.trigger_downstreams:
                self.emit_triggers()

        return transition_valid
```

nemacysts (approved):
lgtm - probably want qui to take a look before merging tho :p

qui (approved):
yep, i agree with the snake_case comment, but looks good to me!

tron/core/actionrun.py
lanceleih: Done 👍

PR #789, opened 2021-04-14
Implementation of skip-and-publish

nemacysts:
Skipping an action that has downstream triggers is error-prone as its easy to forget that you have triggers to emit and what triggers to emit. You prob need to go check your tronfig definitions to see what triggers exist and converting the trigger name to a trigger command can be a little tricky (with date magic).

To solve this, we add a new tronctl command that will skip an action and then publish all triggers for that action

Testing: mostly manual, there are no tronctl tests. `make test itest_bionic` pass, though!

### DIFF
--- a/bin/tronctl
+++ b/bin/tronctl
@@ -191,15 +191,15 @@ def control_objects(args: argparse.Namespace):
                 # query the API and figure out what triggers exist
                 triggers = _get_triggers_for_action(server=args.server, action_identifier=identifier)
                 if triggers is None:
+                    print("Encountered error getting triggers to publish.")
                     # tronctl allows for multiple ids to be passed in for commands, but we want to make it obvious that
                     # we failed to get the downstream triggers so that users know to take action (and don't miss the
                     # error amongst other tronctl output).
-                    print("Encountered error getting triggers to publish.")
                     yield False
                 elif not triggers:
+                    print(f"{identifier} has no triggers to publish.")
                     # TODO: should we check this up-front and refuse to skip if there are no triggers that will be
                     # published rather than carry on under the assumption that the user copy-pasted/typo'd the identifier?
-                    print(f"{identifier} has no triggers to publish.")
                     yield True
 
                 # TODO: this loop should use event_publish(), but we'd need to refactor how the CLI works and stop passing
@@ -210,10 +210,10 @@ def control_objects(args: argparse.Namespace):
                         data={"command": "publish", "event": trigger}
                     )
             else:
+                print(f"Unable to skip {identifier}.")
                 # again, tronctl allows for multiple ids to be passed in for commands, but we want to make it obvious that
                 # we failed to skip something so that users know to take action (and don't miss the error amongst other
                 # tronctl output)
-                print(f"Unable to skip {identifier}.")
                 yield False
 
         else:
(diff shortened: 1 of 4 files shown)


nemacysts:
the diffstat looks big, but there's a lot of comments inflating that number :p 

I'm not super happy with how I had to implement this, but I also didn't wanna go down a massive refactoring hole on what was supposed to be a starter ticket

bin/tronctl
```diff
@@ -129,7 +135,33 @@ def event_discard(args):
...
+            f"/api/jobs/{namespace}.{job_name}/{run_number}/{action_name}",
+        ),
+    )
+    if trigger_response.error:
+        print(f"Unable to fetch downstream triggers for {action_identifier}: {trigger_response.error}")
+        return None
+
+    # triggers are returned by the API as comma-separated values with a space after every comma, which is
+    # not automation-friendly - thus the string-manipulation
+    triggers = trigger_response.content.get('trigger_downstreams', '').replace(" ", "").split(',')
```
qui, line 157: I think `.split(', ')` would do the same thing, so you can skip the replace.

bin/tronctl
```diff
@@ -147,12 +179,50 @@ def control_objects(args):
...
+                    # error amongst other tronctl output).
+                    print("Encountered error getting triggers to publish.")
+                    yield False
+                elif not triggers:
+                    # TODO: should we check this up-front and refuse to skip if there are no triggers that will be
+                    # published rather than carry on under the assumption that the user copy-pasted/typo'd the identifier?
+                    print(f"{identifier} has no triggers to publish.")
+                    yield True
+
+                # TODO: this loop should use event_publish(), but we'd need to refactor how the CLI works and stop passing
```
qui, line 205: If we stopped passing around the full set of args, we could also make this a separate command function rather than adding this to `control_objects`, right?

I agree that would make the code a lot nicer, but I'm ok with following up with it later.

bin/tronctl
```diff
@@ -147,12 +179,50 @@ def control_objects(args):
...
+            # (i.e., this file) to implement this functionality
+            if request(
+                url=urljoin(args.server, tron_id.url),
+                data={"command": "skip"}
+            ):
+                # a single action can have 0..N triggers to publish and these can be arbitrarily named, so we need to
+                # query the API and figure out what triggers exist
+                triggers = _get_triggers_for_action(server=args.server, action_identifier=identifier)
+                if triggers is None:
+                    # tronctl allows for multiple ids to be passed in for commands, but we want to make it obvious that
```
qui, line 194: I'm not sure what decision you're explaining with this comment - is it the `yield False`, so it exits early?

nemacysts: yup - same with the comment below. do you think moving the comment(s) so that the yield(s) follow immediately after would make that clearer? open to suggestions :)

bin/tronctl
nemacysts: yup!

bin/tronctl
nemacysts: oh true, thanks!

bin/tronctl
qui: I don't think you need to necessarily justify it separately, because other operations also yield the result of `request` after every ID, and could fail. (I suppose you changed the behavior of request to make it possible to fail in this PR, but I think exiting early makes sense in general.)

nemacysts: ah cool - I'll remove these comments then

nemacysts: done!

qui (approved):

PR #790, opened 2021-04-14
Initial implementation of GH Actions

nemacysts:
TravisCI is unusable for us and a lot of our OSS projects have already
switched. This PR is mostly based on @EvanKrall's work in Yelp/paasta.


nemacysts:
Tested in https://github.com/nemacysts/Tron/actions/runs/750067234 - which I guess also shows up on this PR?

yelp_package/itest_dockerfiles/mesos/Dockerfile
```diff
@@ -24,9 +24,12 @@ RUN apt-get update > /dev/null && \
 
 RUN echo "deb http://repos.mesosphere.com/ubuntu xenial main" > /etc/apt/sources.list.d/mesosphere.list && \
     apt-key adv --keyserver keyserver.ubuntu.com --recv 81026D0004C44CF7EF55ADF8DF7D54CBE56151BF && \
+    curl -fsSL https://download.docker.com/linux/ubuntu/gpg | gpg --dearmor -o /usr/share/keyrings/docker-archive-keyring.gpg && \
+    echo "deb [arch=amd64 signed-by=/usr/share/keyrings/docker-archive-keyring.gpg] https://download.docker.com/linux/ubuntu xenial stable" > /etc/apt/sources.list.d/docker.list && \
     apt-get update > /dev/null && \
     DEBIAN_FRONTEND=noninteractive apt-get install -y \
-        docker.io=18.09.7-0ubuntu1~16.04.5 \
+        docker-ce \
```
qui, line 31: out of curiosity, why is this change required?

qui (approved):

yelp_package/itest_dockerfiles/mesos/Dockerfile
nemacysts: @qui docker.io is the old name of the docker dpkg (https://docs.docker.com/engine/install/debian/#uninstall-old-versions) - and its no longer available on the repositories/PPAs we're using

PR #806, opened 2021-06-03
Add KubernetesTask scaffolding

nemacysts:
This sets up a basic skeleton of Tron's wrapper around a task_processing
task (and is very similar to the corresponding Mesos version: MesosTask)

### DIFF
--- a/tron/kubernetes.py
+++ b/tron/kubernetes.py
@@ -1,0 +1,93 @@
+import logging
+from logging import Logger
+
+from task_processing.interfaces.event import Event
+from task_processing.plugins.kubernetes.task_config import KubernetesTaskConfig
+
+import tron.metrics as metrics
+from tron.actioncommand import ActionCommand
+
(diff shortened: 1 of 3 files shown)


nemacysts:
(this is a draft until we have KubernetesTaskConfig available in task_proc)

tron/kubernetes.py
```diff
@@ -0,0 +1,93 @@
...
+            )
+            return
+
+        k8s_type = getattr(event, "platform_type", None)
+        self.log.info(f"Got event for task={event_id} (Kubernetes type={k8s_type}).")
+
+        try:
+            self.log_event_info(event=event)
+        except Exception:
+            self.log.exception(f"Unable to log event info for id={event_id}.")
```
lanceleih, line 91: We can also add a print message of the exception, but that can be done later on.

lanceleih (approved):
LGTM

tron/kubernetes.py
nemacysts: log.exception will already do that :) (https://devdocs.io/python~3.6/library/logging#logging.exception)

PR #812, opened 2021-06-10
install js dependencies in tox for use in pre-commit

nemacysts:
This will make it so that JS-based pre-commit hooks will work regardless
of how you invoke our tox testenv.


lanceleih (approved):
LGTM

PR #813, opened 2021-06-11
Add check-requirements and mypy to our tox tests

nemacysts:
* we have requirements-tools installed, but aren't automatically running
  it
* we're starting to add types to new code, so let's have mypy
  automatically check it instead of relying on our editor setups to run
  it for us :p


lanceleih (approved):
LGTM

PR #814, opened 2021-06-11
Scaffold out KubernetesCluster (+ some initial implementation)

nemacysts:
This sets up the skeleton for KubernetesCluster (a k8s version of
MesosCluster) and does some of the basic implementation for several
methods when possible/easy

### DIFF
--- a/tests/kubernetes_test.py
+++ b/tests/kubernetes_test.py
@@ -29,6 +29,14 @@ def mock_kubernetes_cluster():
         yield KubernetesCluster("kube-cluster-a:1234")
 
 
+@pytest.fixture
+def mock_disabled_kubernetes_cluster():
+    with mock.patch("tron.kubernetes.PyDeferredQueue", autospec=True,), mock.patch(
+        "tron.kubernetes.TaskProcessor", autospec=True,
+    ):
+        yield KubernetesCluster("kube-cluster-a:1234", enabled=False)
+
+
 def mock_event_factory(
     task_id: str,
     platform_type: str,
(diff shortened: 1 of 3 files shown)


tron/kubernetes.py
```diff
@@ -91,3 +104,206 @@ def handle_event(self, event: Event) -> None:
...
+
+    def handle_next_event(self, _=None) -> None:
+        """
+        Pops events off of the shared tron<->task_proc queue and processes them.
+
+        We only expect two types of events:
+        * control: events regarding how the task_proc plugin is running - handled directly
+        * task: events regarding how the actual tasks/Pods we're running our doing - forwarded to KubernetesTask
+        """
+        if self.deferred and not self.deferred.called:
```
lanceleih, line 183: Are we checking if self.deferred has been called? I'd assume it to be true if we already have handlers waiting. When does self.deferred.called become true?

nemacysts: self.deferred will either be `None` or a `PyDeferred` object (thus the `if self.deferred`), and until one of the {Call,Err}backs has been called, `self.deferred.call` will be `False` - in this case, `called` will be `False` until `process_event()` or `logError` have been called

tests/kubernetes_test.py
```diff
@@ -73,3 +79,74 @@ def test_handle_event_exit_early_on_misrouted_event(mock_kubernetes_task):
...
+
+def test_stop_default(mock_kubernetes_cluster):
+    # When stopping, tasks should not exit. They will be recovered
+    mock_task = mock.MagicMock()
+    mock_kubernetes_cluster.tasks = {"task_id": mock_task}
+    mock_kubernetes_cluster.stop()
+    assert mock_kubernetes_cluster.deferred is None
+    assert mock_task.exited.call_count == 0
+    assert len(mock_kubernetes_cluster.tasks) == 1
+
```
lanceleih, line 147: Can we also test set_enabled? Then we can check if connect() and stop() are being called properly.

nemacysts: done!

lanceleih (approved):
LGTM!

PR #815, opened 2021-06-22
Add scaffolding for KubernetesClusterRepository

nemacysts:
This is generally patterned after the existing MesosClusterRepository.

NOTE: I had to make some configuration-related changes since it seems
like I missed a couple places in the initial config scaffolding commit
and didn't notice until I added code that actually read said config.


lanceleih (approved):
LGTM

PR #816, opened 2021-07-01
Store kubeconfig path in tron settings, not k8s master address

nemacysts:
kubeconfig files already contain the master address, so no need to
duplicate that here. additionally, we'll need to know what kubeconfig to
use, so lets replace master_address with kubeconfig_path


lanceleih (approved):
LGTM

PR #821, opened 2021-07-23
TRON-1637: Add support for secret_env key on action configs

jfongatyelp:
*Note*: This PR Is based off of #818. I will update when that is merged. 

With k8s jobs, we will be configuring environment variables from Kubernetes Secret resources.  To do this, we need to know the secret resources' names and keys.

This adds a new `secret_env` key to `TronActionConfig`, which will map an envvar to a `ConfigSecretSource`, whose keys (`secret`, `key`)  map to the required values of a Kubernetes [SecretKeySelector](https://v1-18.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#secretkeyselector-v1-core) (`name`, `key`).

This required adding a new dict validator to verify values of a dict passed an expected validator. The existing `build_dict_name_validator` presumed the type being validated had a `name` field, which is not the case for our `ConfigSecretSource`.

### DIFF
--- a/tests/config/config_parse_test.py
+++ b/tests/config/config_parse_test.py
@@ -675,6 +675,7 @@ class TestValidateJobs(TestCase):
                             docker_image="my_container:latest",
                             docker_parameters=[dict(key="label", value="labelA"), dict(key="label", value="labelB"),],
                             env=dict(USER="batch"),
+                            secret_env=dict(MY_SECRET=dict(secret="k8s-secret-name", key="secret_key")),
                             extra_volumes=[dict(container_path="/tmp", host_path="/home/tmp", mode="RO",),],
                         ),
                         dict(
@@ -709,6 +710,7 @@ class TestValidateJobs(TestCase):
                             schema.ConfigParameter(key="label", value="labelB",),
                         ),
                         env={"USER": "batch"},
+                        secret_env={"MY_SECRET": schema.ConfigSecretSource(secret="k8s-secret-name", key="secret_key")},
                         extra_volumes=(
                             schema.ConfigVolume(
                                 container_path="/tmp", host_path="/home/tmp", mode=schema.VolumeModes.RO.value,
--- a/tests/config/config_parse_test.py
+++ b/tests/config/config_parse_test.py
@@ -207,6 +207,29 @@ def make_master_jobs():
             cleanup_action=None,
             expected_runtime=datetime.timedelta(1),
         ),
+        "MASTER.test_job_k8s": make_job(
+            name="MASTER.test_job_k8s",
+            node="NodePool",
+            schedule=schedule_parse.ConfigDailyScheduler(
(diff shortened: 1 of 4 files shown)


tron/config/config_utils.py
```diff
@@ -197,6 +198,19 @@ def validator(value, config_context):
     return validator
 
 
+def build_dict_value_validator(item_validator, allow_empty=False):
+    """Build a validator which validates values of a dict, and returns a dict"""
+
+    def validator(value, config_context):
+        msg = "Duplicate name %%s at %s" % config_context.path
```
nemacysts, line 205: are the double-% on purpose here?

tron/config/config_parse.py
```diff
@@ -249,6 +251,17 @@ class ValidateVolume(Validator):
 valid_volume = ValidateVolume()
 
 
+class ValidateSecretSource(Validator):
+    config_class = ConfigSecretSource
+    validators = {
+        "secret": valid_string,
```
nemacysts, line 257: should we add some comments about what these keys will be holding? (i.e., `secret` isn't actually holding plaintext secrets)

tron/config/config_utils.py
jfongatyelp: Yeah, this was a literal copy/paste of the `build_dict_name_validator` -- this actually feels a bit unnecessary, given we're already assuming we're working with a dict which will prevent us from having duplicate keys.  I'll go ahead and try to return a standard dict instead of this.

tron/config/config_parse.py
jfongatyelp: I will definitely add comments to help readers, but you think this confusing enough that the config key should be named more clearly, eg `secret_name`?

tests/config/config_parse_test.py
```diff
@@ -675,6 +675,7 @@ def test_valid_jobs_success(self):
                             docker_image="my_container:latest",
                             docker_parameters=[dict(key="label", value="labelA"), dict(key="label", value="labelB"),],
                             env=dict(USER="batch"),
+                            secret_env=dict(MY_SECRET=dict(secret="k8s-secret-name", key="secret_key")),
```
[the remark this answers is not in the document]
jfongatyelp: Let me know if the new tests make sense (tl;dr: I added a k8s test job to the expected `master` config, which included one valid secret), then tested the 3 possible cases for ConfigSecretSource validation (missing required key, unknown key, and success).

nemacysts (approved):

tron/config/config_parse.py
nemacysts: it's probably fine as-is with the comment, but I wouldn't be against naming this key `secret_name`

jfongatyelp: Agreed, i've done this and updated Paasta#3154 to match

PR #823, opened 2021-07-28
TRON-1611 Handle KubernetesTask State Machine Transitions

lanceleih:
### Description
Tron now checks to see which state machine it will transition to with every new pod event. Pod events are based on the pod lifecycle phases.

### Test
`make test - PASSED`

### DIFF
--- a/tron/kubernetes.py
+++ b/tron/kubernetes.py
@@ -129,8 +129,10 @@ class KubernetesTask(ActionCommand):
             self.exited(0)
         elif k8s_type == "FAILED":
             self.exited(1)
-        elif k8s_type == "unknown":
-            self.log.warning("Kubernetes does not know anything about this task, it is UNKOWN")
+        elif k8s_type == "lost":
+            # Using 'lost' instead of 'unknown' for now until we are sure that before reconcile() is called,
+            # the tasks inside task_metadata map are all UNKNOWN
+            self.log.warning("Kubernetes does not know anything about this task, it is LOST")
             self.log.warning(
                 "This can happen for any number of reasons, and Tron can't know if the task ran or not at all!"
             )
--- a/tron/kubernetes.py
+++ b/tron/kubernetes.py
@@ -123,13 +123,11 @@ class KubernetesTask(ActionCommand):
         except Exception:
             self.log.exception(f"Unable to log event info for id={event_id}.")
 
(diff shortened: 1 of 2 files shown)


tron/kubernetes.py
```diff
@@ -123,7 +123,42 @@ def handle_event(self, event: Event) -> None:
...
+        if k8s_type == "pending":
+            pass
+        elif k8s_type == "running":
+            self.started()
+        elif k8s_type == "succeeded":
+            self.exited(0)
+        elif k8s_type == "failed":
+            self.exited(1)
+        elif k8s_type == "unknown":
+            self.log.warning("Kubernetes does not know anything about this task, it is UNKOWN")
```
jfongatyelp, line 135: This is interesting, I feel like taskproc isn't submitting all of these `platform_type`s, we only seem to submit `task_event`s with:
* `killed`
* `finished`
* `FAILED` (uh oh, was this a typo?)
* `running`
* `unknown`

Are we sending the others (`pending`, `succeeded`, `failed` (I still think that's a typo)) from some other source?

lanceleih: Nope. This was my mistake! I was misunderstood and was using pod phases instead of the ones emitted by TaskProc. Changed it.

jfongatyelp (approved):
I think this looks sane to me, though I'm not all that familiar with what these KubernetesTask methods actually do.  I think it makes sense to push this out so we can iterate!

tron/kubernetes.py
```diff
@@ -123,7 +123,42 @@ def handle_event(self, event: Event) -> None:
...
+            self.log.warning(
+                "This can happen for any number of reasons, and Tron can't know if the task ran or not at all!"
+            )
+            self.log.warning("If you want Tron to RUN it (again) anyway, retry it with:")
+            self.log.warning(f"    tronctl retry {self.id}")
+            self.log.warning("If you want Tron to NOT run it and consider it as a success, skip it with:")
+            self.log.warning(f"    tronctl skip {self.id}")
+            self.log.warning("If you want Tron to NOT run it and consider it as a failure, fail it with:")
+            self.log.warning(f"    tronctl fail {self.id}")
+            self.exited(None)
```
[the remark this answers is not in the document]
lanceleih: Modified it to be checking for "lost" instead of "unknown".

PR #824, opened 2021-07-30
Install yelp-meteorite when building yelp env

kawaiwanyelp:
### Description
Currently, we aren't installing yelp-meteorite, which means that task-processing isn't emitting any metrics. I copied paasta's approach and made it so that when we're inside Yelp, it'll build with yelp-meteorite.

### Testing
new: `make yelpy`
`make deb_bionic` - i saw the diff apply from `extra_requirements_yelp.txt` >> `requirements.txt`, which made the build install yelp-meteorite

### Notes
I won't release this until next week

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,13 @@ else
 	export PIP_INDEX_URL ?= https://pypi.python.org/simple
 endif
 
+NOOP = true
+ifeq ($(PAASTA_ENV),YELP)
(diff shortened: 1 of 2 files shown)


nemacysts (approved):
i wish there was a way to do this with environment markers so that we could just specify `task_processing[mesos_executor,k8s,metrics]` in requirements-minimal.txt so that check-requirements isn't side-stepped, but oh well :p

PR #826, opened 2021-08-02
Make copy of k8s tasks while iterating to stop them

nemacysts:
I incorrectly "fixed" some code when porting from Mesos -> Kubernetes
and introduced a bug

### DIFF
--- a/tron/kubernetes.py
+++ b/tron/kubernetes.py
@@ -309,7 +309,8 @@ class KubernetesCluster:
         self.queue = PyDeferredQueue()
 
         if fail_tasks:
-            for key, task in self.tasks.items():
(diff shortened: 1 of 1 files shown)


jfongatyelp (approved):

tron/kubernetes.py
```diff
@@ -309,7 +309,8 @@ def stop(self, fail_tasks: bool = False) -> None:
         self.queue = PyDeferredQueue()
 
         if fail_tasks:
-            for key, task in self.tasks.items():
+            # NOTE: we're turning this into a list on purpose: otherwise we're modifying the dict we're iterating over
```
jfongatyelp, line 312: breadcrumbs for the next person++ Thanks!

PR #828, opened 2021-08-06
Actually kill k8s Pods on KubernetesActionRun::kill()

nemacysts:
We had left this method stubbed out until we implemented kill()
in task_processing, but now we can actually finish implemention here.


jfongatyelp (approved):

lanceleih (approved):

nemacysts:
This needs to pull in a new task_proc for https://github.com/Yelp/task_processing/pull/176, but otherwise - things looked good in my tests.

jfongatyelp (approved):

PR #830, opened 2021-08-10
Make tronctl backfill actually run backfills 

kawaiwanyelp:
### Description
The PR makes it so that `tronctl backfill` doesn't print out commands for a user (that becomes the `--dry-run` mode), it actually runs the backfill for them. The purpose is to control the parallelism so that a user doesn't just run all the commands at once, which would overwhelm the cluster.

The way it works is:
1. User runs `tronctl backfill`
2. They will be prompted for confirmation for the dates they want to run backfills for
3. Using asyncio, backfills will be run concurrently. Up to `--max-parallel` backfills may run at any given time. The Tron API is polled for the state of a backfill every 10s.
4. Once all backfills are done, a final report of all the runs is printed out

There are a couple of error handling features:
- If the user does a ctrl-C/SIGINT, all in-progress runs will be `tronctl cancel`ed
- If `--fail-on-error` is set, the first failing backfill run will cause the same behavior

### Testing
`make test`
`tronctl backfill` on a test cluster, including failure cases, failure tolerating cases, cases with more dates to backfill than are allowed to run in paralle, SIGINT-ing


tron/commands/backfill.py
```diff
@@ -0,0 +1,279 @@
...
+import signal
+from typing import List
+from urllib.parse import urljoin
+
+from tron.commands import client
+from tron.commands import display
+from tron.core.actionrun import ActionRun
+
+DEFAULT_MAX_PARALLEL_RUNS = 10
+DEFAULT_POLLING_INTERVAL = 10  # seconds
```
nemacysts, line 15: thoughts on adding a `_S` suffix so that we don't need a ` # seconds` comment?
```suggestion
DEFAULT_POLLING_INTERVAL_S = 10 
```

tron/commands/backfill.py
```diff
@@ -0,0 +1,279 @@
...
+            else:
+                print(
+                    f"Warning: Job run for {self.run_time_str} created, but couldn't determine "
+                    "its name, so its state is considered to be unknown."
+                )
+                self.run_state = ActionRun.UNKNOWN
+
+        return self.run_name
+
+    async def get_run_id(self) -> client.TronObjectIdentifier:
```
nemacysts, line 121: isn't this actually
```suggestion
    async def get_run_id(self) -> Optional[client.TronObjectIdentifier]:
```
it looks like it's possible for us to not get a run_id back from `get_object_type_from_identifier` and just return the default `self.run_id` (which is `None`)

nemacysts (commented):
🔥

bin/tronctl
```diff
@@ -176,13 +240,13 @@ def control_objects(args: argparse.Namespace):
...
                     # published rather than carry on under the assumption that the user copy-pasted/typo'd the identifier?
                     yield True
-
-                # TODO: this loop should use event_publish(), but we'd need to refactor how the CLI works and stop passing
-                # around the full set of args everywhere to do so
-                for trigger in triggers:
-                    yield request(
-                        url=urljoin(args.server, "/api/events"), data={"command": "publish", "event": trigger},
-                    )
+                else:
```
nemacysts, line 243: whoops, nice catch!

nemacysts (approved):

lanceleih (approved):
LGTM

PR #831, opened 2021-08-11
Read k8s tronjob logs from scribereader

nemacysts:
This will only be enabled for k8s actions and, additionally, adds a
separate area for "meta" information that is hidden by default (since
it's usually only of interest to people on CI)


nemacysts:
need to add some tests, but this should be reviewable now

tron/api/adapter.py
```diff
@@ -149,8 +154,30 @@ def _get_alternate_output_paths(self):
...
+                    break
+        return output or []
+
     @toggle_flag("include_stdout")
-    def get_stdout(self):
+    def get_stdout(self) -> List[str]:
+        if isinstance(self._obj, KubernetesActionRun):
+            return read_log_stream_for_action_run(
+                action_run_id=self._obj.id,
+                component="stderr",
```
kawaiwanyelp, line 176: Is this right? Should this be stdout not stderr?

tron/utils/scribereader.py
```diff
@@ -0,0 +1,158 @@
...
+    from clog.readers import StreamTailerSetupError  # type: ignore
+except ImportError:
+    scribereader = None  # sorry folks, you'll need to add your own way to retrieve logs
+
+
+log = logging.getLogger(__name__)
+
+
+@lru_cache(maxsize=1)
+def get_region() -> str:
```
kawaiwanyelp, line 23: nit: The Python junkie in me says turn these three habitat getters into one function and three `functools.partial`s, but it's not at all necessary.

tron/utils/scribereader.py
```diff
@@ -0,0 +1,158 @@
...
+    try:
+        ecosystem = get_ecosystem()
+        superregion = get_superregion()
+        region = get_region()
+    except OSError:
+        log.warning("Unable to read location mapping files from disk, not returning scribereader host/port")
+        return None
+
+    host, port = scribereader.get_tail_host_and_port(
+        ecosystem=ecosystem if ecosystem != "prod" else None, region=region, superregion=superregion,
```
kawaiwanyelp, line 66: nit: I'd add a comment for why ecosystem should be `None` for prod, since it's not obvious at first glance that prod is different for scribereader.

tron/utils/scribereader.py
```diff
@@ -0,0 +1,158 @@
...
+    # NOTE: this is more-or-less what our internal `scribereader` binary does
+    if use_reader:
+        with scribereader.get_stream_reader(
+            stream_name=stream_name, min_date=min_date, max_date=max_date, reader_host=host, reader_port=port,
+        ) as stream:
+            for line in stream:
+                try:
+                    payload = json.loads(line)
+                except json.decoder.JSONDecodeError:
+                    log.error(f"Unable to decode log line from stream ({stream_name}) for {action_run_id}: {line}")
```
kawaiwanyelp, line 116: nit: I wonder if this is something that we should report in the output somehow, like whether or not there were a lot of lines that couldn't be decoded, but it's not strictly necessary.

tron/kubernetes.py
```diff
@@ -53,6 +54,12 @@ def __init__(self, action_run_id: str, task_config: KubernetesTaskConfig, serial
 
         self.log = self.get_event_logger()
 
+        config_str = str(self.get_config())
+        # AWS_SECRET_ACCESS_KEYs are base64-encoded so it uses alphanumerics plus +, /, and =
+        config_str = re.sub("'AWS_SECRET_ACCESS_KEY': '[a-zA-Z0-9+/=]+'", "AWS_SECRET_ACCESS_KEY_REDACTED", config_str)
+        config_str = re.sub("'AWS_ACCESS_KEY_ID': '[a-zA-Z0-9]+'", "AWS_ACCESS_KEY_ID_REDACTED", config_str)
+        self.log.info(f"Kubernetes task {self.get_kubernetes_id()} created with config {config_str}")
```
kawaiwanyelp, line 61: Is this an accidental commit? Feels like a debugging line to check AWS creds, and not something that should be permanent.

kawaiwanyelp (changes requested):
Still needs tests like you said, but also some changes need, mostly minor.

tron/kubernetes.py
nemacysts: I should probably add this in a separate PR - I was just comparing my new "meta" output to what is shown for Mesos and realized that we weren't showing users how their task was launched (this is more-or-less what we do for Mesos tasks)

tron/api/adapter.py
nemacysts: whoops!

tron/utils/scribereader.py
nemacysts: as in something like this?
```Python
from functools import partial
def read_file(path: str) -> str:
    with open(path) as f:
        return f.read().strip()

get_region = partial(read_file, "/nail/etc/region")
...
```

tron/utils/scribereader.py
nemacysts: hopefully this never actually gets hit :p 

that said, that's a good idea and I can add a line to the output somewhere with the count of undecodable lines pretty easily

tron/utils/scribereader.py
kawaiwanyelp: yeah :P but again im just being picky

tests/utils/scribereader_test.py
```diff
@@ -0,0 +1,144 @@
...
+                "component": "stderr",
+                "message": "line 3",
+                "timestamp": "2021-01-02T18:12:09.169421619Z"
+            }""",
+            ]
+        )
+        output = read_log_stream_for_action_run(
+            action_run_id="namespace.job.1234.action",
+            component="stdout",
+            min_date=datetime.datetime.now(),
```
kawaiwanyelp, line 51: nit: calling `now()` looks sketch because it makes the tests look like they depend on the current time, even though its not the case because they are just forwarded to scribereader. id either use as specific datetime or (more likely) comment explaining why this is fine.

tests/utils/scribereader_test.py
```diff
@@ -0,0 +1,144 @@
...
+            max_date=datetime.datetime.now() + datetime.timedelta(hours=1),
+        )
+
+    mock_stream_reader.assert_not_called()
+    mock_stream_tailer.assert_called_once()
```
kawaiwanyelp, line 56: also add called with what args?

tests/utils/scribereader_test.py
```diff
@@ -0,0 +1,144 @@
...
+        )
+        output = read_log_stream_for_action_run(
+            action_run_id="namespace.job.1234.action",
+            component="stdout",
+            min_date=datetime.datetime.now() - datetime.timedelta(days=5),
+            max_date=datetime.datetime.now(),
+        )
+
+    mock_stream_reader.assert_called_once()
+    mock_stream_tailer.assert_called_once()
```
kawaiwanyelp, line 110: same here, args called with here as well?

tests/utils/scribereader_test.py
```diff
@@ -0,0 +1,144 @@
...
+        # so lets make sure we don't call the tailer
+        mock_stream_tailer.return_value.__iter__.side_effect = Exception
+        output = read_log_stream_for_action_run(
+            action_run_id="namespace.job.1234.action",
+            component="stdout",
+            min_date=datetime.datetime.now() - datetime.timedelta(days=5),
+            max_date=datetime.datetime.now() - datetime.timedelta(days=4),
+        )
+
+    mock_stream_reader.assert_called_once()
```
kawaiwanyelp, line 142: args again?

kawaiwanyelp (approved):
fix-n-ship

tests/utils/scribereader_test.py
nemacysts: true! I'll add a comment

PR #835, opened 2021-08-17
Log k8s task config and hostname of agent

nemacysts:
We're missing these for parity with the logging done by our existing
Mesos tasks


kawaiwanyelp (approved):

nemacysts:
I originally had this in my PR for adding log viewing support, but it didn't really belong there :p

I also removed the regex for AWS keys since we'll never be passing plaintext secrets in lieu of using k8s secrets

PR #838, opened 2021-08-20
Allow rollbacks from versions with additional information in state_data

nemacysts:
On restarts, we attempt to recreate ActionRuns from state that we
persisted in DynamoDB (or locally) - this generally breaks if we
persisted fields that don't exist in one version and then ran another
version (e.g., when testing new features on our tronplayground instance)

### DIFF
--- a/tron/core/actionrun.py
+++ b/tron/core/actionrun.py
@@ -148,8 +148,19 @@ class ActionRunAttempt:
 
     @classmethod
     def from_state(cls, state_data):
-        state_data["command_config"] = action.ActionCommandConfig(**state_data["command_config"])
-        return cls(**state_data)
+        # it's possible that we've rolled back to an older Tron version that doesn't support data that we've persisted
+        # (e.g., new fields for an ActionCommandConfig) so ensure that we only load what we currently support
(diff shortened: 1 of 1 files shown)


tron/core/actionrun.py
```diff
@@ -148,8 +148,19 @@ def state_data(self):
 
     @classmethod
     def from_state(cls, state_data):
-        state_data["command_config"] = action.ActionCommandConfig(**state_data["command_config"])
-        return cls(**state_data)
+        # it's possible that we've rolled back to an older Tron version that doesn't support data that we've persisted
+        # (e.g., new fields for an ActionCommandConfig) so ensure that we only load what we currently support
+        valid_command_config_entries_from_state = {
+            field.name: state_data["command_config"][field.name]
+            for field in fields(action.ActionCommandConfig)
```
jfongatyelp, line 155: TIL fields()

jfongatyelp (approved):

tron/core/actionrun.py
nemacysts: yup, `dataclasses`-only, but it was nice :D (I was also surprised to see that we're already using it in this file too )

PR #841, opened 2021-08-31
Support passing Pod labels to task_processing

nemacysts:
We'll need this to set labels required by the PaaSTA Workload Contract
(which is needed to ensure that internal tooling will work as expected
with our Pods)


nemacysts:
this is gonna fail tests until https://github.com/Yelp/task_processing/pull/182 + https://github.com/Yelp/task_processing/pull/181 are merged and a new release is created - but I tested this locally with wheels I built myself

lanceleih (approved):

lanceleih (approved):

PR #842, opened 2021-09-02
Filter per-instance Tronjob logs by cluster

nemacysts:
It's possible that a tronjob in pnw-dev and norcal-devc could log to the
same stream (e.g., they have the same {namespace}.{job}.{action}
identifier and therefore will both log to
stream_paasta_app_output_{namespace}_{job}__{action}).

Currently, if this were to be the case and these jobs ran on the same
days with similar run numbers, we would mix the output from these two
runs when looking at the Tron UI for either of these.

Every task has a PAASTA_CLUSTER environment variable, so we can look at
the environment that we started a task with in order to filter the log
streams for these actions since these streams contain the cluster name
of where the logged message came from.

Therefore, we use this to ensure that we only display the correct log
messages to users instead of potentially mixing the output from jobs
with the same name that log to the same stream from different clusters.


lanceleih (approved):
LGTM

PR #851, opened 2021-11-20
Fix build process to support internal npm registry

jfongatyelp:
This fixes our internal build to use internal npm registry (similar to how we use internal pypi) for tronweb2 builds. 

Because yarn lockfiles explicitly pointed to a particular registry, this just deletes the lockfiles entirely and forces a rebuild based on our requirements in package.json.  While this may cause our builds of tronweb2 to change more often, this seems fine for now given we are not yet widely using tronweb2.

Internal builds with `make itest_xenial`, `make itest_bionic` are confirmed to work on our test instance with access only to internal registries.


Makefile
```diff
@@ -5,9 +5,11 @@ GID:=$(shell id -g)
 
 ifeq ($(findstring .yelpcorp.com,$(shell hostname -f)), .yelpcorp.com)
 	export PIP_INDEX_URL ?= https://pypi.yelpcorp.com/simple
+	export NPM_CONFIG_REGISTRY ?= https://npm.yelpcorp.com/
```
piax93, line 8: I've always seen this being set as lowercase `npm_config_registry`, but maybe it works regardless.

piax93 (approved):

Makefile
jfongatyelp: You are right, they prefer the lower case but both are supposed to work according to the [docs](https://docs.npmjs.com/cli/v8/using-npm/config)
> Config values are case-insensitive, so NPM_CONFIG_FOO=bar will work the same. However, please note that inside scripts npm will set its own environment variables and Node will prefer those lowercase versions over any uppercase ones that you might set. For details see this issue.

However we're actually just hard-coding this later for `YARN_REGISTRY` or passing it on the commandline so we don't leave it to chance.

nemacysts (approved):
I also wouldn't be opposed to deleting the tronweb2 code until we're ready to invest more time into actually making it usable

PR #852, opened 2021-12-01
Add Service Account support for k8s jobs

nemacysts:
We'll use this for Pod Identity - there's a webhook that will inject
a secret token + some environment variables if a Pod has a Service
Account set up.

paasta-tools will be in charge of creating the Service Account that
we'll use if it doesn't exist (as well as setting up the annotations on
that Service Account).

NOTE: tests will fail until https://github.com/Yelp/task_processing/pull/186 is merged/released, but I tested this locally with a manually built task_proc wheel (both automated tests and actually running a Pod)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -79,7 +79,7 @@ s3transfer==0.2.1
 setuptools==41.4.0
 six==1.14.0
 sshpubkeys==3.1.0
-task-processing==0.2.0
+task-processing==0.3.0
 traitlets==4.3.3
 Twisted==20.3.0
 typing-extensions==3.10.0.0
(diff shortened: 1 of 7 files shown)


jfongatyelp (approved):

danielpops (approved):
lgtm, no strong opinions

vkhromov (approved):

vkhromov (approved):

keymone (approved):
lgtm, side note - this is too much of a diff for adding an attribute :(

nemacysts:
> lgtm, side note - this is too much of a diff for adding an attribute :(

@keymone yea - figuring this out the first time I had to do it was negative fun too, thankfully now I have all those other diffs to use as reference now :p

PR #853, opened 2021-12-03
Allow users to use runid for triggers

nemacysts:
Some users have jobs whose output depends on the output of the previous
run - for critical jobs, it's useful to explicitly declare this
dependency so that these jobs never run unless Tron knows that the
previous output is ready/complete (using successful completion as a
proxy for this)

For additional context: https://yelp.slack.com/archives/CA4K8PBLG/p1638402768180500

### DIFF
--- a/tron/command_context.py
+++ b/tron/command_context.py
@@ -2,6 +2,7 @@
 have variables that need to be rendered.
 """
 import operator
+import re
 from functools import reduce
 
 from tron.utils import timeutils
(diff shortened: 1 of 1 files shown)


nemacysts:
I tested this with my local Tron instance using a job config like:
```YAML
jobs:
   kubernetes:
    actions:
      sleep:
        command: env; sleep 60; env
        trigger_downstreams:
          runid: "{runid}"
        triggered_by:
          - MASTER.kubernetes.sleep.runid.{runid-1}
```

the first job/action waited until I manually started it - and then each scheduled run waited until the previous one finished

That said: @keymone i'd like to see if you have any thoughts about how this was implemented - I'm not sure if I should have just only provided hardcoded support for `runid-1` since that's the only usecase we have/can think of for now (I'm also not quite sure from reading the code how long these triggers are held around for - i.e., did my test just work since we always have 1 future run scheduled but will break if someone tries to do something particularly fancy)

nemacysts:
also, I probably should have run the tests locally before creating this PR ;_;

nemacysts:
cc @diegoguemes

keymone (approved):
lgtm!

PR #854, opened 2021-12-06
Pass service_account_name in create_task() during recovery

nemacysts:
I missed this create_task callsite in the initial PR - without this,
we'll fail to recover state when Tron restarts.


nemacysts:
This time I made sure that I tested Tron restarts without blowing away all my local state as well ;_;

jfongatyelp (approved):

PR #857, opened 2022-01-11
Stop using unauthenticated git protocol for behave

nemacysts:
Github is currently browning out access to unauth'd git and will be
fully disabling it in the coming months.


nemacysts:
i'm merging this in as part of another PR

PR #858, opened 2022-01-18
Remove trusty and xenial traces

nemacysts:
We no longer need tron on xenial and trusty has been long gone

There's one last place that uses a xenial image, but it doesn't seem
like mesosphere packages mesos for bionic (and tron on mesos is almost
over) so i've left that alone


nemacysts:
make itest_bionic worked :)

example-cluster/images/xenial
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:xenial-20180123
+FROM ubuntu:bionic
```
jfongatyelp, line 1: Hmm perhaps this file should be renamed `s/xenial/bionic/` as well. or can we just remove this example image?

jfongatyelp (approved):

example-cluster/images/xenial
nemacysts: :lol: didn't even notice the name!

I think we can probably remove this and the rest of the example cluster stuff (especially since we don't use it) unless we rework it to work internally (after things like our pypi blocks), but I'll leave that for future luisp to do :)

PR #609, opened 2019-01-18
Bump pysensu-yelp to pull in JIRA priority support

jvperrin:
This pulls in the changes from https://github.com/Yelp/pysensu-yelp/pull/27 that are included in pysensu-yelp 0.4.1 ([released earlier today](https://pypi.org/project/pysensu-yelp/#history)).

PR #864, opened 2022-03-09
Bump pysensu-yelp to pull in issuetype support

nemacysts:
https://github.com/Yelp/pysensu-yelp/pull/30 was released in
pysensu-yelp==0.4.4 and allows users to set the issuetype for a ticket
(so that sensu-created tickets aren't always of type `Bug` (the default))


nemacysts:
This is pretty similar to https://github.com/Yelp/Tron/pull/609 and should just work (`make test` worked locally)

jfongatyelp (approved):

PR #866, opened 2022-03-17
Create KubernetesActionRuns for executor: spark

nemacysts:
For now, we should be able to just use a KubernetesActionRun for Spark
drivers - worst case, I think we'll just have a couple places where we
branch on `executor` if we absolutely need to do something
Spark-specific and can't just pass that down from paasta-tools

This PR also removes spark_driver_service_account_name - I'm not sure
what I was thinking initially - we don't really need anything to
distinguish between the driver and executor SA since the executor SA
is configured by the arguments used to start the driver (and thus we
can just start the driver pod as we normally would by using the service
account payload that we'd use for a normal tron-launched pod)

### DIFF
--- a/tron/kubernetes.py
+++ b/tron/kubernetes.py
@@ -225,6 +225,8 @@ class KubernetesCluster:
             return self.runner
 
         try:
+            # TODO(TRON-1701): we'll need to figure out a good way to support multiple clusters here
+            # (with each cluster only using a single namespace for tron purposes)
             executor = self.processor.executor_from_config(
                 provider="kubernetes",
                 provider_config={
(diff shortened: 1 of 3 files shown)


tron/kubernetes.py
```diff
@@ -225,6 +225,8 @@ def get_runner(self, kubeconfig_path: str, queue: PyDeferredQueue) -> Optional[S
             return self.runner
 
         try:
+            # TODO(TRON-1701): we'll need to figure out a good way to support multiple clusters here
```
nemacysts, line 228: i have some ideas on how to do this now, but while i'm still working on this code i can just do the "dumb" thing and muck with tron code locally to override the config here to what i actually need for spark

jfongatyelp (approved):

PR #869, opened 2022-04-07
Pass port/field selector envvars to task_proc

nemacysts:
we need this for spark drivers in k8s to work - and without the field selector env vars we're not able to fully adhere to the paasta workload contract

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -79,7 +79,7 @@ s3transfer==0.2.1
 setuptools==41.4.0
 six==1.14.0
 sshpubkeys==3.1.0
-task-processing==0.3.0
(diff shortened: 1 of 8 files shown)


requirements.txt
```diff
@@ -79,7 +79,7 @@ s3transfer==0.2.1
 setuptools==41.4.0
 six==1.14.0
 sshpubkeys==3.1.0
-task-processing==0.3.0
+task-processing==0.4.0
```
nemacysts, line 82: https://github.com/Yelp/task_processing/pull/189

jfongatyelp (approved):

PR #871, opened 2022-06-14
Fix internal build

jfongatyelp:
Our internal build now runs inside a container whose `hostname` no longer matches `yelpcorp.com`.

This instead changes our makefile to set internal-only things based on `PAASTA_ENV`, and a separate change will force our internal build to set `PAASTA_ENV=YELP` when building. 

In addition to the internal build PIP problem, a dependency for one of our node eslint checks was updated to a version that was no longer compatible w/ node <12.0.  To minimize potential breakage on tronweb, I am just pinning to an older version of axe-core for now, and will ticket updating to node 12+ separately.

This got past `make test` on an internal build on k8s: https://jenkins-test.yelpcorp.com/blue/organizations/jenkins/test-jfong-tron/detail/test-jfong-tron/6/pipeline/ e.g. with `15:06:08  Now running on hostname: test-jfong-tron-6-bgkhg-z59c9-8qbsv`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,19 +4,19 @@ UID:=$(shell id -u)
 GID:=$(shell id -g)
 
 ifeq ($(findstring .yelpcorp.com,$(shell hostname -f)), .yelpcorp.com)
-	export PIP_INDEX_URL ?= https://pypi.yelpcorp.com/simple
-	export NPM_CONFIG_REGISTRY ?= https://npm.yelpcorp.com/
 	PAASTA_ENV ?= YELP
 else
-	export PIP_INDEX_URL ?= https://pypi.python.org/simple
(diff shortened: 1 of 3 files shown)


nemacysts (approved):

EvanKrall:
Does this mean anyone at Yelp working on Tron would need to pass PAASTA_ENV=YELP to any `make` call?

jfongatyelp:
> 
> Does this mean anyone at Yelp working on Tron would need to pass PAASTA_ENV=YELP to any `make` call?

Nope, because if PAASTA_ENV is not defined, the hostname you're on should match `*.yelpcorp.com` and then `PAASTA_ENV ?= YELP` will set it.
Verified with a local `make test` on my devbox.

EvanKrall:
Ah, I see, I misread the code. Yeah, that makes sense.

PR #873, opened 2022-08-12
U/emanelsabban/tron UI 1737

EmanElsaban:
This PR limits the amount of log output in Tron UI by truncating the output after exceeding 1000 lines.


tron/utils/scribereader.py
```diff
@@ -77,7 +77,7 @@ def read_log_stream_for_action_run(
     min_date: Optional[datetime.datetime],
     max_date: Optional[datetime.datetime],
     paasta_cluster: Optional[str],
-    max_lines: Optional[int] = None,
```
nemacysts, line 80: it might be a good idea to find out what the average log length is so that we can pick a number that hopefully won't impact folks + make sure that we're in general keeping that number in mind as a target for what we should aim to support at a minimum

we can probably figure this out through science in a couple ways:
* check the log streams for all the jobs that are using k8s in other clusters and see what the average lines per run is by summing things up
* look at the logs on-disk on the last mesos-using cluster and see how big those files are (although, the functionality here is pretty brittle so a lot of these might be empty ;_;)
* see what the thresholds we used for cleaning/truncating the on-disk logs are and extrapolate from that to how many lines we'd like to keep

it might also be a good idea to have this default be configurable without having to re-deploy tron: perhaps we can use srv-configs here (we'll have to use slightly  different tooling than we do internally - the srv-configs library is not open-source so we'll need to use staticconf directly)?

tron/utils/scribereader.py
```diff
@@ -183,6 +186,15 @@ def read_log_stream_for_action_run(
         finally:
             stream.close()
 
+    if truncated_output:
+        formated_today = datetime.datetime.now(datetime.timezone.utc).strftime("%Y-%m-%dT%H:%M:%S.%fZ")
+        output.append(
+            (
+                formated_today,
+                f"This output is truncated. Use this command to view all lines 'scribereader -s {paasta_cluster} -f stream_paasta_app_output_{namespace}_{job_name}__{action}'",
```
nemacysts, line 194:
```suggestion
                f"This output is truncated. Use this command to view all lines 'scribereader -s {paasta_cluster} -f {stream_name}'",
```

tron/utils/scribereader.py
```diff
@@ -183,6 +186,15 @@ def read_log_stream_for_action_run(
         finally:
             stream.close()
 
+    if truncated_output:
+        formated_today = datetime.datetime.now(datetime.timezone.utc).strftime("%Y-%m-%dT%H:%M:%S.%fZ")
```
nemacysts, line 190: we could probably skip having to create a datetime object if we just add this into the very end of the list that we return (in the same way that we add in the malformed lines output)

tron/utils/scribereader.py
nemacysts: i *think* we might also need to maybe add `--min-date` and `--max-date` here so that this line is accurate even if someone is looking at a log a week later?

tests/utils/scribereader_test.py
```diff
@@ -197,3 +197,47 @@ def test_read_log_stream_for_action_run_min_date_and_max_date_in_past():
...
+        )
+    mock_stream_reader.assert_called_once_with(
+        stream_name="stream_paasta_app_output_namespace_job__action",
+        min_date=min_date,
+        max_date=max_date,
+        reader_host="host",
+        reader_port=1234,
+    )
+    mock_stream_tailer.assert_not_called()
+    # print(output[-1].find("truncated. Use this command to view all lines"))
```
nemacysts, line 242: looks like a debug print was left in :)
```suggestion
```

tests/utils/scribereader_test.py
```diff
@@ -197,3 +197,47 @@ def test_read_log_stream_for_action_run_min_date_and_max_date_in_past():
...
+    assert output[-1].find("truncated. Use this command to view all lines") > 0
```
nemacysts, line 243: maybe we just assert that the length of the output is equal to the limit (+ 1 i guess :p) so that we don't need to change this test if we change the message?

nemacysts (commented):
nice! this is looking great - i just left a couple comments to make this even more bulletproof

tron/utils/scribereader.py
cuza: we sort the list by timestamp before returning it because we're occasionally getting data out of order from scribereader

nemacysts: correct - but we know we want this at the very end, so we can just insert at the end

EmanElsaban:
### Testing
I tested this PR in two ways: 

1. Using Tron UI
2. Writing unit tests
-------------------------
1. Using Tron UI:
**results:** Below, the output is truncated after 1000 lines
http://tronplayground1-uswest1cdevc.uswest1-devc.yelpcorp.com:8089/web/#job/compute-infra-test-service.test_output/1803/foo

2. Using unit testing
**results:**  Tests passing
```
(py36) emanelsabban@dev55-uswest1adevc:~/pg/Tron$ py.test -s tests/utils/scribereader_test.py
============================================================================================================ test session starts =============================================================================================================
platform linux -- Python 3.6.9, pytest-6.2.2, py-1.10.0, pluggy-0.13.0
rootdir: /nail/home/emanelsabban/pg/Tron
plugins: asyncio-0.14.0
collected 4 items                                                                                                                                                                                                                            

tests/utils/scribereader_test.py ....

============================================================================================================== warnings summary ==============================================================================================================
.tox/py36/lib/python3.6/site-packages/boto/plugin.py:40
  /nail/home/emanelsabban/pg/Tron/.tox/py36/lib/python3.6/site-packages/boto/plugin.py:40: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
    import imp

-- Docs: https://docs.pytest.org/en/stable/warnings.html
======================================================================================================== 4 passed, 1 warning in 0.17s ========================================================================================================
(py36) emanelsabban@dev55-uswest1adevc:~/pg/Tron$ 
```
```
(py36) emanelsabban@dev55-uswest1adevc:~/pg/Tron$ py.test -s tests/api/resource_test.py
============================================================================================================ test session starts =============================================================================================================
platform linux -- Python 3.6.9, pytest-6.2.2, py-1.10.0, pluggy-0.13.0
rootdir: /nail/home/emanelsabban/pg/Tron
plugins: asyncio-0.14.0
collected 30 items                                                                                                                                                                                                                           

tests/api/resource_test.py ..............................

============================================================================================================== warnings summary ==============================================================================================================
testifycompat/fixtures.py:42
  /nail/home/emanelsabban/pg/Tron/testifycompat/fixtures.py:42: PytestDeprecationWarning: @pytest.yield_fixture is deprecated.
  Use @pytest.fixture instead; they are the same.
    return pytest.yield_fixture(autouse=True, scope="class")(teardown_)

testifycompat/fixtures.py:26
  /nail/home/emanelsabban/pg/Tron/testifycompat/fixtures.py:26: PytestDeprecationWarning: @pytest.yield_fixture is deprecated.
  Use @pytest.fixture instead; they are the same.
    return pytest.yield_fixture(autouse=True)(teardown_)

.tox/py36/lib/python3.6/site-packages/boto/plugin.py:40
  /nail/home/emanelsabban/pg/Tron/.tox/py36/lib/python3.6/site-packages/boto/plugin.py:40: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
    import imp

-- Docs: https://docs.pytest.org/en/stable/warnings.html
======================================================================================================= 30 passed, 3 warnings in 2.19s =======================================================================================================
(py36) emanelsabban@dev55-uswest1adevc:~/pg/Tron$ 
```

tron/utils/scribereader.py
```diff
@@ -186,6 +199,14 @@ def read_log_stream_for_action_run(
...
     # timestamp until we can figure out what's causing this.
     output.sort(key=operator.itemgetter(0))
-    return [line for _, line in output] + (
-        [f"{malformed_lines} encountered while retrieving logs"] if malformed_lines else []
+    return (
+        [line for _, line in output]
+        + ([f"{malformed_lines} encountered while retrieving logs"] if malformed_lines else [])
+        + (
+            [
+                f"This output is truncated. Use this command to view all lines 'scribereader -s {paasta_cluster} -f stream_paasta_app_output_{namespace}_{job_name}__{action} --min-date {min_date} --max-date {max_date}'"
```
nemacysts, line 207: nit:
```suggestion
                f"This output is truncated. Use this command to view all lines 'scribereader -s {paasta_cluster} {stream_name} --min-date {min_date} --max-date {max_date}'"
```

also: max_date can be None, so we'll need to account for that and only include that arg if it's present - i've also noticed that i might have messed up when i was initially writing this code: can you verify that min_date and max_date here are printed out in YYYY-MM-DD format and not a full datetime string?

tron/utils/scribereader.py
```diff
@@ -186,6 +199,14 @@ def read_log_stream_for_action_run(
...
+            ]
+            if truncated_output
+            else []
+        )
     )
```
nemacysts, line 212: this is probably long enough now that we wanna clean this up and do something like: 
```suggestion
    return (
        lines = [line for _, line in output]
        malformed = [f"{malformed_lines} encountered while retrieving logs"] if malformed_lines else []
        
        truncation_message = f"This output is truncated. Use this command to view all lines 'scribereader -s {paasta_cluster} -f stream_paasta_app_output_{namespace}_{job_name}__{action} --min-date {min_date} --max-date {max_date}'"
         truncated = [truncation_message] if truncated_output else []
         
         return lines + malformed + truncated
    )
```

rather than one-lining the whole thing

tron/utils/scribereader.py
EmanElsaban: I think it is datetime string. Below is how it's outputted:

```
"This output is truncated. Use this command to view all lines 'scribereader -s infrastage stream_paasta_app_output_namespace_job__action --min-date 2022-08-13 15:50:10.623651 --max-date 2022-08-14 15:50:10.623656'"]
```

nemacysts: ah, we'll probably wanna call `.date()` on those to get just the YYYY-MM-DD bits

tron/config/static_config.py
```diff
@@ -0,0 +1,23 @@
...
+
+FILENAME = "/nail/srv/configs/tron.yaml"
+NAMESPACE = "tron"
+
+
+def load_yaml_file():
+    staticconf.YamlConfiguration(filename=FILENAME, namespace=NAMESPACE)
+
+
+def build_configuration(filename, namespace):
```
nemacysts, line 14:
```suggestion
def build_configuration(filename: str, namespace: str) -> config.ConfigurationWatcher:
```
 
(could also maybe name this build_configuration_watcher(), but that's pretty nitpicky :p)

tron/config/static_config.py
```diff
@@ -0,0 +1,23 @@
...
+    config_loader = partial(staticconf.YamlConfiguration, filename, namespace=namespace)
+    reloader = config.ReloadCallbackChain(namespace)
+    return config.ConfigurationWatcher(config_loader, filename, min_interval=0, reloader=reloader)
+
+
+# Load configuration from 'tron.yaml' into namespace 'tron'
+def get_config_watcher():
```
nemacysts, line 21:
```suggestion
def get_config_watcher() -> config.ConfigurationWatcher:
```

tron/config/static_config.py
```diff
@@ -0,0 +1,23 @@
+from functools import partial
+
+import staticconf
+from staticconf import config
+
+FILENAME = "/nail/srv/configs/tron.yaml"
+NAMESPACE = "tron"
+
+
+def load_yaml_file():
```
nemacysts, line 10:
```suggestion
def load_yaml_file() -> staticconf.YamlConfiguration:
```

nemacysts (commented):
whoops, forgot to leave some comments about adding type hints

tron/config/static_config.py
EmanElsaban: `` load_yaml_file()`` doesn't have a return value. Should I still add a type hint to the return value of this function? 
or 
```
def load_yaml_file() -> None
```

nemacysts: whoops, my brain made up a return: you're right, this should be `-> None`

tron/utils/scribereader.py
```diff
@@ -77,11 +82,16 @@ def read_log_stream_for_action_run(
     min_date: Optional[datetime.datetime],
     max_date: Optional[datetime.datetime],
     paasta_cluster: Optional[str],
-    max_lines: Optional[int] = None,
+    max_lines: Optional[int] = -1,
```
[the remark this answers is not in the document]
EmanElsaban: yep, it makes sense

cuza (approved):

PR #875, opened 2022-09-01
This adds rookout to tron (TRON-1764)

bobokun:
Allows you to enable rookout for Tron using config files. (TRON-1764)


tron/utils/rookout.py
```diff
@@ -0,0 +1,92 @@
...
+import os
+import platform
+from typing import Optional
+
+import rook
+import staticconf  # type: ignore
+
+import tron
+from tron.config.static_config import get_config_watcher
+from tron.config.static_config import NAMESPACE
```
nemacysts, line 11: the public tests/itests are probably going to be broken 'cause srv-configs aren't available on GHA

we could figure out a good way to mock these in both - but maybe we should move the rookout config out of here and either into tron's soaconfigs or just environment variables that we set in puppet in the systemd unit file. 

srv-configs really shines for things that can change without a "reboot", but this rookout code will only ever be called at startup so we don't necessarily need to deal with getting staticconf to work with our public tests if we move this config out of there :p

tron/utils/rookout.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def prepare_rookout_token() -> None:
+    """Load rookout token into memory while we are still root"""
+    global ROOKOUT_TOKEN
+    try:
+        with open(ROOKOUT_TOKEN_PATH, encoding="utf-8") as _rookout_token_file:
+            ROOKOUT_TOKEN = _rookout_token_file.read().strip()
+    except OSError as exc:
+        log.warning("Failed to load rookout token: %s", exc)
```
nemacysts, line 46:
```suggestion
        log.exception("Failed to load rookout token")
```

tron/utils/rookout.py
```diff
@@ -0,0 +1,92 @@
...
+log = logging.getLogger(__name__)
+
+
+def get_version_sha() -> Optional[str]:
+    version_sha = None
+    try:
+        with open(f"{os.path.dirname(os.path.abspath(__file__))}/../VERSION_SHA") as file:
+            version_sha = file.read().rstrip()
+    except OSError as exc:
+        log.warning("Failed to read the file VERSION_SHA: %s. Loading master as default.", exc)
```
nemacysts, line 31:
```suggestion
        log.exception("Failed to read the file VERSION_SHA. Using master as default.")
```

tron/utils/rookout.py
```diff
@@ -0,0 +1,92 @@
...
+        log.info("Rookout token not loaded, not enabling SDK")
+        return
+
+    connection_config = {
+        "host": staticconf.read("rookout.controller", namespace=NAMESPACE, default=DEFAULT_ROOKOUT_HOST,),
+        "port": staticconf.read_int("rookout.controller_port", namespace=NAMESPACE, default=DEFAULT_ROOKOUT_PORT,),
+    }
+
+    if connection_config != ROOKOUT_CONFIG:
+        log.info("Stopping Rookout SDK (reconfig)")
```
nemacysts, line 73: will this ever actually be called for a normal tron reconfigure?

tron/utils/rookout.py
```diff
@@ -0,0 +1,92 @@
...
+            token=ROOKOUT_TOKEN,
+            labels={
+                "app": "tron",
+                "hostname": get_hostname(),
+                "ecosystem": open("/nail/etc/ecosystem", encoding="utf-8").read(),
+                "superregion": open("/nail/etc/superregion", encoding="utf-8").read(),
+                "version": tron.__version__,
+            },
+            log_to_stderr=True,
+            git_commit=get_version_sha(),
```
nemacysts, line 86: i wonder if we can point this at a tag and reuse `tron.__version__` so that we can skip the VERSION_SHA fun :p

tron/utils/rookout.py
```diff
@@ -0,0 +1,92 @@
...
+        log.warning("Failed to read the file VERSION_SHA: %s. Loading master as default.", exc)
+    return version_sha
+
+
+def get_hostname() -> str:
+    return platform.node()
+
+
+def prepare_rookout_token() -> None:
+    """Load rookout token into memory while we are still root"""
```
nemacysts, line 40: trond should always be running as the `tron` user afaik

tron/trondaemon.py
```diff
@@ -85,7 +85,7 @@ def run(self):
                 signal.SIGUSR1: self._handle_debug,
             }
             signal.pthread_sigmask(signal.SIG_BLOCK, signal_map.keys())
-
+            self._run_rookout()
```
nemacysts, line 88: it'd be nice to wrap this with a try-except just to ensure that any issues with rookout/rookout setup don't stop tron from starting

tron/utils/rookout.py
bobokun: I tried doing that and rookout unfortunately does not like that. 👎

tron/utils/rookout.py
bobokun: I've modified the tests to mock the functions

tests/rookout_test.py
```diff
@@ -0,0 +1,82 @@
...
+
+@pytest.fixture()
+def setup_config():
+    tron_config = {"rookout.enable": True}
+    yield staticconf.testing.MockConfiguration(tron_config, namespace=NAMESPACE)
+
+
+@patch("rook.start")
+@patch("rook.stop")
+@patch("tron.utils.rookout.get_config_watcher", autospec=True)
```
nemacysts, line 19: hare we able to use the staticconf.testing utils instead of directly mocking?

tron/utils/rookout.py
bobokun: Okay, I understand what you mean. What do you think is the better option?
Option 1:
1. Add the following `ROOKOUT_TOKEN_PATH, ROOKOUT_HOST, ROOKOUT_PORT, ROOKOUT_ENABLE` in tron's soaconfigs.
2. Modifying tron's api to trigger rookout reload whenever any of these variables change.

Option2:
1. Add the following `ROOKOUT_TOKEN_PATH, ROOKOUT_HOST, ROOKOUT_PORT, ROOKOUT_ENABLE`  as env variables in puppet so when these get modified it'll just restart tron with the modified variables.

Option 2 would definitely be the easier approach. and I don't think these variables would be changing often at all.

nemacysts: option 2 sounds best imo - especially since the rookout docs make it seem like the only thing that rookout supports reloading for is updated labels

tron/utils/rookout.py
```diff
@@ -0,0 +1,77 @@
...
+    global ROOKOUT_TOKEN
+    try:
+        with open(ROOKOUT_TOKEN_PATH, encoding="utf-8") as _rookout_token_file:
+            ROOKOUT_TOKEN = _rookout_token_file.read().strip()
+    except OSError:
+        log.exception("Failed to load rookout token")
+
+
+def enable_rookout() -> None:
+    """Enable rookout if srv_config enables it"""
```
nemacysts, line 45: nit:
```suggestion
    """Enable rookout if configured"""
```

tron/utils/rookout.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+def prepare_rookout_token() -> None:
+    """Load rookout token into memory"""
+    global ROOKOUT_TOKEN
+    try:
+        with open(ROOKOUT_TOKEN_PATH, encoding="utf-8") as _rookout_token_file:
+            ROOKOUT_TOKEN = _rookout_token_file.read().strip()
+    except OSError:
+        log.exception("Failed to load rookout token")
```
nemacysts, line 41: hmm, thoughts on having this function return the token (or None)? 

if we do that, then we don't need to use `global` anywhere (which is usually something to be avoided if possible)

tests/rookout_test.py
```diff
@@ -0,0 +1,23 @@
...
+    rook_start.return_value = True
+    enable_rookout()
+    assert rook_start.call_count == 0
+
+
+@patch("rook.start")
+@patch("tron.utils.rookout.ROOKOUT_ENABLE", True)
+@patch("tron.utils.rookout.ROOKOUT_TOKEN", "not-a-real-token")
+def test_enable_rookout(rook_start: MagicMock):
+    """Test with configured staticconf"""
```
nemacysts, line 20: nit: we're not using staticconf for this anymore
```suggestion
    rook_start.return_value = True
    enable_rookout()
    assert rook_start.call_count == 0


@patch("rook.start")
@patch("tron.utils.rookout.ROOKOUT_ENABLE", True)
@patch("tron.utils.rookout.ROOKOUT_TOKEN", "not-a-real-token")
def test_enable_rookout(rook_start: MagicMock):
```

nemacysts (commented):
nice! just left some minor nits/comments - nothing major tho :)

tron/utils/rookout.py
bobokun: oops, I knew I missed one of these coments 😅

tron/utils/rookout.py
bobokun: yup I was thinking to do that as well to avoid global. Good call  👍

tests/rookout_test.py
bobokun: good catch

nemacysts (approved):

nemacysts:
looks like there's a failing test: `FAILED tests/rookout_test.py::test_enable_rookout - FileNotFoundError: [Errno...`

bobokun:
> looks like there's a failing test: `FAILED tests/rookout_test.py::test_enable_rookout - FileNotFoundError: [Errno...`
Hmm...I'm mocking `rook.start` though, so it shouldn't be even entering that function. Do I need to mock `get_ecosystem()` and `get_superregion()` as well?

nemacysts (approved):
might be worth coming back to this in the future and updating the tests to use the context-manager form of mocking, but there's enough tests that look like this already that this is probably fine as-is :)

tron/utils/scribereader.py
```diff
@@ -25,45 +26,12 @@
...
-    Discover what region we're running in by reading this information from on-disk facts.
-
-    Yelpers: for more information, see y/habitat
-    """
-    with open("/nail/etc/region") as f:
-        return f.read().strip()
-
-
-@lru_cache(maxsize=1)
-def get_superregion() -> str:
```
tzhu-yelp, line 40: This mock.patch was not updated: https://github.com/Yelp/Tron/blob/4da0f48/tests/utils/scribereader_test.py#L27

PR #884, opened 2022-11-21
Detect abnormal successful exits in k8s

nemacysts:
this is kinda wild: we're seeing that a kubelet will sometimes fail to start a container (usually due to what appear to be race conditions like those mentioned in https://github.com/kubernetes/kubernetes/issues/100047#issuecomment-797624208 and then decide that these Pods should be phase=Succeeded with an exit code of 0 - even though the container never actually started.

So far, we've noticed that when this happens, the `finished_at` and `reason` fields will be `None` - thus we'll check for at least one of these conditions to detect an abnormal exit and actually "fail" the affected action

### DIFF
--- a/tron/kubernetes.py
+++ b/tron/kubernetes.py
@@ -135,8 +135,42 @@ class KubernetesTask(ActionCommand):
         if k8s_type == "running":
             self.started()
         elif k8s_type == "finished":
-            self.exited(0)
+            raw_object = getattr(event, "raw", {})
+            container_statuses = raw_object.get("status", {}).get("container_statuses", [])
+            abnormal_exit = False
+
+            if len(container_statuses) > 1:
+                # shouldn't happen right now, but who knows what future us will do :p
+                self.log.error(
+                    "Got an event for a Pod with multiple containers - not inspecting payload to verify success."
+                )
+            else:
+                main_container_statuses = container_statuses[0]
+                main_container_state = main_container_statuses.get("state")
+                if main_container_state is None or main_container_state.get("terminated") in None:
+                    self.log.error("Got an event with missing state - assuming success.")
+                else:
+                    termination_metadata = main_container_state.get("terminated")
+                    # this is kinda wild: we're seeing that a kubelet will sometimes fail to start a container (usually
+                    # due to what appear to be race conditons like those mentioned in
+                    # https://github.com/kubernetes/kubernetes/issues/100047#issuecomment-797624208) and then decide that
+                    # these Pods should be phase=Succeeded with an exit code of 0 - even though the container never actually
+                    # started. So far, we've noticed that when this happens, the finished_at and reason fields will be None
+                    # and thus we'll check for at least one of these conditions to detect an abnormal exit and actually "fail"
+                    # the affected action
+                    # NOTE: hopefully this won't
+                    abnormal_exit = (
+                        termination_metadata.get("exit_code") == 0 and termination_metadata.get("finished_at") is None
+                    )
+                    if abnormal_exit:
+                        self.log.warning("Container never started due to a Kubernetes/infra flake!")
+                        self.log.warning(
+                            f"If automatic retries are not enabled, run `tronctl retry {self.id}` to retry."
+                        )
+
+            self.exited(int(abnormal_exit))
         elif k8s_type == "failed":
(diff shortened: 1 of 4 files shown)


tron/kubernetes.py
```diff
@@ -135,8 +135,42 @@ def handle_event(self, event: Event) -> None:
...
         elif k8s_type == "finished":
-            self.exited(0)
+            raw_object = getattr(event, "raw", {})
+            container_statuses = raw_object.get("status", {}).get("container_statuses", [])
+            abnormal_exit = False
+
+            if len(container_statuses) > 1:
+                # shouldn't happen right now, but who knows what future us will do :p
+                self.log.error(
+                    "Got an event for a Pod with multiple containers - not inspecting payload to verify success."
```
bobokun, line 145: can we log the actual contents of the container_status so we can investigate this as well?

tron/kubernetes.py
```diff
@@ -135,8 +135,42 @@ def handle_event(self, event: Event) -> None:
...
+                    # this is kinda wild: we're seeing that a kubelet will sometimes fail to start a container (usually
+                    # due to what appear to be race conditons like those mentioned in
+                    # https://github.com/kubernetes/kubernetes/issues/100047#issuecomment-797624208) and then decide that
+                    # these Pods should be phase=Succeeded with an exit code of 0 - even though the container never actually
+                    # started. So far, we've noticed that when this happens, the finished_at and reason fields will be None
+                    # and thus we'll check for at least one of these conditions to detect an abnormal exit and actually "fail"
+                    # the affected action
+                    # NOTE: hopefully this won't
+                    abnormal_exit = (
+                        termination_metadata.get("exit_code") == 0 and termination_metadata.get("finished_at") is None
```
bobokun, line 163: maybe we can also check for reason as well? Something like
`termination_metadata.get("exit_code") == 0 and (termination_metadata.get("finished_at") is None or termination_metadata.get("reason") is None)`

bobokun (commented):
Left some suggestions

tron/kubernetes.py
```diff
@@ -135,8 +135,42 @@ def handle_event(self, event: Event) -> None:
...
+                else:
+                    termination_metadata = main_container_state.get("terminated")
+                    # this is kinda wild: we're seeing that a kubelet will sometimes fail to start a container (usually
+                    # due to what appear to be race conditons like those mentioned in
+                    # https://github.com/kubernetes/kubernetes/issues/100047#issuecomment-797624208) and then decide that
+                    # these Pods should be phase=Succeeded with an exit code of 0 - even though the container never actually
+                    # started. So far, we've noticed that when this happens, the finished_at and reason fields will be None
+                    # and thus we'll check for at least one of these conditions to detect an abnormal exit and actually "fail"
+                    # the affected action
+                    # NOTE: hopefully this won't
```
jfongatyelp, line 161: Looks like an unfinished thought here :p

tron/kubernetes.py
```diff
@@ -135,8 +135,42 @@ def handle_event(self, event: Event) -> None:
         if k8s_type == "running":
             self.started()
         elif k8s_type == "finished":
-            self.exited(0)
+            raw_object = getattr(event, "raw", {})
+            container_statuses = raw_object.get("status", {}).get("container_statuses", [])
```
jfongatyelp, line 139: Do we know if k8s will ever give us a `null` but existent`container_statuses` (i.e. the `get()` will give us `None` which will make our `len()` fail later)?  I think I've seen funky stuff like 
```
container_statuses = pod_status.get('status', {}).get('container_statuses')
container_statuses = container_statuses if container_statuses else []
```
in our code that means to save ourselves from those weird edge cases, but not sure it's necessary here.

jfongatyelp (commented):
++ to the later TODO code hint to make sure ooms get tracked correctly again & the user friendly retry message!

tron/kubernetes.py
nemacysts: I think we'd want `and` instead of `or` if we wanna match what we've been seeing exactly

tron/kubernetes.py
nemacysts: whoops :p

jfongatyelp (approved):

bobokun (approved):

bobokun (approved):

PR #886, opened 2022-11-29
TRON-1825: Add additional failure handling of specific errors

bobokun:
Adds handling of spot interruptions and k8s scaling down in Tron to make it easier for users to know what caused the Kubernetes failure.


tron/core/actionrun.py
```diff
@@ -214,6 +214,8 @@ class ActionRun(Observable):
     EXIT_KUBERNETES_NOT_CONFIGURED = -7
     EXIT_KUBERNETES_TASK_INVALID = -8
     EXIT_KUBERNETES_ABNORMAL = -9
+    EXIT_KUBERNETES_SPOT_INTERRUPTION = 137
+    EXIT_KUBERNETES_NODE_SCALEDOWN = 143
```
jfongatyelp, line 218: historically 137s indicated OOMs and 143s were other external kills e.g. via docker or via tronctl kill (see https://yelpwiki.yelpcorp.com/questions/111183238/answers/111183255)

I don't think we should override these expectations. 

It might make more sense to treat these special cases with negative exit codes as with the other system-level errors like KUBERNETES_NOT_CONFIGURED.

bobokun: The only reason why I used those numbers is so that it matches with the error codes from the events we were getting from Kubernetes. I can change it to negative error codes but then it wouldn't be the same code as the events we are getting from K8s.

nemacysts: i'm not really sure why k8s has decided to co-opt `137` as an exit code for anything other than OOMKills :/

that said, i think it's perfectly fine to not have the tron exit code here match the k8s exit code (especially when the k8s exit code would be actively misleading) :)

tron/kubernetes.py
```diff
@@ -148,35 +148,58 @@ def handle_event(self, event: Event) -> None:
                 self.log.error(f"Event with >1 || 0 containers: {raw_object}")
             else:
                 main_container_statuses = container_statuses[0]
-                main_container_state = main_container_statuses.get("state")
-                if main_container_state is None or main_container_state.get("terminated") is None:
+                main_container_state = main_container_statuses.get("state", {}) or {}
+                main_container_last_state = main_container_statuses.get("lastState", {}) or {}
+                if (main_container_state is None or main_container_state.get("terminated") is None) and (
+                    main_container_last_state is None or main_container_last_state.get("terminated") is None
+                ):
```
nemacysts, line 155: it might aid readability here if we refactor both sides of the `and` to separate variables

e.g.,
```suggestion
                event_missing_state = main_container_state is None or main_container_state.get("terminated") is None
                event_missing_previous_state =  main_container_last_state is None or main_container_last_state.get("terminated") is None
                if event_missing_state and event_missing_previous_state:
```

that said: is this correct? seems like if one of these is fully populated and the other isn't, we'd continue onwards with execution

tron/kubernetes.py
```diff
@@ -148,35 +148,58 @@ def handle_event(self, event: Event) -> None:
...
+                            state_termination_metadata.get("finishedAt") is None
+                            and state_termination_metadata.get("reason") is None
+                        ):
+                            # -9 is EXIT_KUBERNETES_ABNORMAL from actionrun.py - we just can't import that value right now since there'd be a circular dependency
+                            exit_code = -9
+                            self.log.warning("Container never started due to a Kubernetes/infra flake!")
+                            self.log.warning(
+                                f"If automatic retries are not enabled, run `tronctl retry {self.id}` to retry."
+                            )
+                    elif k8s_type == "failed" or k8s_type == "killed":
```
nemacysts, line 181: nit: 
```suggestion
                    elif k8s_type in {"failed", "killed"}:
```
(maybe also extract this to a named constant?)

tron/kubernetes.py
```diff
@@ -148,35 +148,58 @@ def handle_event(self, event: Event) -> None:
...
+                        ):
+                            exit_code = -10
+                            self.log.warning("Tronjob failed due to spot interruption.")
+                        # Handling K8s scaling down a node
+                        elif state_termination_metadata.get("exitCode", 1) == 143 and (
+                            state_termination_metadata.get("reason", "") == "Error"
+                        ):
+                            exit_code = -11
+                            self.log.warning("Tronjob failed due to Kubernetes scaling down a node.")
+                        if exit_code in [-10, -11]:
```
nemacysts, line 195: `[-10, -11]` seems like a good candidate for a named constant (and we could even make it a set) :)

tron/kubernetes.py
```diff
@@ -148,35 +148,58 @@ def handle_event(self, event: Event) -> None:
...
+                            self.log.warning(
+                                f"If automatic retries are not enabled, run `tronctl retry {self.id}` to retry."
+                            )
+                            self.log.warning(
+                                "If this job is idempotent, then please consider increasing the number of retries to your action. If your job is not idempotent then please set this job to run on the stable pool rather than the default."
```
nemacysts, line 200: i'm a little worried about folks seeing this message and raising the retry count significantly when their currently configured # of retries is fine (e.g., they got unlucky once with a spot termination, but then the automatic retry worked)

should we reword this to `If this action is idempotent, then please consider enabling automatic retries for your action. If your action is not idempotent, then please configure this action to run on the stable pool rather than the default.`

tron/kubernetes.py
```diff
@@ -148,35 +148,58 @@ def handle_event(self, event: Event) -> None:
...
-                main_container_state = main_container_statuses.get("state")
-                if main_container_state is None or main_container_state.get("terminated") is None:
+                main_container_state = main_container_statuses.get("state", {}) or {}
+                main_container_last_state = main_container_statuses.get("lastState", {}) or {}
+                if (main_container_state is None or main_container_state.get("terminated") is None) and (
+                    main_container_last_state is None or main_container_last_state.get("terminated") is None
+                ):
                     self.log.error("Got an event with missing state - assuming success.")
                     self.log.error(f"Event with missing state: {raw_object}")
                 else:
```
nemacysts, line 158: hmm, i wonder if we should extract some of this logic into a function (or functions) - the initial abnormal exit code was already a bit pushing it in terms of inlining code :p

nemacysts (commented):
nice! once this is ready this should help a ton :)

tron/kubernetes.py
```diff
@@ -134,11 +134,11 @@ def handle_event(self, event: Event) -> None:
 
         if k8s_type == "running":
             self.started()
-        elif k8s_type == "finished":
+        elif k8s_type == "finished" or k8s_type == "failed" or k8s_type == "killed":
             raw_object = getattr(event, "raw", {}) or {}
             pod_status = raw_object.get("status", {}) or {}
             container_statuses = pod_status.get("containerStatuses", []) or []
-            abnormal_exit = False
+            exit_code = 0 if k8s_type == "finished" else 1
```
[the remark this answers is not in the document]
bobokun: do you mean grabbing whatever exit code is coming from k8s under state/last_state and using that rather than defaulting it to 1?

tron/kubernetes.py
bobokun: I originally had it as `or` instead of `and` but it was giving undesirable results. I forget why I changed it..

tron/kubernetes.py
```diff
@@ -148,35 +148,58 @@ def handle_event(self, event: Event) -> None:
...
+                            exit_code = -9
+                            self.log.warning("Container never started due to a Kubernetes/infra flake!")
+                            self.log.warning(
+                                f"If automatic retries are not enabled, run `tronctl retry {self.id}` to retry."
+                            )
+                    elif k8s_type == "failed" or k8s_type == "killed":
+                        # Handling spot terminations
+                        if last_state_termination_metadata.get("exitCode", 1) == 137 and (
+                            "deleted" in last_state_termination_metadata.get("message", "")
+                            and last_state_termination_metadata.get("reason", "") == "ContainerStatusUnknown"
```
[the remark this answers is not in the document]
bobokun: I understand your point for the `==` but the problem with `"deleted" in None` would cause an error, which is why I added those defaults to begin with.

tron/kubernetes.py
bobokun: I agree, lets add this in a separate PR though :)

tron/kubernetes.py
```diff
@@ -148,35 +152,71 @@ def handle_event(self, event: Event) -> None:
...
+                            exit_code = exitcode.EXIT_KUBERNETES_NODE_SCALEDOWN
+                            self.log.warning("Tronjob failed due to Kubernetes scaling down a node.")
+                        if exit_code in KUBERNETES_EXIT_CODE_EXCEPTIONS:
+                            self.log.warning(
+                                f"If automatic retries are not enabled, run `tronctl retry {self.id}` to retry."
+                            )
+                            self.log.warning(
+                                "If this action is idempotent, then please consider enabling automatic retries for your action. If your action is not idempotent, then please configure this action to run on the stable pool rather than the default."
+                            )
+                    else:
```
nemacysts, line 211: i'm not sure we'll actually hit this case: the `if` / `elif` cases above this capture all the possible cases, right?

I think we'd want this at a slight different level: if the flow of the existing code was something like:
```
if abnormal_exit:
    ...
elif successful_run:
    ...
elif spot_interruption or node_scaledown:
    ...
then we'd want something like 
```
if abnormal_exit:
    ...
elif successful_run:
    ...
elif spot_interruption or node_scaledown:
    ...
else: # this case, where there was a "normal" failure
    ...
```

right now the logic is something like the following, though:
```
if success:
    ...
elif failure:
    ...
else: # block we're talking about
    ...
```
so it doesn't seem like we'll ever actually hit this case 

to verify: i think we could setup a tronjob with a tiny memory limit and then have it execute a python command that just does something like `print(bytearray(512000000))` - this should result in an oomkill and tron showing 137 as an exit code if i'm reading the code incorrectly - but if i'm reading it correctly, this test case will never hit this code

File: tron/utils/exitcode.py [hunk 10]
```diff
@@ -0,0 +1,26 @@
+# TRON-1826 Refactored code and moved exit codes to a new file to be imported in both Kubernetes.py and actionrun.py
```
nemacysts, line 1: suggestion: it's probably fine to obviate this comment - future CI ideally won't know/need to know where this came from originally :)

tests/kubernetes_test.py
```diff
@@ -151,6 +151,7 @@ def test_handle_event_exit_on_finished(mock_kubernetes_task):
             success=True,
         )
     )
+    assert mock_kubernetes_task.exit_status == -9
```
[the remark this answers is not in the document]
bobokun: oops. you're absolutely right.

tron/kubernetes.py
bobokun: you're right, I had this in the wrong place

nemacysts (approved):

PR #887, opened 2022-12-02
TRON-1723: Add user attribution into user agent when invoking tronctl commands

jfongatyelp:
This adds an option to our tron Client to add the current username to the useragent when connecting to the API.

This then adds attribution to any client calls through tronctl (though i also see an argument for just adding it to everything by default too, if we wanna go that route).

Manually ran a few tronview commands and confirmed user-agent remained the same in the logs, and confirmed tronctl commands added the user attribution

Tronview:
```
Dec  2 10:00:36 tronplayground1-uswest1cdevc tron.api.www.access b'"127.0.0.1" - - [02/Dec/2022:18:00:35 +0000] "GET /api/jobs/paasta-contract-monitor.long_job?include_action_runs=0&num_runs=10 HTTP/1.1" 200 5452 "-" "Tron Command/1.23.7 +http://github.com/Yelp/Tron"'
```

Tronctl:
```
Dec  1 18:46:04 tronplayground1-uswest1cdevc tron.api.www.access b'"127.0.0.1" - - [02/Dec/2022:02:46:03 +0000] "GET /api/ HTTP/1.1" 200 1428 "-" "Tron Command/1.23.7 +http://github.com/Yelp/Tron (jfong)"'
Dec  1 18:46:04 tronplayground1-uswest1cdevc tron.api.www.access b'"127.0.0.1" - - [02/Dec/2022:02:46:03 +0000] "POST /api/jobs/paasta-contract-monitor.long_job HTTP/1.1" 200 67 "-" "Tron Command/1.23.7 +http://github.com/Yelp/Tron (jfong)"'
Dec  1 18:46:04 tronplayground1-uswest1cdevc tron.api.www.access b'"127.0.0.1" - - [02/Dec/2022:02:46:03 +0000] "GET /api/jobs/paasta-contract-monitor.long_job/36806?include_action_graph=0&include_action_runs=0 HTTP/1.1" 200 448 "-" "Tron Command/1.23.7 +http://github.com/Yelp/Tron (jfong)"'
```

### DIFF
--- a/tron/commands/client.py
+++ b/tron/commands/client.py
@@ -2,10 +2,13 @@
 A command line http client used by tronview, tronctl, and tronfig
 """
 import logging
+import os
 import urllib.error
 import urllib.parse
 import urllib.request
 from collections import namedtuple
+from typing import Dict
+from typing import Mapping
 
 import tron
 from tron.config.schema import MASTER_NAMESPACE
@@ -68,8 +71,11 @@ def build_http_error_response(exc):
     return Response(exc.code, exc.msg, content)
 
 
-def request(uri, data=None, headers=None, method=None):
+def request(uri, data=None, headers=None, method=None, user_attribution=False):
     log.info("Request to %s with %s", uri, data)
+    headers = headers or default_headers
+    if user_attribution:
+        headers = ensure_user_attribution(headers)
     request = build_url_request(uri, data, headers=headers, method=method)
     try:
         response = urllib.request.urlopen(request)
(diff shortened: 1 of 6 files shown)


lanceleih:
This looks good to me, but just to confirm, this shouldn't affect Tronjobs run through SSH? It looks like this change is only when the job is run through tronctl manually.

bobokun (approved):
LGTM. If we merge this, can we also create the release with #886 alongside in the next version?

tron/commands/client.py
```diff
@@ -91,16 +97,27 @@ def build_get_url(url, data=None):
         return url
 
 
+def ensure_user_attribution(headers: Mapping[str, str]) -> Dict[str, str]:
+    headers = headers.copy()
+    if "User-Agent" not in headers:
+        headers["User-Agent"] = USER_AGENT
+    headers["User-Agent"] += f' ({os.environ.get("USER", "anonymous")})'
```
nemacysts, line 104: i wonder if there's a way for us to get the hostname of the box that a user executed the tronctl-CLUSTER wrapper from in case we can't reach someone that's abusing tron by scripting tronview/tronctl or something :p

nemacysts (approved):

tron/commands/client.py
jfongatyelp: What if we add a flag to tronctl to add more info to useragent arbitrarily, and then ensure our wrappers can set that ?

lanceleih (approved):

tron/commands/client.py
nemacysts: that's not a bad idea - i had a possibly worse idea that we just update the cluster wrapper to set an ORIG_HOST env var before executing tronctl on the master (since it doesn't seem like SSH exposes anything that we can use for this purpose) and using that value, but your approach seems more extensible if we ever decide we wanna add more info :p

PR #890, opened 2023-01-30
Adding ExecStopPost to systemd unit file to log tron exit status

cuza:
This PR adds a ExecStopPost to the systemd unit file to log tron exit status

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -2,7 +2,7 @@ Source: tron
 Section: admin
 Priority: optional
 Maintainer: Daniel Nephin <dnephin@yelp.com>
-Build-Depends: debhelper (>= 7), python3.6-dev, libdb5.3-dev, libyaml-dev, libssl-dev, libffi-dev, dh-virtualenv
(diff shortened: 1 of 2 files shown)


debian/control
```diff
@@ -2,7 +2,7 @@ Source: tron
 Section: admin
 Priority: optional
 Maintainer: Daniel Nephin <dnephin@yelp.com>
-Build-Depends: debhelper (>= 7), python3.6-dev, libdb5.3-dev, libyaml-dev, libssl-dev, libffi-dev, dh-virtualenv
+Build-Depends: bsdutils, debhelper (>= 7), python3.6-dev, libdb5.3-dev, libyaml-dev, libssl-dev, libffi-dev, dh-virtualenv
```
jfongatyelp, line 5: `Build-Depends` in this `Source` package section means that we need these packages in order to just build the package -- we don't actually depend on logger to build the deb, we just need it when the package is installed (i.e. when we're actually using the tron systemd unit file where we added the logger command).

So this should move down below to the `Depends` line (11).

nemacysts (commented):
lgtm after the control file is updated :)

debian/control
cuza: ooooopppsss!!!

nemacysts (approved):

PR #896, opened 2023-03-02
Adds Python 3.8; build for Jammy

bobokun:
Splits the commits into multiple commits for easier review.

This reverts PR https://github.com/Yelp/Tron/pull/879 and fixes dependencies and tests


nemacysts (approved):

ahaswell:
Could you check if 3.8 is an easy bump as 3.7 support ends in June.

.pre-commit-config.yaml
```diff
@@ -49,5 +49,5 @@ repos:
     rev: 22.3.0
     hooks:
       - id: black
-        language_version: python3.7
```
nemacysts: nit: we can probably get rid of this - this should be configured by the default_language_version at the top of this file

bin/tronctl
```diff
@@ -103,7 +103,7 @@ def parse_date(date_string):
 def parse_cli():
     parser = cmd_utils.build_option_parser()
     subparsers = parser.add_subparsers(dest="command", title="commands", help="Tronctl command to run")
-    subparsers.required = True  # add_subparsers only supports required arg from py37
+    subparsers.required = True  # add_subparsers only supports required arg from py38
```
nemacysts, line 106:
```python
subparsers = parser.add_subparsers(dest="command", title="commands", help="Tronctl command to run", required=True)
``` 
should work here now that we're gonna be running on 3.7+ :)

(this comment was referring to in `required - Whether or not a subcommand must be provided, by default False (added in 3.7)` https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_subparsers

setup.py
```diff
@@ -19,7 +19,7 @@
     classifiers=[
         "Programming Language :: Python",
         "Programming Language :: Python :: 3.6",
```
nemacysts, line 21: whoops, missed this before - can we also remove this line?

nemacysts (approved):
thank you! i have some minor nits/suggestions, but none are blocking

nemacysts:
ah, also need to update https://github.com/Yelp/Tron/blob/7de44d2ae50f87ea75519cc8c82ffb187b3157ee/itest.sh#L22

nemacysts (approved):

yelp_package/bionic/Dockerfile
```diff
@@ -18,9 +18,10 @@ RUN apt-get -q update && \
         libyaml-dev \
         libssl-dev \
         libffi-dev \
-        python3.6-dev \
+        python3.8-dev \
         python3-pip \
         python-tox \
+        rust-all \
```
EvanKrall, line 24: Rust?

EvanKrall (approved):

PR #900, opened 2023-03-07
Upgrades requirements and adds caching on get loggers

bobokun:
This PR upgrades some requirements to help us as we are seeing events being lost by Tron in pnw-devc on Python 3.8

- Upgrades Twisted to the latest version to support python 3.8
- Upgrades axe-core as previous version was incompatible
- adds lru_cache to logging to stop kubernetes from [locking on get loggers](https://github.com/kubernetes-client/python/issues/1867)
- Upgrades all dependencies recommended by dependabot to address security severities and further future proof.

### DIFF
--- a/tron/kubernetes.py
+++ b/tron/kubernetes.py
@@ -11,7 +11,7 @@ from task_processing.interfaces.event import Event
 from task_processing.plugins.kubernetes.task_config import KubernetesTaskConfig
 from task_processing.runners.subscription import Subscription
 from task_processing.task_processor import TaskProcessor
-from twisted.internet.defer import Deferred  # type: ignore  # need to upgrade twisted
+from twisted.internet.defer import Deferred
 from twisted.internet.defer import logError
 
 import tron.metrics as metrics
@@ -337,11 +337,12 @@ class KubernetesCluster:
         * control: events regarding how the task_proc plugin is running - handled directly
         * task: events regarding how the actual tasks/Pods we're running our doing - forwarded to KubernetesTask
         """
-        if self.deferred and not self.deferred.called:
+        if self.deferred is not None and not self.deferred.called:
             log.warning("Already have handlers waiting for next event in queue, not adding more")
             return
 
         self.deferred = self.queue.get()
+        assert self.deferred is not None
 
         # we want to process the event we just popped off the queue, but we also want
         # to form a sort of event loop, so we add two callbacks:
--- a/tron/kubernetes.py
+++ b/tron/kubernetes.py
@@ -342,8 +342,9 @@ class KubernetesCluster:
             return
 
         self.deferred = self.queue.get()
-        assert self.deferred is not None
-
+        if self.deferred is None:
+            log.warning("Unable to get a handler for next event in queue - this should never happen!")
+            return
         # we want to process the event we just popped off the queue, but we also want
         # to form a sort of event loop, so we add two callbacks:
         # * one to actually deal with the event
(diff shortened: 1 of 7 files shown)


tron/trondaemon.py
```diff
@@ -54,6 +55,7 @@ def setup_logging(options):
     # Show stack traces for errors in twisted deferreds.
     if options.debug:
         defer.setDebugging(True)
+    logging.getLogger = lru_cache(maxsize=None)(logging.getLogger)
```
nemacysts, line 58: we should have a decent bit of explanation here - we'll want to:
* mention why we're doing this
* mention that this is a bit of a hack and that ideally we'd get rid of it
* link to the kubernetes-client issue that this attempts to workaround so that we can remove this whenever we get rid of this hack
* maybe point folks at misc/jolt#148 internally for more details?

tron/kubernetes.py
```diff
@@ -337,11 +337,12 @@ def handle_next_event(self, _=None) -> None:
         * control: events regarding how the task_proc plugin is running - handled directly
         * task: events regarding how the actual tasks/Pods we're running our doing - forwarded to KubernetesTask
         """
-        if self.deferred and not self.deferred.called:
+        if self.deferred is not None and not self.deferred.called:
             log.warning("Already have handlers waiting for next event in queue, not adding more")
             return
 
         self.deferred = self.queue.get()
+        assert self.deferred is not None
```
nemacysts, line 345: i'm 99.99% sure this assert will never trigger - but just to be safe:
```suggestion
        if self.deferred is None:
            log.warning("Unable to get a handler for next event in queue - this should never happen!")
            return
```
should make mypy happy as well as ensure that we don't crash the whole process should the impossible be possible :)

bobokun: yup that is a much more elegant solution 😅

bobokun:
I've upgraded the dependencies from all the PRs requested from dependabot for futureproofing.

tron/trondaemon.py
```diff
@@ -54,6 +55,11 @@ def setup_logging(options):
     # Show stack traces for errors in twisted deferreds.
     if options.debug:
         defer.setDebugging(True)
+    # Adds lru_cache to getLogger as we are seeing kubernetes-client locking on logging causing potential events to be missed
+    # This is a workaround and ideally we would want to remove this once this is fixed on kubernetes-client library
+    # For more details on the issue visit: https://github.com/kubernetes-client/python/issues/1867
+    # misc/jolt#148 has a similar workaround for this issue as well. See https://github.yelpcorp.com/misc/jolt/pull/148
```
nemacysts, line 61: fix-n-ship:
```suggestion
    # Cache getLogger calls as we are seeing kubernetes-client locking on logging causing event processing to be extremely delayed
    # This is a workaround and ideally we would want to remove this once this is fixed upstream in kubernetes-client
    # For more details: https://github.com/kubernetes-client/python/issues/1867
    # For Yelpers, misc/jolt#148 has a similar workaround for this issue internally.
```

nemacysts (approved):
I have some minor revisions to the explanatory comment, but feel free to ship afterwards :)

I also did some checking on tronplayground and verified that the cached getLogger is being used everywhere :)

aantn:
Hey all, sorry to jump in on an old random thread but I was curious about the loggers fix and why the code for that was later removed. Looking to solve the problem myself for an app that uses the Kubernetes API (and I'm the person who originally opened https://github.com/kubernetes-client/python/issues/1867)

PR #903, opened 2023-03-09
TRON-1859: Updates date arithmetic to handle whitespace

bobokun:
```
>>> from tron.utils.timeutils import DateArithmetic
>>> da = DateArithmetic()
>>> da.parse('shortdate - 1')
'2023-03-08'
>>> da.parse('shortdate-1')
'2023-03-08'
>>> da.parse('shortdate - 2')
'2023-03-07'
```

### DIFF
--- a/tron/utils/timeutils.py
+++ b/tron/utils/timeutils.py
@@ -93,7 +93,7 @@ class DateArithmetic:
         based on date format (Ex: seconds for unixtime, days for day).
         """
         dt = dt or current_time()
-
+        date_str = date_str.replace(" ", "")
         match = cls.DATE_TYPE_PATTERN.match(date_str)
         if not match:
             return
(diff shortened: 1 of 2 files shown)


tron/utils/timeutils.py
```diff
@@ -93,7 +93,7 @@ def parse(cls, date_str, dt=None):
         based on date format (Ex: seconds for unixtime, days for day).
         """
         dt = dt or current_time()
-
+        date_str = date_str.replace(" ", "")
```
nemacysts, line 96: this should probably be a regex replace for whitespace - otherwise, `'shortdate -  1' would still fail to parse

bobokun: hm, I'm not sure how 'shortdate - 1' would fail to parse with my logic. I try not to use regex since it's slower 😅

nemacysts: ah - I was just mis-remembering how `replace()` worked :p (for some reason I had it in my head that it wouldn't replace multiple consecutive spaces)

nemacysts (approved):

tron/utils/timeutils.py
nemacysts: (in other words, my brain stops working after 5pm 🤣 )

PR #906, opened 2023-05-16
COMPINFRA-2565: enforce an upper limit on backfill concurrency

gmdfalk:
Problem 
-----
The Kubernetes control plane can be overwhelmed if a high number of concurrent backfills are executed. We currently set a relatively high default of `10` and enforce no upper limit at all.

Solution
-----
* Reduce the default `--max-parallel` from `10` to `3` to reduce the baseline pressure of Tron backfill jobs.
* Introduce and enforce a hard limit of `10` concurrent backfills per tronctl invocation

In a future PR, we might want to track the total count of running backfills per user in Zookeeper but this change should already safeguard against overwhelming Tron with backfills in the meantime.

Signed-off-by: Max Falk <gfalk@yelp.com>

### DIFF
--- a/bin/tronctl
+++ b/bin/tronctl
@@ -376,6 +376,9 @@ def backfill(args):
     if not args.id:
         print("Error: must provide at least one id argument")
         yield False
+    if args.max_parallel > LIMIT_MAX_PARALLEL_RUNS:
+        raise SystemExit(f"The flag --max-parallel exceeds the allowed limit of {LIMIT_MAX_PARALLEL_RUNS}. " +
+                         "Please reach out to the Tron team if you need to run backfills with higher limits.")
 
     if args.start_date:
         if args.end_date is None:
@@ -390,9 +393,6 @@ def backfill(args):
         print_backfill_cmds(job_name, date_strs)
         yield True
     else:
-        if args.max_parallel > LIMIT_MAX_PARALLEL_RUNS:
-            raise SystemExit(f"The flag --max-parallel exceeds the allowed limit of {LIMIT_MAX_PARALLEL_RUNS}. " +
-                             "Please reach out to the Tron team for configuring safe limits.")
         if confirm_backfill(job_name, date_strs):
             loop = asyncio.get_event_loop()
             try:
--- a/bin/tronctl
+++ b/bin/tronctl
@@ -170,7 +170,7 @@ def parse_cli():
             "If it needs a lot, keep this number low, because there may not be "
             "enough resources in the cluster too satisfy the demand, which can "
             "adversely affect other jobs. "
-            "The default is %(default) and cannot exceed the --max-parallel-limit."
+            "The default is %(default)s."
         ),
     )
     backfill_parser.add_argument(
@@ -301,8 +301,8 @@ def control_objects(args: argparse.Namespace):
             # and changing the API would be painful, so instead we call skip + publish separately from the client
             # (i.e., this file) to implement this functionality
             if request(
-               url=urljoin(args.server, tron_id.url),
-               data={"command": "skip"},
+                url=urljoin(args.server, tron_id.url),
+                data={"command": "skip"},
             ):
                 # a single action can have 0..N triggers to publish and these can be arbitrarily named, so we need to
                 # query the API and figure out what triggers exist
--- a/bin/tronctl
+++ b/bin/tronctl
@@ -174,18 +174,6 @@ def parse_cli():
         ),
     )
     backfill_parser.add_argument(
-        "--max-parallel-limit",
-        type=int,
-        dest="max_parallel_limit",
-        default=LIMIT_MAX_PARALLEL_RUNS,
-        help=(
-            "This flag controls the maximum value allowed for --max-parallel to "
-            "safeguard against overloading Tron. Please consult the Tron team before "
-            "you change this value for your job to avoid potential site events. "
-            "The default is %(default)."
-        ),
-    )
-    backfill_parser.add_argument(
         "--fail-on-error",
         dest="fail_on_error",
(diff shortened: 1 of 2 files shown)


bin/tronctl
```diff
@@ -300,8 +313,8 @@ def control_objects(args: argparse.Namespace):
             # and changing the API would be painful, so instead we call skip + publish separately from the client
             # (i.e., this file) to implement this functionality
             if request(
-                url=urljoin(args.server, tron_id.url),
-                data={"command": "skip"},
+                    url=urljoin(args.server, tron_id.url),
+                    data={"command": "skip"},
```
bobokun, line 317: Any reason we need to add another indent here?

bin/tronctl
```diff
@@ -371,6 +384,25 @@ def move(args):
     yield request(urljoin(args.server, "/api/jobs"), data)
 
 
+def get_validated_max_parallel(max_parallel, max_parallel_limit):
+    if max_parallel <= 0:
+        raise SystemExit(f"The flag --max-parallel must be a positive integer.")
+    if max_parallel_limit <= 0:
+        raise SystemExit(f"The flag --max-parallel-limit must be a positive integer.")
```
kawaiwanyelp, line 391: nit: s/`SystemExit`/`argparse.ArgumentTypeError`/g

bin/tronctl
```diff
@@ -371,6 +384,25 @@ def move(args):
...
+
+    if max_parallel_limit > LIMIT_MAX_PARALLEL_RUNS:
+        print(f"The flag --max-parallel-limit is overwritten to {max_parallel_limit} which is higher than " +
+              f"the suggested limit of {LIMIT_MAX_PARALLEL_RUNS}. If you haven't consulted the Tron team yet, " +
+              f"then please do so now to avoid overloading Tron with too many backfill jobs.")
```
kawaiwanyelp, line 396: This warning of "If you haven't consulted the Tron team yet, ..." is a little strange considering that the code currently doesn't stop them. I would either remove the warning, or request user confirmation (like a "have you consulted the team" thing) before continuing.

bin/tronctl
```diff
@@ -371,6 +384,25 @@ def move(args):
...
+
+    if max_parallel > max_parallel_limit:
+        print(f"The flag --max-parallel is set to {max_parallel} but the --max-parallel-limit only allows " +
+              f"{max_parallel_limit}. Overriding concurrency to enforce the limit of {max_parallel_limit}.")
```
kawaiwanyelp, line 400: I would exit here with an error saying the user is past the limit and should consult the Tron team if they need that many parallel runs, instead of proceeding with a `max_parallel` value that the user didn't intend.

bin/tronctl
```diff
@@ -371,6 +384,25 @@ def move(args):
     yield request(urljoin(args.server, "/api/jobs"), data)
 
 
+def get_validated_max_parallel(max_parallel, max_parallel_limit):
```
kawaiwanyelp, line 387: Please write a unit test for this function.

kawaiwanyelp (changes requested):

bin/tronctl
```diff
@@ -169,7 +170,19 @@ def parse_cli():
...
+        ),
+    )
+    backfill_parser.add_argument(
+        "--max-parallel-limit",
+        type=int,
+        dest="max_parallel_limit",
+        default=LIMIT_MAX_PARALLEL_RUNS,
+        help=(
+            "This flag controls the maximum value allowed for --max-parallel to "
+            "safeguard against overloading Tron. Please consult the Tron team before "
```
nemacysts, line 183:
```suggestion
            "safeguard against overloading Kubernetes. Please consult the #tron before "
```

bin/tronctl
```diff
@@ -169,7 +170,19 @@ def parse_cli():
...
+            "you change this value for your job to avoid potential site events. "
```
nemacysts, line 184:
```suggestion
            "you change this value for your job to avoid potential incidents. "
```

bin/tronctl
nemacysts, line 387: nit:
```suggestion
def get_validated_max_parallel(max_parallel: int, max_parallel_limit: int) -> int:
```

(although, I think we don't yet mypy this file)

bin/tronctl
nemacysts: ++ to requesting confirmation

bin/tronctl
nemacysts: ++

nemacysts: we'll also need to figure out what to do if someone does need more - do we just modify a local copy of tronctl to strip out this limiter and run the backfill for them?

tron/commands/backfill.py
```diff
@@ -13,7 +13,8 @@
 from tron.commands import display
 from tron.core.actionrun import ActionRun
 
-DEFAULT_MAX_PARALLEL_RUNS = 10
+DEFAULT_MAX_PARALLEL_RUNS = 3
```
nemacysts, line 16: !8ball `1` is a good default value?

bin/tronctl
nemacysts: we can probably get black to format this file - i bet it's skipping it since it has no extension

nemacysts (commented):
i'm going to revert some of the jammy/python upgrade stuff that is stopping us from releasing `master` onto our clusters and then fixup the comments :)

nemacysts:
we'll probably want to merge https://github.com/Yelp/Tron/pull/907 first so that we're back to a releaseable state

nemacysts (approved):
lgtm as a first step :)

bobokun (approved):

PR #909, opened 2023-05-26
TRON-1636: Pass secret_volumes to taskproc from kubernetes

Molaire:
Ticket: [TRON-1636](https://jira.yelpcorp.com/browse/TRON-1636)

This passes the serializable secret_volumes args to task_processor, but also create the needed validators for secret volumes to be parsed correctly in Tron.

- Validator enforces the 1 item in items array and secret_name must match key of item, in SecretVolumeItem
- Tested with patched wheel of https://github.com/Yelp/task_processing/pull/188. So once this task_proc PR is merged, we should update requirements.txt

This was tested with 2 real tron jobs and patched paasta and tron on our `tron-infrastage` box.
Success: https://fluffy.yelpcorp.com/i/bsJZQ3TlDR9N6fx3dNjT6WQfT0NVGh03.html
Jobs config: https://sourcegraph.yelpcorp.com/sysgit/yelpsoa-configs/-/blob/compute-infra-test-service/tron-infrastage.yaml?L94-119


tests/config/config_parse_test.py
```diff
@@ -1138,6 +1167,94 @@ def test_k8s_default_volumes(self):
...
+                container_path="/b/c",
+                default_mode="0644",
+                items=[dict(key="secret1", path="abc"), dict(key="secret1", path="abcd", mode="777"),],
+            ),
+            dict(secret_volume_name="abc", secret_name="secret1", container_path="/b/c", items=[],),
+            dict(secret_volume_name="abc", secret_name="secret1", container_path="/b/c",),
+        ],
+    )
+    def test_valid(self, config):
+        pass
```
Molaire, line 1255: oops

tests/config/config_parse_test.py
```diff
@@ -1138,6 +1162,94 @@ def test_k8s_default_volumes(self):
...
+
+    @pytest.mark.parametrize(
+        "config",
+        [
+            dict(
+                secret_volume_name="abc",
+                secret_name="secret1",
+                container_path="/b/c",
+                default_mode="0644",
+                items=[dict(key="secret1", path="abc"), dict(key="secret1", path="abcd", mode="777"),],
```
danielpops, line 1243: as we discovered last week, having multiple entries in the items list is probably not valid or expected. it's probably not necessary to include this test data and/or there should be a pre-condition check somewhere that treats this as an error

danielpops (approved):

tron/config/config_parse.py
```diff
@@ -270,6 +272,61 @@ class ValidateSecretSource(Validator):
 valid_secret_source = ValidateSecretSource()
 
 
+def valid_permission_mode(value, config_context):
```
nemacysts, line 275: super-nit:
```suggestion
def valid_permission_mode(value: Union[str, int], config_context: ConfigContext) -> str:
```

that said: thoughts on also add some validation either in paasta or in the soaconfigs hooks/tests?

tron/config/config_parse.py
```diff
@@ -270,6 +272,61 @@ class ValidateSecretSource(Validator):
...
+    try:
+        decimal_value = int(
+            str(value), base=8
+        )  # take in permission mode as string or int representation of an octal number. Goes from 0 to 4095 in decimal.
+    except ValueError:
+        error_msg = "Could not parse %s as octal permission mode at %s"
+        raise ConfigError(error_msg % (value, config_context.path))
+    if decimal_value > 4095 or decimal_value < 0:
+        error_msg = "Octal permission mode %s out of bound at %s"
+        raise ConfigError(error_msg % (value, config_context.path))
```
nemacysts, line 285: just to make this look more consistent with modern python:
```suggestion
        error_msg = "Could not parse {} as octal permission mode at {}"
        raise ConfigError(error_msg.format(value, config_context.path))
    if decimal_value > 4095 or decimal_value < 0:
        error_msg = "Octal permission mode {} out of bound at {}"
        raise ConfigError(error_msg.format(value, config_context.path))
```

tron/config/config_parse.py
```diff
@@ -270,6 +272,61 @@ class ValidateSecretSource(Validator):
...
+    validators = {
+        "container_path": valid_string,
+        "secret_volume_name": valid_string,
+        "secret_name": valid_string,
+        "default_mode": valid_permission_mode,
+        "items": build_list_of_type_validator(valid_secret_volume_item, allow_empty=True),
+    }
+
+    def post_validation(self, valid_input, config_context):
+        # check if each key of items matches the secret_name, and propagate default_mode.
```
nemacysts, line 320: how are we propagating default mode?

nemacysts: might also be good to leave a comment here about why we're checking that key == secret name (our secret system creates single-key k8s Secrets with the key being the secret name)

nemacysts (commented):
nice!

tron/config/config_parse.py
Molaire: I was thinking about adding it in yelpsoa pre-commit, as that will be the best place for developers to catch secret_volumes misconfiguration

tron/config/config_parse.py
Molaire: Added propagation.

tests/config/config_parse_test.py
Molaire: Added single-item restriction

danielpops (approved):

tron/config/config_parse.py
```diff
@@ -270,6 +273,68 @@ class ValidateSecretSource(Validator):
...
+        "container_path": valid_string,
+        "secret_volume_name": valid_string,
+        "secret_name": valid_string,
+        "default_mode": valid_permission_mode,
+        "items": build_list_of_type_validator(valid_secret_volume_item, allow_empty=True),
+    }
+
+    def post_validation(self, valid_input, config_context):
+        """Propagate default mode and enforce the secret-key match."""
+        # Our secrets will really only ever have one key, so weirdly we only care about a single item of this array AND it must have the same name as the secret (which is the single key)
```
nemacysts, line 322: we really should have picked a different config format for re-naming keys :/ 

although, I guess re-using the k8s format leaves us open to adding multi-key secret support eventually - that is to say, this is less weird and more an unfortunate side-effect of our choices :p

tron/config/config_parse.py
```diff
@@ -270,6 +273,68 @@ class ValidateSecretSource(Validator):
...
+        items = valid_input.get("items", [])
+
+        if len(items) > 1:
+            raise ConfigError("There is more than one item in the items array. Weirdly we don't want that.")
```
nemacysts, line 326:
```suggestion
            raise ConfigError("There is more than one item in the items array. This is unsupported as we don't support multi-key secrets.")
```

tron/core/action.py
```diff
@@ -31,6 +32,7 @@ class ActionCommandConfig:
     docker_parameters: set = field(default_factory=set)
     env: dict = field(default_factory=dict)
     secret_env: dict = field(default_factory=dict)
+    secret_volumes: List[ConfigSecretVolume] = field(default_factory=list)
```
nemacysts, line 35: hmm, should we be consistent with `extra_volumes` and use a set here?

tron/kubernetes.py
```diff
@@ -52,6 +53,23 @@ def combine_volumes(defaults: Collection[ConfigVolume], overrides: Collection[Co
     return list(result.values())
 
 
+def make_secret_volumes_serializable(secret_volumes: Collection[ConfigSecretVolume]) -> List[dict]:
```
nemacysts, line 56: it'd be cool if we could make `ConfigSecretVolume` have an `_asdict` that did this for us - but reading some mypy GH issues it seems like overriding methods on namedtuples won't always typecheck/work?

this would mostly be to have this code look consistent with extra_volumes - i.e.,
```Python
...
secret_volumes=[
    volume._asdict()
    for volume in secret_volumes
],
...
```

nemacysts: although, doing some testing in a python repl seems to imply that this will work!

tron/utils/scribereader.py
```diff
@@ -214,11 +214,11 @@ def read_log_stream_for_action_run(
         location_selector = f"-s {paasta_cluster}"
     truncation_message = (
         [
-            f"This output is truncated. Use this command to view all lines: scribereader {location_selector} {stream_name} --min-date {min_date.date()} --max-date {max_date.date()} | jq --raw-output 'select(.tron_run_number=={int(run_num)}) | .message'"
+            f"This output is truncated. Use this command to view all lines: scribereader {location_selector} {stream_name} --min-date {min_date.date()} --max-date {max_date.date()} | jq 'select(.tron_run_number=={int(run_num)}) | .message'"
         ]
         if max_date
         else [
-            f"This output is truncated. Use this command to view all lines: scribereader {location_selector} {stream_name} --min-date {min_date.date()} | jq --raw-output 'select(.tron_run_number=={int(run_num)}) | .message'"
+            f"This output is truncated. Use this command to view all lines: scribereader {location_selector} {stream_name} --min-date {min_date.date()} | jq 'select(.tron_run_number=={int(run_num)}) | .message'"
```
nemacysts, line 221: you'll probably want to pull master so that this change doesn't get undone

tron/core/action.py
Molaire: I had troubles with that due to many tests trying to hash it when it's a set

nemacysts: ah, I guess extra_volumes is something more hashable then? no worries then :)

requirements-minimal.txt
```diff
@@ -22,5 +22,6 @@ PyYAML>=5.1
 requests
 task_processing[mesos_executor,k8s]
 Twisted>=19.7.0
+typing-extensions>=3.10.0.0
```
nemacysts, line 25: we don't directly import typing_extensions, do we? requirements-minimal should only contain libraries that tron is directly importing, not transitive dependencies

nemacysts, line 25:
```suggestion
```

this line shouldn't be here since tron doesn't directly import this - we have a bit of a funky requirements setup here since some of our dependencies aren't public, but the typing_extensions dependency should be coming from task_processing which is a public dependency

can you share (either here or in slack) the error you're seeing without this line?

nemacysts (approved):

PR #915, opened 2023-07-07
Update to the latest task_proc

nemacysts:
This version pulls in a fix that stops pods from being launched with a null request. This has been running fine in infrastage where I've verified that Pods are indeed being launched with the correct metadata

I've verified this as well with unit tests in task_proc :)


nemacysts:
https://github.com/Yelp/task_processing/pull/210 is what this version bump brings in

bobokun (approved):

PR #923, opened 2023-08-07
Allow Tron to use more than 10k file descriptors

EvanKrall:


EvanKrall:
I've tested this change on stageg manually, it seems to take effect at least:

```
$ grep '^Limit\|^Max open files' /proc/$(pgrep trond)/limits
Limit                     Soft Limit           Hard Limit           Units
Max open files            100000               100000               files
```

jfongatyelp (approved):

gmdfalk (approved):

nemacysts (approved):

PR #925, opened 2023-08-14
Un-symlink itest dockerfiles

nemacysts:
Something appears to have changed with either GHA or the docker-compose version we're pulling in and the existing symlink setup looks to be confusing docker-compose.

We can probably cleanup (or delete) these cluster itests since they're all mesos-related, but that's a bigger change for another day :)


Molaire:
It's green! Well done

Molaire (approved):

jfongatyelp (approved):

PR #926, opened 2023-08-21
Catch all exceptions in k8s submit_command()

nemacysts:
It's entirely possible that creating a task_processing task (and/or submitting one) can result in an exception. At the moment, this results in the affected ActionRun getting stuck in the Starting state - but this is a lie and means that the normal monitoring/alerting on failed runs does not kick in.

### DIFF
--- a/tron/core/actionrun.py
+++ b/tron/core/actionrun.py
@@ -9,7 +9,6 @@ from typing import Optional
 
 from dataclasses import dataclass
 from dataclasses import fields
-from pyrsistent import InvariantException
 from twisted.internet import reactor
 
 from tron import command_context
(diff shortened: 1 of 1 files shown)


jfongatyelp (approved):

bobokun (approved):
lgtm, if an exception does occur will it show up in META in tronweb? If not it might be useful to add the details there so users have a better understanding of why the job failed

Molaire (approved):

nemacysts:
> lgtm, if an exception does occur will it show up in META in tronweb? If not it might be useful to add the details there so users have a better understanding of why the job failed

@bobokun not atm - I *think* that i could probably get it to show up there if i did something like `task.log.exception(...)` - but i'm a little wary of doing that in case there's something funky in creating the task that means we never even initialized the logger

nemacysts:
(merging in master so that tests pass)

PR #930, opened 2023-11-16
A quick script to check if a job took longer than x time to run and update python in precommit

EmanElsaban:
This script outputs the jobs that takes longer than x time to run. We can specify the time using ``--time``.  It's an easy way to get the jobs than to look at the Tron UI


tron/bin/check_job_exceeding_time.py
```diff
@@ -0,0 +1,95 @@
...
+    cmd_utils.setup_logging(args)
+    cmd_utils.load_config(args)
+    client = Client(args.server, args.cluster_name)
+    result = []
+
+    url_index = client.index()
+    if args.job is None:
+        jobs = client.jobs(include_job_runs=True)
+        for job in jobs:
+            check_job_time(job=job, time_limit=args.time_limit, result=result)
```
nemacysts, line 81: non-blocking suggestion: could we make this a pure function by having it return a list of affected runs? 

(and then do something like `results.extend(check_job_time(...))`?

tron/bin/check_job_exceeding_time.py
```diff
@@ -0,0 +1,95 @@
...
+import pytimeparse
+
+from tron.commands import cmd_utils
+from tron.commands.client import Client
+
+
+log = logging.getLogger("check_exceeding_time")
+
+
+def parse_cli():
```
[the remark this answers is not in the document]
nemacysts: and depending on how easy it is to trace the tron code to get the type of some of these things, it might be worth using `reveal_type` (see https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html#when-you-re-puzzled-or-when-things-are-complicated) to have mypy tell you what it thinks the types are

tron/bin/check_job_exceeding_time.py
nemacysts, line 1: super-nit: should this be something like `get_jobs_exceeding_runtime`? usually `check_*` functions are intended to be used as monitoring checks and will return non-zero if things are failing checks (versus the current behavior here, which is just printing)

tron/bin/check_job_exceeding_time.py
```diff
@@ -0,0 +1,95 @@
...
+        type=int,
+        dest="time_limit",
+        default=18000,
+    )
+    args = parser.parse_args()
+    return args
+
+
+def check_if_time_exceeded(job_runs, job_expected_runtime, result):
+    states_to_check = {"queued", "scheduled", "cancelled", "skipped"}
```
nemacysts, line 33: nit: we repeat this - might be worth extracting this to a named constant

tron/bin/check_job_exceeding_time.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+def is_job_run_exceeding_expected_runtime(job_run, job_expected_runtime):
+    states_to_check = {"queued", "scheduled", "cancelled", "skipped"}
+    if (
+        job_expected_runtime is not None
+        and job_run.get(
+            "state",
+            "unknown",
+        )
```
nemacysts, line 51: do we need to do this check in both places? looks like this check is also being done by the function that calls this

tron/bin/check_job_exceeding_time.py
```diff
@@ -0,0 +1,95 @@
...
+        not in states_to_check
+    ):
+        duration_seconds = pytimeparse.parse(job_run.get("duration", ""))
+        if duration_seconds and duration_seconds > job_expected_runtime:
+            return True
```
nemacysts, line 56: nit: 
```suggestion
        return duration_seconds and duration_seconds > job_expected_runtime:
```

tron/bin/check_job_exceeding_time.py
```diff
@@ -0,0 +1,95 @@
...
+    return False
+
+
+def check_job_time(job, time_limit, result):
+    job_runs = sorted(
+        job.get("runs", []),
+        key=lambda k: (k["end_time"] is None, k["end_time"], k["run_time"]),
+        reverse=True,
+    )
```
nemacysts, line 65: just curious: are we sorting so that the output is ordered later on? should we sort at printing time so that we only need to sort once if so?

tron/bin/check_job_exceeding_time.py
```diff
@@ -0,0 +1,95 @@
...
+    if args.job is None:
+        jobs = client.jobs(include_job_runs=True)
+        for job in jobs:
+            check_job_time(job=job, time_limit=args.time_limit, result=result)
+    else:
+        job_url = client.get_url(args.job)
+        job = client.job_runs(job_url)
+        check_job_time(job=job, client=client, url_index=url_index, result=result)
+
+    if result is None:
```
nemacysts, line 87: should this be `if not result`? we default `result` to `[]` - which is not `None`

tron/bin/check_job_exceeding_time.py
EmanElsaban: I don't think so, Idk why I initially thought the state might change so I added it in both places

.pre-commit-config.yaml
```diff
@@ -41,8 +42,8 @@ repos:
...
         files: ^tests/.*\.py$
   - repo: http://github.com/psf/black
-    rev: 19.10b0
+    rev: 23.3.0
     hooks:
       - id: black
-        language_version: python3.6
-        args: [--target-version, py36]
+        language_version: python3.8
+        args: [--target-version, py38]
```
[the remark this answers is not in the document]
EmanElsaban: its failed test cause:
```
ERROR: InvocationError for command '/home/runner/work/Tron/Tron/.tox/py36/bin/pre-commit run --all-files' (exited with code 1)
```
but when I try to run this locally I don't have py36 on this branch, its .tox/py38/...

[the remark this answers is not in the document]
EmanElsaban: then when I run make test after make clean:
```
(py38) emanelsabban@dev208-uswest1adevc:~/pg/Tron$ make test
tox -e py36
py36 create: /nail/home/emanelsabban/pg/Tron/.tox/py36
ERROR: InterpreterNotFound: python3.6
________________________________________________________________________________________________________ summary _________________________________________________________________________________________________________
ERROR:  py36: InterpreterNotFound: python3.6
make: *** [Makefile:64: test] Error 1
```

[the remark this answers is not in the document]
nemacysts: oh, that's unfortunate :/

I'm just a tad worried about breaking the py36 build and/or somehow allowing py36+ code to sneak in and break at runtime (on py36)

tron/bin/get_jobs_exceeding_runtime.py
```diff
@@ -0,0 +1,86 @@
...
+        help="This is used to specify the time that if any job exceeds will show. Defaults to 5 hours",
+        type=int,
+        dest="time_limit",
+        default=18000,
+    )
+    args = parser.parse_args()
+    return args
+
+
+def check_if_time_exceeded(job_runs, job_expected_runtime) -> list:
```
nemacysts, line 36:
```suggestion
def check_if_time_exceeded(job_runs: List[Dict[str, Any]], job_expected_runtime: int) -> List[str]:
```

i'm sort of guessing about the `job_runs` typing here - that said, if we're just using this script for the current project it's probably fine to be a little lax with the types and just do the easy ones (e.g.,
```suggestion
def check_if_time_exceeded(job_runs: List, job_expected_runtime: int) -> List[str]:
```

tron/bin/get_jobs_exceeding_runtime.py
```diff
@@ -0,0 +1,86 @@
...
+        if job_run.get("state", "unknown") not in STATES_TO_CHECK:
+            if is_job_run_exceeding_expected_runtime(
+                job_run,
+                job_expected_runtime,
+            ):
+                result.append(job_run["id"])
+    return result
+
+
+def is_job_run_exceeding_expected_runtime(job_run, job_expected_runtime) -> bool:
```
nemacysts, line 48:
```suggestion
def is_job_run_exceeding_expected_runtime(job_run: Dict[str, Any], job_expected_runtime: int) -> bool:
```

tron/bin/get_jobs_exceeding_runtime.py
```diff
@@ -0,0 +1,86 @@
...
+    if job_expected_runtime is not None:
+        duration_seconds = pytimeparse.parse(job_run.get("duration", ""))
+        return duration_seconds and duration_seconds > job_expected_runtime
+    return False
+
+
+def check_job_time(job, time_limit) -> list:
```
nemacysts, line 55:
```suggestion
def check_job_time(job: Dict[str, Any], time_limit: int) -> List[str]:
```

nemacysts (approved):
i've got some minor comments re: the types - but if i'm blocking you feel free to ignore them as they're not particularly serious and we can always come back and improve them :)

bobokun (dismissed):

nemacysts:
@EmanElsaban is this still something we're interested in merging?

nemacysts:
(doesn't look like we need this anymore)

PR #932, opened 2023-12-04
Check to see if we error on any events in handle_event function -  TRON-2051

EmanElsaban:
This PR adds a try/except block to check to see if we error on any event in handle_event function. Erroring on any event means we don't handle this event at all and thus we don't update Tron's state machine for it

### DIFF
--- a/tron/kubernetes.py
+++ b/tron/kubernetes.py
@@ -1,4 +1,5 @@
 import logging
+import json
 from logging import Logger
 from typing import cast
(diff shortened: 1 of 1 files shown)


nemacysts:
(we figured out that the issue we'd been looking for was really a bad interaction between the k8s client lib and python loggers - every model added a new logger and every setLevel call operated on all the loggers that had been created, which slowed down over time)

PR #934, opened 2024-01-17
Upgrading Tron to py3.8 + patching it with the fix

EmanElsaban:
This PR includes python upgrades for 3.8, the patch to stop tron jobs from getting stuck, fixes to some mypy issues/ tests failing and files edited by the precommit hook + additional PR in task_processing for adding logging and handling of ADDED events (PR : https://github.com/Yelp/task_processing/pull/212)


tron/commands/client.py
```diff
@@ -18,7 +18,7 @@
 
     assert simplejson  # Pyflakes
 except ImportError:
-    import json as simplejson
+    import json as simplejson  # type: ignore
```
nemacysts, line 21: at some point we should replace all these simplejson usages with using/importing the stdlib version (since they should be the same thing)

but that's for another day!

tron/core/action.py
```diff
@@ -56,14 +56,14 @@ class Action:
...
-    executor: str = None
-    trigger_downstreams: (bool, dict) = None
-    triggered_by: set = None
-    on_upstream_rerun: str = None
-    trigger_timeout: datetime.timedelta = None
+    retries: Optional[int] = None
+    retries_delay: Optional[datetime.timedelta] = None
+    expected_runtime: Optional[datetime.timedelta] = None
+    executor: Optional[str] = None
+    trigger_downstreams: (bool, dict) = None  # type: ignore
```
nemacysts, line 63: i *think* this is probably meant to be `Union[bool, dict]`?

tron/core/actionrun.py
```diff
@@ -1300,9 +1302,7 @@ def kill(self, final: bool = True) -> Optional[str]:
 
         return "\n".join(msgs)
 
-    def handle_action_command_state_change(
-        self, action_command: ActionCommand, event: str, event_data=None
-    ) -> Optional[bool]:
+    def handle_action_command_state_change(self, action_command: ActionCommand, event: str, event_data=None):
```
nemacysts, line 1305: looks like this lost the `-> Optional[bool]`?

tron/kubernetes.py
```diff
@@ -274,21 +262,9 @@ def handle_event(self, event: Event) -> None:
...
-                    json.dumps(
-                        {
-                            "event": str(event),
-                            "exception": type(e),
-                            "exception_message": str(e),
-                        }
-                    ),
-                )
+            self.log.exception(
+                f"unable to handle an event for id={event_id} for event={str(event)}, exception type: {type(e)} and exception message: {str(e)}"
```
nemacysts, line 266: log.exception() will print out the exception below - we can just log the info we want to add on top of the exception itself

tron/serialize/runstate/dynamodb_state_store.py
```diff
@@ -24,7 +24,7 @@ def __init__(self, name, dynamodb_region, stopping=False) -> None:
         self.dynamodb_region = dynamodb_region
         self.table = self.dynamodb.Table(name)
         self.stopping = stopping
-        self.save_queue = OrderedDict()
+        self.save_queue = OrderedDict()  # type: OrderedDict
```
nemacysts, line 27: we should be able to type this as `self.save_queue: OrderedDict = OrderedDict()`

that said, not sure why mypy is suddenly unhappy here

tests/commands/retry_test.py
```diff
@@ -181,26 +187,3 @@ def test_wait_for_retry_deps_done(fake_retry_action, mock_client_request, event_
         data=dict(command="retry", use_latest_command=1),
         user_attribution=True,
     )
-
-
-@mock.patch.object(retry, "RetryAction", autospec=True)
-def test_retry_actions(mock_retry_action, mock_client, event_loop):
```
Molaire, line 187: I previously removed this for the validation-only Jammy tron version. I think this is relevant for an actual tron release

Makefile
nemacysts, line 1: we probably want the existing pip index url shenanigans :)

Makefile
```diff
@@ -78,17 +80,14 @@ debitest_%: deb_% _itest_%
 itest_%: debitest_%
 	@echo "itest $* OK"
 
-cluster_itests:
-	tox -e cluster_itests
-
 dev:
-	SSH_AUTH_SOCK=$(SSH_AUTH_SOCK) .tox/py36/bin/trond --debug --working-dir=dev -l logging.conf --host=0.0.0.0
+	SSH_AUTH_SOCK=$(SSH_AUTH_SOCK) .tox/py38/bin/trond --debug --working-dir=dev -l logging.conf --host=0.0.0.0 --web-path=/nail/home/emanelsabban/pg/Tron/tronweb
```
nemacysts, line 84:
```suggestion
	SSH_AUTH_SOCK=$(SSH_AUTH_SOCK) .tox/py38/bin/trond --debug --working-dir=dev -l logging.conf --host=0.0.0.0
```

dev/config/MASTER.yaml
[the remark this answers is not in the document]
EmanElsaban: Okay will remove the current config from MASTER.yaml and create later a wiki page with instructions

debian/rules
```diff
@@ -18,11 +20,15 @@ override_dh_auto_test:
 override_dh_virtualenv:
 	echo $(PIP_INDEX_URL)
 	dh_virtualenv --index-url $(PIP_INDEX_URL) \
-	            --extra-pip-arg  --trusted-host=169.254.255.254 \
-                --python=/usr/bin/python3.6 \
-                --preinstall cython==0.29.36 \
-                --preinstall pip==9.0.1 \
-                --preinstall setuptools==46.1.3
```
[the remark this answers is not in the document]
EmanElsaban: Cool, will bring back the above config and just change python to 3.8 and pip to 18.1 version

tron/kubernetes.py
EmanElsaban: I added all the info needed in the self.log.exception message, such as event_id, the event that failed, type of the exception and exception message

EmanElsaban: ohh you mean add the extra info in a self.log.info message following the exception message?

tests/commands/retry_test.py
EmanElsaban: I kept the test and it seems to be passing

tron/kubernetes.py
```diff
@@ -122,136 +125,144 @@ def handle_event(self, event: Event) -> None:
...
-            container_statuses = pod_status.get("containerStatuses", []) or []
-            exit_code = 0 if k8s_type == "finished" else exitcode.EXIT_KUBERNETES_ABNORMAL
-
-            if len(container_statuses) > 1 or len(container_statuses) == 0:
-                # shouldn't happen right now, but who knows what future us will do :p
-                self.log.error(
-                    "Got an event for a Pod with zero or multiple containers - not inspecting payload to verify success."
+            # Might be that somewhere in this function, something is throwing an exception and we're not handling it
+            # wrap this whole function in a big block of try/except block
+            # if handle_event errors out for any reason then we will never see this event at all
```
nemacysts, line 131: small suggestion re: re-wording your debugging notes :)
```suggestion
            # we wrap this entire thing in a try-except as otherwise an error in 
            # logging (which is useful, but not critical) will result in us not
            # processing an event at all (which is critical!)
```

tron/kubernetes.py
```diff
@@ -338,12 +349,14 @@ def handle_next_event(self, _=None) -> None:
...
-        if self.deferred and not self.deferred.called:
+        if self.deferred is not None and not self.deferred.called:
             log.warning("Already have handlers waiting for next event in queue, not adding more")
             return
 
         self.deferred = self.queue.get()
-
+        if self.deferred is None:
+            log.warning("Unable to get a handler for next event in queue - this should never happen!")
+            return
```
nemacysts, line 359: hmm, i wonder if we should crash here - if this has happened we're going to completely stop processing events

nemacysts: or maybe we should add in a fake deferred so that we can still add in the callbacks and retry later?

Makefile
```diff
@@ -11,38 +11,40 @@ endif
 
 NOOP = true
 ifeq ($(PAASTA_ENV),YELP)
-	# This index must match the Ubuntu codename in the dockerfile.
-	DOCKER_PIP_INDEX_URL ?= http://169.254.255.254:20641/bionic/simple/
+	export PIP_INDEX_URL ?= http://169.254.255.254:20641/jammy/simple/
```
nemacysts, line 14: we removed a comment here, but we still need to handle what the comment was referring to - i'm assuming that if we do `make itest_bionic` locally it will fail since it'll pull down jammy wheels in a bionic container

nemacysts: (i left a comment in the PR that added this index url stuff for a way to handle multiple codenames)

tron/kubernetes.py
EmanElsaban: we tend to return if this happens with the log warning message

nemacysts: if that happens, we'll never process events again since we setup a psuedo loop here with the callbacks - which is not great since we'll likely not notice this until it's too late to easily recover the current state of things with a tron restart given that the completed pods will have aged out by the time we notice this

tron/kubernetes.py
```diff
@@ -338,22 +349,25 @@ def handle_next_event(self, _=None) -> None:
...
+        if self.deferred is not None and not self.deferred.called:
             log.warning("Already have handlers waiting for next event in queue, not adding more")
             return
 
         self.deferred = self.queue.get()
-
+        if self.deferred is None:
+            log.warning("Unable to get a handler for next event in queue - this should never happen!")
+            # TODO: figure out how to recover if we were unable to get a handler
+            # Not adding a callback is very bad here as this means we will never handle this event
```
nemacysts, line 360: super-nit if we want to be extra verbose for future-us:
```suggestion
            # Not adding a callback is very bad here as this means we will never handle future events
            # This hasn't happened to date, and if it does hopefully the errback added in a previous
            # handle_next_event() will get us back on track for now 
```

tron/core/actionrun.py
```diff
@@ -1046,7 +1176,7 @@ def submit_command(self, attempt: ActionRunAttempt) -> Optional[KubernetesTask]:
                 service_account_name=attempt.command_config.service_account_name,
                 ports=attempt.command_config.ports,
             )
-        except Exception:
+        except InvariantException:
```
nemacysts, line 1179: technically, we could have other exceptions thrown here (e.g., `build_environment()` or `filehandler.OutputStreamSerializer()` could throw - but those do seem somewhat unlikely

nemacysts (approved):

Makefile
```diff
@@ -11,7 +11,7 @@ endif
 
 NOOP = true
 ifeq ($(PAASTA_ENV),YELP)
-	export PIP_INDEX_URL ?= https://pypi.yelpcorp.com/simple
+	export PIP_INDEX_URL ?= http://169.254.255.254:20641/$*/simple/
```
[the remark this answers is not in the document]
EmanElsaban: yupp, same I thought I would need to add it at first to the targets themselves :p

tron/core/actionrun.py
EmanElsaban: hmm should we change this to ``except Exception`` to catch all other exceptions based on the hierarchy  shown here: https://martinxpn.medium.com/exception-hierarchy-python-58-100-days-of-python-9d8585e6569b#:~:text=The%20Exception%20Hierarchy,class%20in%20the%20exception%20hierarchy.

nemacysts: yea, unless we're extra-sure that none of the other things we do here will throw, it's probably a good idea to change this back to `except Exception`

EmanElsaban: or 
```
except (InvariantException, Exception):
```

EmanElsaban: I'm honestly not sure if anything else throws or not since I didn't really touch this code, I will change it in a separate follow up PR  and not here

EmanElsaban: LOL, I'm so confused as of why it shows  as added in this PR. This change was added three years ago: https://github.com/Yelp/Tron/blame/b1118433b10c06a6f15b668d2f5c184115d701e5/tron/core/actionrun.py

EmanElsaban: according to this PR it should be Exception: https://github.com/Yelp/Tron/pull/926/files , not sure what brought it back in this PR

nemacysts (approved):

PR #935, opened 2024-01-19
SEC-18515: Bump boto3 version to enable implicit IMDSv2 usage

Molaire:
Sorry for revert, I pushed to master by accident initially.
Ticket: [SEC-18515](https://jira.yelpcorp.com/browse/SEC-18515)

We want to bump boto3 version here so tron boxes stop making IMDSv1 calls.
This is mimicking a py3.6-compatible bump I did here: https://github.yelpcorp.com/services/biz_pages/pull/111

I don't think we need to release a Jammy version, because I don't expect yelpsoa validation to use tron's boto3

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,8 +7,8 @@ aws-sam-translator==1.15.1
 aws-xray-sdk==2.4.2
 backcall==0.1.0
 boto==2.49.0
-boto3==1.9.249
-botocore==1.12.249
+boto3==1.17.44
+botocore==1.20.112
 bsddb3==6.2.6
 cachetools==4.2.1
 certifi==2019.11.28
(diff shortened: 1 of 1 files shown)


nemacysts:
@Molaire what's your timeline look like? we're likely going to be rolling out the fix for tron on newer pythons very soon, so depending on how @EmanElsaban and @KaspariK feel, we can bundle this in that release or release this separately after they're done

danielpops (approved):

Molaire:
Best case scenario: we are wishing to be ready to enforce imdsv2 end of January

Molaire:
Closing because Jammy upgrade will bring boto3 to a good version

PR #939, opened 2024-02-12
Adding tron run id to pod labels for k8s

wilmer05:
Adding run id for tron tasks into the labels of the pods in kubernetes.

### DIFF
--- a/tests/bin/action_runner_test.py
+++ b/tests/bin/action_runner_test.py
@@ -146,3 +146,26 @@ class TestBuildEnvironment:
             TRON_RUN_NUM="10",
             TRON_ACTION="bar.baz",
         )
+
+
+class TestBuildLabels:
+    def test_build_labels(self):
+        labels = action_runner.build_labels("MASTER.foo.10.bar")
+
+        assert labels == dict(
+            TRON_RUN_NUM="10",
+        )
+
+    def test_build_labels_invalid_run_id(self):
+        labels = action_runner.build_labels("asdf")
+
+        assert labels == dict(
+            TRON_RUN_NUM="UNKNOWN",
+        )
+
+    def test_build_labels_too_long_run_id(self):
+        labels = action_runner.build_labels("MASTER.foo.10.bar.baz")
+
+        assert labels == dict(
+            TRON_RUN_NUM="10",
+        )
--- a/tron/bin/action_runner.py
+++ b/tron/bin/action_runner.py
@@ -93,6 +93,24 @@ def build_environment(run_id, original_env=None):
     return new_env
 
 
+def build_labels(run_id, original_labels=None):
+    if original_labels is None:
+        original_labels = dict()
+
+    try:
+        run_num = run_id.split(".", maxsplit=3)[2]
+    except IndexError:
+        # if we can't parse the run_id, we don't want to abort, so just
+        # set these semi-arbitrarily
+        run_num = "UNKNOWN"
+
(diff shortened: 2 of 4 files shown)


tron/bin/action_runner.py
```diff
@@ -93,6 +93,24 @@ def build_environment(run_id, original_env=None):
     return new_env
 
 
+def build_labels(run_id, original_labels=None):
```
nemacysts, line 96: small suggestion:
```suggestion
def build_labels(
    run_id: str,
    original_labels: Optional[Dict[str, str]]=None,
) -> Dict[str, str]:
```
not 100% necessary since we're not running mypy on this file yet, but it'd be nice to add these in right now since we're pretty sure with the typing here :)

tron/bin/action_runner.py
```diff
@@ -93,6 +93,24 @@ def build_environment(run_id, original_env=None):
...
+    if original_labels is None:
+        original_labels = dict()
+
+    try:
+        run_num = run_id.split(".", maxsplit=3)[2]
```
nemacysts, line 101: small suggestion:
```suggestion
        # reminder: the format here is "namespace.job.run_num.action"
        _, _, run_num, _ = run_id.split(".", maxsplit=3)
```
would allow us to drop the `[2]` and add a little documentation for future yelpers readin this function :)

tron/bin/action_runner.py
```diff
@@ -93,6 +93,24 @@ def build_environment(run_id, original_env=None):
...
+    except IndexError:
+        # if we can't parse the run_id, we don't want to abort, so just
+        # set these semi-arbitrarily
+        run_num = "UNKNOWN"
+
+    new_labels = dict(original_labels)
+    new_labels["TRON_RUN_NUM"] = run_num
```
nemacysts, line 108: (no action needed) hopefully we can update our python version soon and use the dictionary merging operator :p

tron/bin/action_runner.py
```diff
@@ -93,6 +93,24 @@ def build_environment(run_id, original_env=None):
...
+
+    logging.debug(new_labels)
```
nemacysts, line 110:
```suggestion
```

tests/bin/action_runner_test.py
```diff
@@ -146,3 +146,26 @@ def test_build_environment_too_long_run_id(self):
             TRON_RUN_NUM="10",
             TRON_ACTION="bar.baz",
         )
+
+
+class TestBuildLabels:
```
nemacysts, line 151: suggestion: should we add some testcases where there's existing labels as well?

tron/bin/action_runner.py
wilmer05: added it now :)

tron/bin/action_runner.py
wilmer05: added

tron/bin/action_runner.py
wilmer05: removed

tests/bin/action_runner_test.py
wilmer05: Added them now 👍

nemacysts (approved):

wilmer05:
Added a more yelpy name for the label now @nemacysts

gmdfalk (approved):
We don't use `tron.yelp.com` as label prefix anywhere yet. But i guess this is a good a time as any to start.

PR #940, opened 2024-02-13
TRON-1850: Include starting pods in check for stuck jobs

jfongatyelp:
We can sometimes get stuck in state `starting` if we somehow miss events coming in from k8s (or a pod somehow itself get stuck before it starts running).

This should now consider these situations as `stuck` jobs and actions alongside those that are actually running too long or waiting on a trigger, and adds some tests to verify. (These tests were confirmed to fail w/o the changes to include 'starting' jobs/actions).

### DIFF
--- a/tron/bin/check_tron_jobs.py
+++ b/tron/bin/check_tron_jobs.py
@@ -226,7 +226,7 @@ def get_relevant_run_and_state(job_content):
         state = run.get("state", "unknown")
         if state in ["failed", "succeeded", "unknown", "skipped"]:
             return run, State(state)
-        elif state in ["running", "waiting"]:
+        elif state in ["running", "waiting", "starting"]:
             action_state = is_action_failed_or_unknown(run)
             if action_state != State.SUCCEEDED:
                 return run, action_state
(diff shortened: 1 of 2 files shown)


tron/bin/check_tron_jobs.py
```diff
@@ -319,7 +320,7 @@ def get_relevant_action(*, action_runs, last_state, actions_expected_runtime):
                     actions_expected_runtime,
                 ):
                     return action_run
-                if action_state == "running":
+                if action_state == "running" or action_state == "starting":
```
nemacysts, line 323:
```suggestion
                if action_state in {"running", "starting"}:
```

tron/bin/check_tron_jobs.py
```diff
@@ -251,7 +251,7 @@ def is_job_stuck(
 ):
     next_run_time = None
     for job_run in job_runs:
-        states_to_check = {"running", "waiting"}
+        states_to_check = {"running", "waiting", "starting"}
```
nemacysts, line 254: heh, i like how we're pretty inconsistent re: sets and lists in this file :p

nemacysts: we could also move this outside the loop so that we're not re-creating a set multiple times, but it's probably fine to punt on refactoring this script :p

nemacysts (approved):

tron/bin/check_tron_jobs.py
jfongatyelp: Hm good call, though i also briefly saw we had https://github.com/Yelp/Tron/blob/master/tron/core/jobrun.py#L456 which makes me wonder if we should just reuse this and why didn't we already, but agree I'll leave the archaeology for another time.

nemacysts (approved):

bobokun (approved):

PR #941, opened 2024-02-14
Stop keeping old copies of eventbus files

nemacysts:
We've never restored from these "backups", and as long as we ensure that the current file points to a fully written file, it's fine to only keep at most 2 files around: the actual current file and a temporary "new" file.

To ensure that we're not pointing at a half-written file, we continue using the age-old pattern for atomic file updates: write to another file and swap a symlink once the write is complete

### DIFF
--- a/tron/eventbus.py
+++ b/tron/eventbus.py
@@ -144,9 +144,10 @@ class EventBus:
         duration = time.time() - started
         log.info(f"log read from disk, took {duration:.4}s")
 
-    def sync_save_log(self, reason):
+    def sync_save_log(self, reason: str) -> bool:
         started = time.time()
         new_file = os.path.join(self.log_dir, f"{int(started)}.pickle")
+        previous_file = os.path.realpath(os.path.join(self.log_dir, "current.pickle"))
         try:
             with open(new_file, "xb") as f:
                 pickle.dump(self.event_log, f)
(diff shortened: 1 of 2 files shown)


tron/eventbus.py
```diff
@@ -144,9 +145,18 @@ def sync_load_log(self):
...
+    def sync_save_log(self, reason: str) -> bool:
         started = time.time()
         new_file = os.path.join(self.log_dir, f"{int(started)}.pickle")
+        previous_file: Optional[str] = os.path.realpath(os.path.join(self.log_dir, "current"))
+        # if we're starting  a fresh Tron server, there won't be a current symlink
+        # and the above line will give us the path to what will eventually be the current
+        # symlink...which is undesirable since we clean up whatever this points to :p
+        # we can tell if this is happening since this previous_file variable should
+        # always point to a file that ends in .pickle under normal operation
+        if previous_file and previous_file.endswith("current"):
```
bobokun, line 157: is the `previous_file.endswith("current")` part redundant? Since we are defining it in line 151, we can assume it would always end in "current" if it is not None.

nemacysts: it would be, but `current` is a symlink and `os.path.realpath` will return the actual path of whatever you give it (so if you give it a normal file/dir, it'll return the same thing - but if you give it a symlink it'll return the symlink target

bobokun (approved):

PR #948, opened 2024-04-03
Adding logs to Tron to indicate start and scheduling times - TRON-2152

EmanElsaban:
This PR is for the tickets: [TRON-2153](https://jira.yelpcorp.com/browse/TRON-2153) and [TRON-2152](https://jira.yelpcorp.com/browse/TRON-2152)

### DIFF
--- a/bin/trond
+++ b/bin/trond
@@ -4,6 +4,7 @@ import argparse
 import faulthandler
 import logging
 import os
+import time
 import traceback
 
 import pkg_resources
@@ -168,10 +169,10 @@ def setup_environment(args):
 
 def main():
     args = parse_cli()
-
+    tron_start_time = time.time()
     setup_environment(args)
     trond = trondaemon.TronDaemon(args)
-    trond.run()
+    trond.run(tron_start_time=tron_start_time)
 
 
 if __name__ == "__main__":
--- a/bin/trond
+++ b/bin/trond
@@ -172,7 +172,7 @@ def main():
     boot_time = time.time()
     setup_environment(args)
     trond = trondaemon.TronDaemon(args)
-    trond.run(boot_time=boot_time)
+    trond.run(boot_time)
 
 
 if __name__ == "__main__":
--- a/bin/trond
+++ b/bin/trond
@@ -169,10 +169,10 @@ def setup_environment(args):
 
 def main():
     args = parse_cli()
-    tron_start_time = time.time()
+    boot_time = time.time()
     setup_environment(args)
     trond = trondaemon.TronDaemon(args)
-    trond.run(tron_start_time=tron_start_time)
+    trond.run(boot_time=boot_time)
 
 
 if __name__ == "__main__":
--- a/tron/core/actionrun.py
+++ b/tron/core/actionrun.py
@@ -301,7 +301,6 @@ class ActionRun(Observable):
         on_upstream_rerun=None,
         trigger_timeout_timestamp=None,
         original_command=None,
-        start_schedule_jobs=True,
     ):
         super().__init__()
         self.job_run_id = maybe_decode(job_run_id)
@@ -1135,8 +1134,6 @@ class MesosActionRun(ActionRun, Observer):
 class KubernetesActionRun(ActionRun, Observer):
     """An ActionRun that executes the command on a Kubernetes cluster."""
 
-    start_schedule_jobs = True
-
     def submit_command(self, attempt: ActionRunAttempt) -> Optional[KubernetesTask]:
         """
         Attempt to run a given ActionRunAttempt on the configured Kubernetes cluster.
@@ -1195,11 +1192,11 @@ class KubernetesActionRun(ActionRun, Observer):
         self.watch(task)
 
         try:
-            if self.start_schedule_jobs:
+            if k8s_cluster.start_schedule_jobs:
                 log.info(
                     f"We will start submitting tasks (i.e scheduling jobs). Starting with {task.get_kubernetes_id()}"
                 )
-                self.start_schedule_jobs = False
+                k8s_cluster.start_schedule_jobs = False
             k8s_cluster.submit(task)
         except Exception:
             log.exception(f"Unable to submit task for ActionRun {self.id}")
--- a/tron/core/actionrun.py
+++ b/tron/core/actionrun.py
@@ -1194,7 +1194,7 @@ class KubernetesActionRun(ActionRun, Observer):
         try:
             if k8s_cluster.start_schedule_jobs:
                 log.info(
-                    f"We will start submitting tasks (i.e scheduling jobs). Starting with {task.get_kubernetes_id()}"
+                    f"We will start submitting tasks (i.e running the jobs). Starting with {task.get_kubernetes_id()}"
                 )
                 k8s_cluster.start_schedule_jobs = False
             k8s_cluster.submit(task)
--- a/tron/core/actionrun.py
+++ b/tron/core/actionrun.py
@@ -301,6 +301,7 @@ class ActionRun(Observable):
         on_upstream_rerun=None,
         trigger_timeout_timestamp=None,
         original_command=None,
+        start_schedule_jobs=True,
     ):
         super().__init__()
         self.job_run_id = maybe_decode(job_run_id)
@@ -1134,6 +1135,8 @@ class MesosActionRun(ActionRun, Observer):
 class KubernetesActionRun(ActionRun, Observer):
     """An ActionRun that executes the command on a Kubernetes cluster."""
 
+    start_schedule_jobs = True
+
     def submit_command(self, attempt: ActionRunAttempt) -> Optional[KubernetesTask]:
         """
         Attempt to run a given ActionRunAttempt on the configured Kubernetes cluster.
@@ -1192,6 +1195,11 @@ class KubernetesActionRun(ActionRun, Observer):
         self.watch(task)
 
         try:
+            if self.start_schedule_jobs:
+                log.info(
+                    f"We will start submitting tasks (i.e scheduling jobs). Starting with {task.get_kubernetes_id()}"
+                )
+                self.start_schedule_jobs = False
             k8s_cluster.submit(task)
         except Exception:
             log.exception(f"Unable to submit task for ActionRun {self.id}")
--- a/tron/core/actionrun.py
+++ b/tron/core/actionrun.py
@@ -1192,11 +1192,6 @@ class KubernetesActionRun(ActionRun, Observer):
         self.watch(task)
 
         try:
-            if k8s_cluster.start_schedule_jobs:
-                log.info(
-                    f"We will start submitting tasks (i.e running the jobs). Starting with {task.get_kubernetes_id()}"
-                )
-                k8s_cluster.start_schedule_jobs = False
             k8s_cluster.submit(task)
         except Exception:
             log.exception(f"Unable to submit task for ActionRun {self.id}")
--- a/tron/core/job_collection.py
+++ b/tron/core/job_collection.py
@@ -1,4 +1,5 @@
 import logging
+import time
 
 from tron.core.job import Job
 from tron.utils import collections
@@ -40,6 +41,8 @@ class JobCollection:
             else:
                 return config.namespace == namespace_to_reconfigure
 
+        # The below statements don't get called until self.state_watcher.watch_all is called
+        # in apply_config. This will go through the job's configs and build a scheduler for them
         seq = (factory.build(config) for config in job_configs.values() if reconfigure_filter(config))
         return map_to_job_and_schedule(filter(self.add, seq))
 
@@ -66,7 +69,14 @@ class JobCollection:
         job_scheduler.schedule_reconfigured()
         return True
 
-    def restore_state(self, job_state_data, config_action_runner):
+    def restore_state(self, job_state_data, config_action_runner, tron_start_time=None):
+        # we will start looping through the jobs and their runs and restore
+        # state for each run in for a job and as we restore the state
+        # we will also schedule the next runs for that job
+        if tron_start_time is not None:
+            log.info(
+                f"Tron is restoring state for jobs and will start scheduling them! Time elapsed since Tron started {time.time()- tron_start_time}"
+            )
         for name, state in job_state_data.items():
             self.jobs[name].restore_state(state, config_action_runner)
         log.info(f"Loaded state for {len(job_state_data)} jobs")
--- a/tron/core/job_collection.py
+++ b/tron/core/job_collection.py
@@ -41,8 +41,7 @@ class JobCollection:
             else:
                 return config.namespace == namespace_to_reconfigure
 
-        # The below statements don't get called until self.state_watcher.watch_all is called
-        # in apply_config. This will go through the job's configs and build a scheduler for them
+        # NOTE: as this is a generator expression, we will only go through job configs and build a scheduler for them once something iterates over us (i.e, once `self.state_watcher.watch_all()` is called)
         seq = (factory.build(config) for config in job_configs.values() if reconfigure_filter(config))
         return map_to_job_and_schedule(filter(self.add, seq))
 
@@ -69,13 +68,15 @@ class JobCollection:
         job_scheduler.schedule_reconfigured()
         return True
 
-    def restore_state(self, job_state_data, config_action_runner, tron_start_time=None):
-        # we will start looping through the jobs and their runs and restore
-        # state for each run in for a job and as we restore the state
-        # we will also schedule the next runs for that job
-        if tron_start_time is not None:
+    def restore_state(self, job_state_data, config_action_runner, boot_time=None):
+        """
+        Loops through the jobs and their runs in order to restore
+        state for each run. As we restore state, we will also schedule the next
+        runs for each job
+        """
+        if boot_time is not None:
             log.info(
-                f"Tron is restoring state for jobs and will start scheduling them! Time elapsed since Tron started {time.time()- tron_start_time}"
+                f"Tron is restoring state for jobs and will start scheduling them! Time elapsed since Tron started {time.time()- boot_time}"
             )
         for name, state in job_state_data.items():
             self.jobs[name].restore_state(state, config_action_runner)
--- a/tron/core/job_collection.py
+++ b/tron/core/job_collection.py
@@ -1,5 +1,4 @@
 import logging
-import time
 
 from tron.core.job import Job
 from tron.utils import collections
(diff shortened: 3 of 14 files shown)


bin/trond
```diff
@@ -168,10 +169,10 @@ def setup_environment(args):
 
 def main():
     args = parse_cli()
-
+    tron_start_time = time.time()
```
nemacysts, line 172: can we re-use the `boot_time` naming that already exists? i think in mcp.py we have a variable that we call `boot_time` for when the mcp object is created (might actually be nice to unify that with this, actually)

tron/trondaemon.py
```diff
@@ -126,14 +138,15 @@ def _run_mcp(self):
...
-            self.mcp.initial_setup()
+            self.mcp.initial_setup(tron_start_time=tron_start_time)
         except Exception as e:
             msg = "Error in configuration %s: %s"
             log.exception(msg % (config_path, e))
             raise
 
     def _run_reactor(self):
         """Run the twisted reactor."""
+        # Starts Tron server
```
nemacysts, line 149:
```suggestion
        # this is what actually starts the Tron server by starting the Twisted event loop
```

tron/trondaemon.py
```diff
@@ -85,11 +85,23 @@ def run(self):
                 signal.SIGUSR1: self._handle_debug,
             }
             signal.pthread_sigmask(signal.SIG_BLOCK, signal_map.keys())
-
-            self._run_mcp()
+            log.info("Starting the setup process of mcp, api, manhole and runnning the reactor")
```
nemacysts, line 88:
```suggestion
            log.info("Starting setup processes...")
```

tron/trondaemon.py
```diff
@@ -85,11 +85,23 @@ def run(self):
...
+                f"The Tron API has finished setup. Time elapsed since Tron started {time.time() - tron_start_time}s"
+            )
             self._run_manhole()
+            log.info(
+                f"The manhole has finished setup. Time elapsed since Tron started {time.time() - tron_start_time}s"
+            )
             self._run_reactor()
+            log.info(
+                f"Twisted reactor has started. The Tron API should be up and ready now to receive requests. Time elapsed since Tron started {time.time() - tron_start_time}s"
+            )
```
nemacysts, line 104:
```suggestion
            log.info(
                f"Twisted reactor has started. The Tron API should be up and ready now to receive requests. Time elapsed since Tron started: {time.time() - tron_start_time}s"
            )
            log.info("Setup complete!")
```

tron/trondaemon.py
```diff
@@ -85,11 +85,23 @@ def run(self):
                 signal.SIGUSR1: self._handle_debug,
             }
             signal.pthread_sigmask(signal.SIG_BLOCK, signal_map.keys())
-
-            self._run_mcp()
+            log.info("Starting the setup process of mcp, api, manhole and runnning the reactor")
+            self._run_mcp(tron_start_time=tron_start_time)
+            log.info(
+                f"The master control plane has finished setup. Time elapsed since Tron started {time.time() - tron_start_time}s"
+            )
```
nemacysts, line 92:
```suggestion
            log.info(
                f"Master Control Program (MCP) setup complete. Time elapsed since Tron started: {time.time() - tron_start_time}s"
            )
```

tron/trondaemon.py
```diff
@@ -85,11 +85,23 @@ def run(self):
...
             self._run_www_api()
+            log.info(
+                f"The Tron API has finished setup. Time elapsed since Tron started {time.time() - tron_start_time}s"
```
nemacysts, line 95:
```suggestion
                f"Tron API setup complete. Time elapsed since Tron started: {time.time() - tron_start_time}s"
```

tron/trondaemon.py
```diff
@@ -85,11 +85,23 @@ def run(self):
...
+            )
             self._run_manhole()
+            log.info(
+                f"The manhole has finished setup. Time elapsed since Tron started {time.time() - tron_start_time}s"
```
nemacysts, line 99:
```suggestion
                f"Manhole setup complete. Time elapsed since Tron started: {time.time() - tron_start_time}s"
```

tron/serialize/runstate/statemanager.py
```diff
@@ -263,6 +265,7 @@ def update_from_config(self, state_config):
             return False
 
         self.shutdown()
+        # The function below will spin up a thread that will be saving into dynamodb, which will run as daemon
```
nemacysts, line 268:
```suggestion
        # NOTE: this will spin up a thread that will constantly persist data into dynamodb
```

tron/serialize/runstate/statemanager.py
```diff
@@ -146,6 +146,8 @@ def restore(self, job_names, skip_validation=False):
             self._restore_metadata()
 
         jobs = self._restore_dicts(runstate.JOB_STATE, job_names)
+        # jobs should be a dictionary that contains  job name and number of runs
+        # {'Master.k8s': {'run_nums':[0], 'enabled': True}, 'Master.cits_test_frequent_1': {'run_nums'; [1,0], 'enabled': True}}
```
nemacysts, line 150: would be nice to add a TypedDict or something for the return type of `_restore_dicts()` so that this comment isn't necessary :)

nemacysts, line 150:
```suggestion
        # jobs should be a dictionary that contains  job name and number of runs
        # {'MASTER.k8s': {'run_nums':[0], 'enabled': True}, 'MASTER.cits_test_frequent_1': {'run_nums': [1,0], 'enabled': True}}
```

just to make this more accurate

tron/mcp.py
```diff
@@ -150,15 +163,15 @@ def get_job_collection(self):
     def get_config_manager(self):
         return self.config
 
-    def restore_state(self, action_runner):
+    def restore_state(self, action_runner, tron_start_time=None):
         """Use the state manager to retrieve to persisted state and apply it
         to the configured Jobs.
         """
-        log.info("restoring")
+        log.info("restoring from DynamoDB")
```
nemacysts, line 170:
```suggestion
        log.info("Restoring from DynamoDB")
```
would be more consistent with the other logs we're adding (sentence cased)

tron/mcp.py
```diff
@@ -53,30 +56,34 @@ def reconfigure(self, namespace=None):
...
                 self.config.load().get_master().action_runner,
             ),
+            tron_start_time=tron_start_time,
         )
         # Any job with existing state would have been scheduled already. Jobs
         # without any state will be scheduled here.
         self.jobs.run_queue_schedule()
 
-    def apply_config(self, config_container, reconfigure=False, namespace_to_reconfigure=None):
+    def apply_config(self, config_container, reconfigure=False, namespace_to_reconfigure=None, tron_start_time=None):
```
nemacysts, line 86: this will get called pretty frequently post-boot, so i don't think we want to pass through the boot time here

tron/mcp.py
```diff
@@ -103,16 +110,22 @@ def apply_config(self, config_container, reconfigure=False, namespace_to_reconfi
...
             namespace_to_reconfigure,
         )
+        # We will build the schedulers once the watcher is invoked
+        log.info(
+            f"Tron built the schedulers for Tron jobs internally! Time elapsed since Tron started {time.time() - tron_start_time}s"
+        )
         self.state_watcher.watch_all(updated_jobs, [Job.NOTIFY_STATE_CHANGE, Job.NOTIFY_NEW_RUN])
 
     def build_job_scheduler_factory(self, master_config, job_graph):
+        """Creates an action runner if option specified in tronfigs and job scheduler instances for Tron jobs"""
```
nemacysts, line 128:
```suggestion
        """Creates JobSchedulerFactory, which are how Tron tracks job schedules internally"""
```

tron/mcp.py
```diff
@@ -103,16 +110,22 @@ def apply_config(self, config_container, reconfigure=False, namespace_to_reconfi
 
         # TODO: unify NOTIFY_STATE_CHANGE and simplify this
         self.job_graph = JobGraph(config_container)
+        # Builds a job scheduler factory for Tron jobs, which will schedule internally in Tron the jobs
```
nemacysts, line 113:
```suggestion
        # this factory is how Tron internally manages scheduling jobs
```

tron/eventbus.py
```diff
@@ -140,6 +140,7 @@ def _has_event(self, event_id):
 
     def sync_load_log(self):
         started = time.time()
+        # we are loading the pickle object from the "current" file
```
nemacysts, line 143: imo, we can probably remove this comment as its mostly restating what the code is doing - not why (or adding additional information)

tron/mcp.py
```diff
@@ -22,6 +22,8 @@ def apply_master_configuration(mapping, master_config):
     def get_config_value(seq):
         return [getattr(master_config, item) for item in seq]
 
+    # mapping consists of different functions in MCP mapped to their config from tronfig in Master.yaml
```
nemacysts, line 25:
```suggestion
    # Map various MASTER.yaml config options to functions that will apply said options
```

tron/core/jobrun.py
```diff
@@ -462,6 +462,7 @@ def get_first_queued(self, node=None):
         )
 
     def get_scheduled(self):
+        # find the scheduled run for the job and return it
```
nemacysts, line 465:
```suggestion
        # find the scheduled runs for the job and return them
        # in most cases, there should just be a single run - but it's possible that a delayed job could have N scheduled runs built up
```

tron/mcp.py
```diff
@@ -38,7 +40,8 @@ def __init__(self, working_dir, config_path):
         self.context = command_context.CommandContext()
         self.state_watcher = statemanager.StateChangeWatcher()
         self.boot_time = time.time()
-        log.info("initialized")
+        current_time = time.strftime("%a, %d %b %Y %H:%M:%S", time.localtime())
+        log.info(f"Initialized. Tron started on {current_time}!")
```
nemacysts, line 44: should this re-use `boot_time`?

tron/mcp.py
```diff
@@ -53,30 +56,34 @@ def reconfigure(self, namespace=None):
             log.exception(f"reconfigure failure: {e.__class__.__name__}: {e}")
             raise e
 
-    def _load_config(self, reconfigure=False, namespace_to_reconfigure=None):
+    def _load_config(self, reconfigure=False, namespace_to_reconfigure=None, tron_start_time=None):
```
nemacysts, line 59: this will get called pretty frequently post-boot, so i don't think we want to pass through the boot time here

tron/mcp.py
```diff
@@ -53,30 +56,34 @@ def reconfigure(self, namespace=None):
...
+                tron_start_time=tron_start_time,
             )
 
-    def initial_setup(self):
+    def initial_setup(self, tron_start_time=None):
         """When the MCP is initialized the config is applied before the state.
         In this case jobs shouldn't be scheduled until the state is applied.
         """
-        self._load_config()
+        # The job schedules have been applied in first step
```
nemacysts, line 73: what does first step refer to here?

tron/mcp.py
```diff
@@ -53,30 +56,34 @@ def reconfigure(self, namespace=None):
...
+        self._load_config(tron_start_time=tron_start_time)
+        # Now the jobs will actually get scheduled once the state for action runs are restored
```
nemacysts, line 75:
```suggestion
        # jobs will also get scheduled (internally) once the state for action runs are restored
```

tron/core/actionrun.py
```diff
@@ -1134,6 +1135,8 @@ def handle_action_command_state_change(self, action_command, event, event_data=N
 class KubernetesActionRun(ActionRun, Observer):
     """An ActionRun that executes the command on a Kubernetes cluster."""
 
+    start_schedule_jobs = True
```
nemacysts, line 1138: I'm a little iffy on adding a boolean here - can we maybe add something to where we initialize the task_proc client instead?

tron/core/job_collection.py
```diff
@@ -40,6 +41,8 @@ def reconfigure_filter(config):
             else:
                 return config.namespace == namespace_to_reconfigure
 
+        # The below statements don't get called until self.state_watcher.watch_all is called
+        # in apply_config. This will go through the job's configs and build a scheduler for them
```
nemacysts, line 45:
```suggestion
        # NOTE: as this is a generator expression, we will only go through job configs and build a scheduler for them once something iterates over us (i.e, once `self.state_watcher.watch_all()` is called)
```

tron/core/job_collection.py
```diff
@@ -66,7 +69,13 @@ def update(self, new_job_scheduler):
         job_scheduler.schedule_reconfigured()
         return True
 
-    def restore_state(self, job_state_data, config_action_runner):
+    def restore_state(self, job_state_data, config_action_runner, tron_start_time=None):
+        # we will start looping through the jobs and their runs and restore
+        # state for each run in for a job and as we restore the state
+        # we will also schedule the next runs for that job
```
nemacysts, line 75: this could probably be this functions docstring :)

```suggestion
        """
        Loops through the jobs and their runs in order to restore
        state for each run. As we restore state, we will also schedule the next 
        runs for each job
        """
```

nemacysts (commented):
ty for adding more comments for how things work!

the bulk of my comments are generally just tweaking wording (and can be ignored if you'd like) - there's only a couple that we should definitely dig into (e.g., logging when the first k8s schedule happens as well as some of the time logs for functions that get called both for startup as well as for new config deployment)

bin/trond
EmanElsaban: yea, I was using it at first but then added tron_start_time to start timing as soon as we enter main instead of waiting till we get into "MasterControlProgram" class

tron/mcp.py
EmanElsaban: to load_config function, we first load the configs and apply them then second step would be the restore_state function

EmanElsaban: will change it to ``The job schedule factories will be created in the function below``

tron/kubernetes.py
```diff
@@ -279,12 +279,15 @@ def __init__(
...
         pod_launch_timeout: Optional[int] = None,
+        start_schedule_jobs: bool = True,
     ):
         # general k8s config
         self.kubeconfig_path = kubeconfig_path
         self.enabled = enabled
         self.default_volumes: Optional[List[ConfigVolume]] = default_volumes or []
         self.pod_launch_timeout = pod_launch_timeout or DEFAULT_POD_LAUNCH_TIMEOUT_S
+        # This flag will be used to indicate in logs the first task we start scheduling and running on Kubernetes
+        self.start_schedule_jobs = start_schedule_jobs
```
nemacysts, line 290: i think we can probably skip this logic and say that we can start scheduling jobs after `self.connect()` has been called further below :)

tron/mcp.py
```diff
@@ -30,15 +32,16 @@ def get_config_value(seq):
...
         super().__init__()
         self.jobs = JobCollection()
         self.working_dir = working_dir
         self.config = manager.ConfigManager(config_path)
         self.context = command_context.CommandContext()
         self.state_watcher = statemanager.StateChangeWatcher()
-        self.boot_time = time.time()
-        log.info("initialized")
+        self.boot_time = boot_time
+        current_time = time.strftime("%a, %d %b %Y %H:%M:%S", time.localtime(boot_time))
```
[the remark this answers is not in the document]
EmanElsaban: yup it prints
```
2024-04-05 05:03:11,461 tron.trondaemon INFO Starting setup processes...
2024-04-05 05:03:19,723 tron.mcp INFO Initialized. Tron started on Fri, 05 Apr 2024 05:03:16!
```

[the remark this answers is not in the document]
EmanElsaban: I think if we don't pass anything to localtime() or pass None , it converts the current time instead

tron/kubernetes.py
EmanElsaban: hmm i wonder if we should add that log message after self.connect to indicate that we connected to task_proc and to also keep the current log produced after the flag "start_schedule_jobs". Because we do all the connection stuff under ``self._run_mcp(boot_time=boot_time)`` meanwhile we actually start submitting tasks to k8s when we are waiting after we run the reactor, at this point in the code
```
            while True:
                signum = signal.sigwait(list(signal_map.keys()))
```

nemacysts: the log message from the `start_schedule_jobs` flag will probably be somewhat unreliable if we succeed in lowering the restart time as it will depend on job schedules 

that said: the task_proc stuff does not run on the twisted reactor, so it should be possible for it to run beforehand

(although, I guess that until the reactor is running there won't be any new tasks submitted to task_proc)

nemacysts: in general, i'm a little hesitant to add a state-indicating boolean just for logging purposes, so perhaps we can try to use the tron/trondaemon.py logging as a proxy timing metric instead?

EmanElsaban: Actually, they both happen after the reactor runs under ``submit_command`` function. I wonder if we should still keep both though cause there are some function calls between them (i.e time spent in other functions)

tron/mcp.py
```diff
@@ -154,11 +165,16 @@ def restore_state(self, action_runner):
         """Use the state manager to retrieve to persisted state and apply it
         to the configured Jobs.
         """
-        log.info("restoring")
+        log.info("Restoring from DynamoDB")
         states = self.state_watcher.restore(self.jobs.get_names())
         MesosClusterRepository.restore_state(states.get("mesos_state", {}))
-
+        log.info(
+            f"Tron will start restoring state for the jobs and will start scheduling them! Time elapsed since Tron started {time.time()- self.boot_time}"
```
nemacysts, line 172: huh, i just moved this to my landscape-oriented monitor and noticed that `black` didn't format the f-string
```suggestion
            f"Tron will start restoring state for the jobs and will start scheduling them! Time elapsed since Tron started {time.time() - self.boot_time}"
```

tron/mcp.py
```diff
@@ -154,11 +165,16 @@ def restore_state(self, action_runner):
...
+        )
         self.jobs.restore_state(states.get("job_state", {}), action_runner)
+        log.info(
+            f"Tron completed restoring state for the jobs. Time elapsed since Tron started {time.time()- self.boot_time}"
```
nemacysts, line 176: same nit as above
```suggestion
            f"Tron completed restoring state for the jobs. Time elapsed since Tron started {time.time() - self.boot_time}"
```

tron/trondaemon.py
```diff
@@ -75,7 +75,7 @@ def __init__(self, options):
         self.signals = {signal.SIGINT: signal.default_int_handler}
         self.manhole_sock = f"{self.options.working_dir}/manhole.sock"
 
-    def run(self):
+    def run(self, boot_time=None):
```
nemacysts, line 78: (chatting about this internally: but on closer inspection, i think we want to make sure that this is set or add some guard clauses to the log statements below - otherwise, we're leaving a bit of a footgun around)

nemacysts (commented):
generally lgtm - i can ship once we resolve the optional boot_time comment

(the other two are stylistic nits)

tron/serialize/runstate/dynamodb_state_store.py
```diff
@@ -136,7 +136,7 @@ def _consume_save_queue(self):
                 log.error(error)
                 with self.save_lock:
                     self.save_queue[key] = val
-        duration = time.time() - start
```
nemacysts, line 139: was this intentional?

EmanElsaban: nopee, will commit the past version

nemacysts (approved):

PR #949, opened 2024-04-11
Adding yelp_clog S3LogsReader

yaroliakh:
Enabling `S3LogsReader` from yelp_clog as an alternative method for reading tron Paasta logs.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,8 +7,8 @@ aws-sam-translator==1.15.1
 aws-xray-sdk==2.4.2
 backcall==0.1.0
 boto==2.49.0
-boto3==1.26.85
-botocore==1.29.86
+boto3==1.34.80
+botocore==1.34.80
 bsddb3==6.2.7
 cachetools==4.2.1
 certifi==2022.12.7
@@ -75,7 +75,7 @@ requests==2.25.0
 requests-oauthlib==1.2.0
 responses==0.10.6
 rsa==4.9
-s3transfer==0.6.0
+s3transfer==0.10.1
 setuptools==65.5.1
 six==1.15.0
 sshpubkeys==3.1.0
--- a/tests/utils/scribereader_test.py
+++ b/tests/utils/scribereader_test.py
@@ -3,13 +3,89 @@ from unittest import mock
 
 import pytest
 
+import tron.utils.scribereader
 from tron.utils.scribereader import read_log_stream_for_action_run
 
 try:
     import scribereader  # noqa: F401
+    from clog.readers import S3LogsReader
 except ImportError:
-    pytest.skip("scribereader not available, skipping tests", allow_module_level=True)
+    pytest.skip("yelp logs readers not available, skipping tests", allow_module_level=True)
 
+
+def static_conf_patch(args={}):
+    return lambda arg, namespace, default=None: args.get(arg)  
+
+
+def test_read_log_stream_for_action_run_not_available():    
+    orig_scribereader = tron.utils.scribereader.scribereader
+    orig_S3LogsReader = tron.utils.scribereader.S3LogsReader
+    tron.utils.scribereader.scribereader = None
+    tron.utils.scribereader.S3LogsReader = None
+    try:
+        output =  tron.utils.scribereader.read_log_stream_for_action_run(
+            "namespace.job.1234.action",
+            component='stdout',
+            min_date=datetime.datetime.now(),
+            max_date=datetime.datetime.now(),
+            paasta_cluster='fake'
+        )
+    finally:
+        tron.utils.scribereader.scribereader = orig_scribereader
+        tron.utils.scribereader.S3LogsReader = orig_S3LogsReader
+    assert 'unable to display logs' in output[0]
+
+
+def test_read_log_stream_for_action_run_yelp_clog():
+    with mock.patch(
+        "staticconf.read", autospec=True, side_effect=static_conf_patch({
+            "logging.use_s3_reader": True,
+            "logging.max_lines_to_display": 1000
+        })
+    ), mock.patch(
+        "tron.config.static_config.build_configuration_watcher",
+        autospec=True,
+    ), mock.patch(
+        "tron.config.static_config.load_yaml_file",
+        autospec=True,
+    ), mock.patch(
+        "tron.utils.scribereader.S3LogsReader", autospec=True
+    ) as mock_s3_reader:
+        
+        mock_s3_reader.return_value.get_log_reader.return_value = iter(
+            [
+                """{
+                "tron_run_number": 1234,
+                "component": "stdout",
+                "message": "line 1",
+                "timestamp": "2021-01-02T18:10:09.169421619Z",
+                "cluster": "fake"
+            }""",
+                """{
+                "tron_run_number": 1234,
+                "component": "stdout",
+                "message": "line 2",
+                "timestamp": "2021-01-02T18:11:09.169421619Z",
+                "cluster": "fake"
+            }""",
+                """{
+                "tron_run_number": 1234,
+                "component": "stderr",
+                "message": "line 3",
+                "timestamp": "2021-01-02T18:12:09.169421619Z",
+                "cluster": "fake"
+            }""",
+            ]
+        )
+
+        output = read_log_stream_for_action_run(
+            "namespace.job.1234.action",
+            component='stdout',
+            min_date=datetime.datetime.now(),
+            max_date=datetime.datetime.now(),
+            paasta_cluster='fake'
+        )
+    assert output == ["line 1", "line 2"]
 
 def test_read_log_stream_for_action_run_min_date_and_max_date_today():
     # NOTE: these tests don't actually depend on the current time apart from
@@ -35,7 +111,7 @@ def test_read_log_stream_for_action_run_min_date_and_max_date_today():
         "tron.config.static_config.build_configuration_watcher",
         autospec=True,
     ), mock.patch(
-        "staticconf.read", autospec=True, return_value=1000
+        "staticconf.read", autospec=True, side_effect=static_conf_patch({"logging.max_lines_to_display": 1000})
     ), mock.patch(
         "tron.config.static_config.load_yaml_file",
         autospec=True,
@@ -111,7 +187,7 @@ def test_read_log_stream_for_action_run_min_date_and_max_date_different_days():
         "tron.config.static_config.build_configuration_watcher",
         autospec=True,
     ), mock.patch(
-        "staticconf.read", autospec=True, return_value=1000
+        "staticconf.read", autospec=True, side_effect=static_conf_patch({"logging.max_lines_to_display": 1000})
     ), mock.patch(
         "tron.config.static_config.load_yaml_file",
         autospec=True,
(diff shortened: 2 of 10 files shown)


tron/utils/scribereader.py
```diff
@@ -90,146 +141,108 @@ def read_log_stream_for_action_run(
...
-    lines = [line for _, line in output]
-    malformed = [f"{malformed_lines} encountered while retrieving logs"] if malformed_lines else []
+
+    # yelp_clog S3LogsReader is a newer reader that is supposed to replace scribe readers eventually.
+    if use_s3_reader:
+        end_date = max_date.date() if max_date else today
+
+        log.debug("Using S3LogsReader to retrieve logs")
+        s3_reader = S3LogsReader(ecosystem).get_log_reader(log_name=stream_name, min_date=start_date, max_date=end_date)
+        paasta_logs.iterate_logs(s3_reader, max_lines)
```
nemacysts, line 183: i wonder if we should rename `iterate_logs` as we're really fetching them - not returning an iterator

tron/utils/scribereader.py
```diff
@@ -90,146 +141,108 @@ def read_log_stream_for_action_run(
...
         ]
-    host, port = get_scribereader_host_and_port()  # type: ignore  # the None case is covered by the check above
 
-    # this should never fail since get_scribereader_host_and_port() will have also called get_superregion() and we've ensured that
-    # that file exists by getting to this point
     if paasta_cluster is None:
-        paasta_cluster = get_superregion()
+        paasta_cluster = superregion
+
+    paasta_logs = PaaSTALogsIterator(component, paasta_cluster, action_run_id)
```
nemacysts, line 171: i also wonder if we should rename this class since it's not returning an iterator

tron/utils/scribereader.py
nemacysts: also: thoughts on having this function return the log line? atm, it reads a bit weirdly since you'd expect a return value - but we're actually having the class store the log lines instead

tron/utils/scribereader.py
```diff
@@ -90,146 +141,108 @@ def read_log_stream_for_action_run(
...
+
+    # yelp_clog S3LogsReader is a newer reader that is supposed to replace scribe readers eventually.
+    if use_s3_reader:
+        end_date = max_date.date() if max_date else today
+
+        log.debug("Using S3LogsReader to retrieve logs")
+        s3_reader = S3LogsReader(ecosystem).get_log_reader(log_name=stream_name, min_date=start_date, max_date=end_date)
+        paasta_logs.iterate_logs(s3_reader, max_lines)
+    else:
+        end_date = max_date.date() if max_date else None  # type: ignore
```
nemacysts, line 185: just curious: why are we telling mypy to ignore this line?

tron/utils/scribereader.py
```diff
@@ -90,146 +141,108 @@ def read_log_stream_for_action_run(
...
+        use_tailer = today in {start_date, end_date}
+        use_reader = start_date != today and end_date is not None
+
+        host, port = get_scribereader_host_and_port(ecosystem, superregion, region)  # type: ignore  # the None case is covered by the check above
```
nemacysts, line 189: we should be able to remove this `type: ignore` by fixing the return typing of `get_scribereader_host_and_port` - generally, `type: ignore` should be a last resort :)

tron/utils/scribereader.py
```diff
@@ -90,146 +141,108 @@ def read_log_stream_for_action_run(
...
+            except socket.timeout:
+                return [
+                    f"Unable to connect to stream {stream_name} - if this is the first time this action has run and you "
+                    "expected output, please wait a couple minutes and refresh."
+                ]
+            finally:
+                stream.close()
+
+    # for logs that use Kafka topics with multiple partitions underneath or retrieved by S3LogsReader,
+    # data ordering is not guarantied - so we'll sort based on log timestamp set by producer.
```
nemacysts, line 228: nit:
```suggestion
    # data ordering is not guaranteed - so we'll sort based on log timestamp set by producer.
```

requirements.txt
nemacysts, line 1: just double-checking: does check-requirements pass locally?

because of our hacky requirements setup here, it might also be worth adding the extra yelp reqs to requirements/requirements-minimal to make sure everything is pinned :)

tron/utils/scribereader.py
```diff
@@ -59,15 +64,7 @@ def get_ecosystem() -> str:
...
-    try:
-        ecosystem = get_ecosystem()
-        superregion = get_superregion()
-        region = get_region()
-    except OSError:
-        log.warning("Unable to read location mapping files from disk, not returning scribereader host/port")
-        return None
-
+def get_scribereader_host_and_port(ecosystem, superregion, region: str) -> Optional[Tuple[str, int]]:
     # NOTE: Passing in an ecosystem of prod is not supported by scribereader
```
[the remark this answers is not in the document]
yaroliakh: still valid, as in case of using scirbereader tailler it needs to get scribe service host and port, which will raise KeyError when called with "prod" ecosystem argument.

tron/utils/scribereader.py
yaroliakh: valid point, it's definitely not an iterator but rather does the "iteration"

tron/utils/scribereader.py
yaroliakh: agree, it's not ideal but the primary reason for using class store is to avoid repeated sorting of lines when iterator is called multiple times for different readers, as in case of scribereader reader & tailler. With class store we  can call sort once after all the logs are fetched. Even though with the current `max_lines` pretty low this is not an issue at all, I was wondering if we should increase it to reduce not so rare occurrences of `This output is truncated` for certain jobs.

requirements.txt
yaroliakh: hm, `check-requirements` works locally.  
> adding the extra yelp reqs to requirements/requirements-minima

wouldn't it break github CI,as yelp internal packages won't be available?

tron/utils/scribereader.py
```diff
@@ -89,147 +134,115 @@ def read_log_stream_for_action_run(
...
+                return [
+                    f"Unable to connect to stream {stream_name} - if this is the first time this action has run and you "
+                    "expected output, please wait a couple minutes and refresh."
+                ]
+            finally:
+                stream.close()
+
+    # for logs that use Kafka topics with multiple partitions underneath or retrieved by S3LogsReader,
+    # data ordering is not guarantied - so we'll sort based on log timestamp set by producer.
+    lines = paasta_logs.sort_log_lines()
```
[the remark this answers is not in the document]
yaroliakh: yes, `scribereader` still doesn't provide sorting as both stream_reader and stream_tailer can return unordered logs when the underlying Kafka topic has more than one partition. `paasta logs` does also [sorting](https://github.com/Yelp/paasta/blob/d945ff1af51a0703379419881f9d9a2ae7c69bac/paasta_tools/cli/cmds/logs.py#L898-L901), and it's actually somehow even more similar now with `PaaSTALogs` class that will be used to fetch and filters logs. 
Thus sorting is still needed on the client side (tron, paasta, yelp_clog, etc) until we figure out something on the backend.

tron/utils/scribereader.py
yaroliakh: updated `end_date`, but before it was complaining something about type mismatch

tron/utils/scribereader.py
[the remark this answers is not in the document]
nemacysts: ah gotcha - i always assumed that paasta logs was just dumping what `scribereader` returned - TIL :p

requirements.txt
nemacysts: ah, sorry - i was thinking faster than i was typing: `it might also be worth temporarily adding the extra yelp reqs to requirements/requirements-minimal to make sure everything is pinned :)` - temporarily adding them would let you run check-requirements and make sure everything is pinned correctly :)

tron/utils/scribereader.py
nemacysts: increasing max_lines should be fine - it's more of a protection against jobs that spew gigabytes of logs and run multiple times :)

i'm hoping that at some point in the future we try to re-architect things a bit here so that we can store these logs in a more efficient way for tron to query them, but that's for future us!

nemacysts: we're also currently only sorting once in the pre-PR code though, right?

yaroliakh: right, nothing changes in this regard. With `PaastaLogs` just try to avoid extra code duplication with addition of the the third reader. I think we can simplify it once again once we remove scribe readers

nemacysts:
(also tagging some folks working on tron stuff this quarter to make sure they're aware of the logging changes)

requirements.txt
yaroliakh: I see now. yes, there are some errors, i'll need to update it

nemacysts (commented):
lgtm - i can ship once you figure out the local test issues

(i'm not super enamoured with storing the logs in the class, but my objections here are mostly 'cause it doesn't read naturally to my eyes - which isn't really a reason to block :p)

tron/utils/scribereader.py
yaroliakh: so to rephrase:
- one point for using class store is to sort only once when all the logs have been fetched 
- secondly, to mask the details of paasta logs sorting inside the class itself (as we should care only about `lines`), rather than returning `output` with a list of tuples and doing sorting and extracting of `lines` outside it

yaroliakh:
@nemacysts I was struggling a bit with mypy import errors, so decided to update mypy as it seems to support smart ignore conditions in code, but eventually ended up just adding a global ignore rules into mypy.ini. 
Hopefully having new mypy version won't be an issue, otherwise I can revert it back if needed

requirements.txt
yaroliakh: these have been updated in `yelp_package/extra_requirements_yelp.txt`

tests/utils/scribereader_test.py
```diff
@@ -3,12 +3,83 @@
...
 
 try:
     import scribereader  # noqa: F401
+    from clog.readers import S3LogsReader  # noqa: F401
 except ImportError:
-    pytest.skip("scribereader not available, skipping tests", allow_module_level=True)
+    pytest.skip("yelp logs readers not available, skipping tests", allow_module_level=True)
+
+
+def static_conf_patch(args={}):
```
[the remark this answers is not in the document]
nemacysts: (might also be worth a comment here since i'm not 100% sure why this is necessary atm)

tests/utils/scribereader_test.py
```diff
@@ -3,12 +3,83 @@
...
+    return lambda arg, namespace, default=None: args.get(arg)
+
+
+def test_read_log_stream_for_action_run_not_available():
+    tron.utils.scribereader.scribereader_available = False
+    tron.utils.scribereader.s3reader_available = False
```
nemacysts, line 22: i *think* setting these with a mock might be a tad cleaner since that would take care of setting things back without needing a finally block

nemacysts (approved):

tests/utils/scribereader_test.py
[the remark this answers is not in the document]
yaroliakh: Updated, thanks. Primarily adding this because without explicit patching of specific staticconf values and instead using `return_value=` in `unittest.mock.patch` it causes side effects when `staticconf.read` is called more than once, as in case with addition of new reader functionality.

PR #950, opened 2024-04-17
Parallelizing execution of restoring in Tron - TRON-2161

EmanElsaban:
This PR adds multiple threads to when we call batchgetitem and to when we are restoring jobs. It also changes the implementation of _restore_run_for_jobs method to get back all runs state for a job instead of getting state for each run one by one. This will allow us to make usage of the multiple threads added when calling batchGetItem 

### Testing
- Unit tests pass
- I tested locally and I see the jobs appear in the UI and in the right order
-  test by how much are we speeding the restore function execution and tron restart timing with this code
results: https://fluffy.yelpcorp.com/i/nXz1pVFtJ15TK3j05fS83hRxLVQn7mtR.html (more about experiment: https://yelp.slack.com/archives/C063D815FBK/p1714501884787069)

### DIFF
--- a/tron/core/job_scheduler.py
+++ b/tron/core/job_scheduler.py
@@ -25,7 +25,7 @@ class JobScheduler(Observer):
         self.watch(job)
 
     def restore_state(self, job_state_data, config_action_runner):
-        """Restore the job state and schedule any JobRuns."""
+        """Load the job state and schedule any JobRuns."""
         job_runs = self.job.get_job_runs_from_state(job_state_data)
         for run in job_runs:
             self.job.watch(run)
--- a/tron/core/job_scheduler.py
+++ b/tron/core/job_scheduler.py
@@ -32,7 +32,7 @@ class JobScheduler(Observer):
         self.job.runs.runs.extend(job_runs)
         log.info(f"{self} restored")
 
-        # Tron will recover any action run that has UNKNOWN status and k
+        # Tron will recover any action run that has UNKNOWN status
         # and will start connecting to task_proc
         recovery.launch_recovery_actionruns_for_job_runs(
             job_runs=job_runs,
--- a/tron/core/job_scheduler.py
+++ b/tron/core/job_scheduler.py
@@ -32,6 +32,8 @@ class JobScheduler(Observer):
         self.job.runs.runs.extend(job_runs)
         log.info(f"{self} restored")
 
+        # Tron will recover any action run that has UNKNOWN status and k
+        # and will start connecting to task_proc
         recovery.launch_recovery_actionruns_for_job_runs(
             job_runs=job_runs,
             master_action_runner=config_action_runner,
--- a/tron/serialize/runstate/dynamodb_state_store.py
+++ b/tron/serialize/runstate/dynamodb_state_store.py
@@ -9,6 +9,7 @@ import time
 from collections import defaultdict
 from collections import OrderedDict
 from typing import DefaultDict
+from typing import List
 
 import boto3  # type: ignore
 
@@ -16,7 +17,7 @@ from tron.metrics import timer
 
 OBJECT_SIZE = 400000
 MAX_SAVE_QUEUE = 500
-MAX_ATTEMPTS = 11
+MAX_ATTEMPTS = 10
 log = logging.getLogger(__name__)
 
 
@@ -50,7 +51,7 @@ class DynamoDBStateStore:
         vals = self._merge_items(first_items, remaining_items)
         return vals
 
-    def chunk_keys(self, keys: list) -> list:
+    def chunk_keys(self, keys: List[dict]) -> List[List[dict]]:
         """Generates a list of chunks of keys to be used to read from DynamoDB"""
         # have a for loop here for all the key chunks we want to go over
         cand_keys_chunks = []
@@ -64,7 +65,7 @@ class DynamoDBStateStore:
         # let's avoid potentially mutating our input :)
         cand_keys_list = copy.copy(table_keys)
         attempts_to_retrieve_keys = 0
-        while attempts_to_retrieve_keys < MAX_ATTEMPTS and len(cand_keys_list) != 0:
+        while len(cand_keys_list) != 0:
             with concurrent.futures.ThreadPoolExecutor(max_workers=5) as executor:
                 responses = [
                     executor.submit(
@@ -83,8 +84,14 @@ class DynamoDBStateStore:
             for resp in concurrent.futures.as_completed(responses):
                 items.extend(resp.result()["Responses"][self.name])
                 # add any potential unprocessed keys to the thread pool
-                if resp.result()["UnprocessedKeys"].get(self.name):
+                if resp.result()["UnprocessedKeys"].get(self.name) and attempts_to_retrieve_keys < MAX_ATTEMPTS:
                     cand_keys_list.append(resp.result()["UnprocessedKeys"][self.name]["Keys"])
+                elif attempts_to_retrieve_keys >= MAX_ATTEMPTS:
+                    failed_keys = resp.result()["UnprocessedKeys"][self.name]["Keys"]
+                    error = Exception(
+                        f"tron_dynamodb_restore_failure: failed to retrieve items with keys \n{failed_keys}\n from dynamodb\n{resp.result()}"
+                    )
+                    raise error
             attempts_to_retrieve_keys += 1
         return items
 
--- a/tron/serialize/runstate/dynamodb_state_store.py
+++ b/tron/serialize/runstate/dynamodb_state_store.py
@@ -1,4 +1,5 @@
 import concurrent.futures
+import copy
 import logging
 import math
 import os
@@ -15,6 +16,7 @@ from tron.metrics import timer
 
 OBJECT_SIZE = 400000
 MAX_SAVE_QUEUE = 500
+MAX_ATTEMPTS = 11
 log = logging.getLogger(__name__)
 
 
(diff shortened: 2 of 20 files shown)


tron/serialize/runstate/dynamodb_state_store.py
```diff
@@ -47,38 +48,42 @@ def restore(self, keys) -> dict:
...
+            # chunks of 100 keys will be in this list
+            cand_keys_chunks.append(keys[i : min(len(keys), i + 100)])
+        return cand_keys_chunks
+
+    def _get_items(self, table_keys: list) -> object:
+        items = []
+        # precompute the cand_keys and then all we gotta do is submit stuff to the thread pool using the precomputed keys
+        cand_keys_list = self.chunk_keys(table_keys)
+        count = 0
+        while count < len(cand_keys_list):
```
nemacysts, line 65: i'm not quite sure we're doing the same thing as before with the new code: if i'm reading the old code correctly, we did 10 attempts per-chunk to get all keys for a given chunk

tron/serialize/runstate/dynamodb_state_store.py
```diff
@@ -47,38 +48,42 @@ def restore(self, keys) -> dict:
...
+            with concurrent.futures.ThreadPoolExecutor(max_workers=5) as executor:
+                responses = [
+                    executor.submit(
+                        self.client.batch_get_item, RequestItems={self.name: {"Keys": key, "ConsistentRead": True}}
                     )
-                    raise error
-                else:
-                    break
+                    for key in cand_keys_list
```
nemacysts, line 71: suggestion i would renaming the variables here a bit - it took me a while to realize that `key` here is not a single key but a chunk of keys 

maybe something like:
```python
with concurrent.futures.ThreadPoolExecutor(max_workers=5) as executor:
    futures = [
        executor.submit(
            self.client.batch_get_item, RequestItems={self.name: {"Keys": chunked_keys, "ConsistentRead": True}}
        )
        for chunked_keys in self.chunk_keys(table_keys)
    ]
```

nemacysts: ah, we can't inline the self.chunk_keys() bit 'cause we update cand_keys_list below...

nemacysts: ah, actually - i think we'd have to update table_keys, but this would otherwise work the same :)

tron/serialize/runstate/statemanager.py
```diff
@@ -146,11 +150,21 @@ def restore(self, job_names, skip_validation=False):
...
-            job_state["runs"] = self._restore_runs_for_job(job_name, job_state)
+
+        # second, restore the runs for each of the jobs restored above
+        with concurrent.futures.ThreadPoolExecutor(max_workers=3) as executor:
+            # start the threads and mark each future with it's job name
+            # this is useful so that we can index the job name later to add the runs to the jobs dictionary
+            results = {
+                executor.submit(self._restore_runs_for_job, job_name, job_state): job_name
+                for job_name, job_state in jobs.items()
+            }
```
nemacysts, line 165: question: should this be a list of futures? i'm kinda surprised that this works - i would have expected `concurrent.futures.as_completed` to be unhappy with getting passed a dict

tron/serialize/runstate/statemanager.py
```diff
@@ -160,15 +174,16 @@ def restore(self, job_names, skip_validation=False):
...
+        # with self._lock:
         for run_num in run_nums:
             key = jobrun.get_job_run_id(job_name, run_num)
-            run_state = list(self._restore_dicts(runstate.JOB_RUN_STATE, [key]).values())
-            if not run_state:
-                log.error(f"Failed to restore {key}, no state found for it")
-            else:
-                runs.append(run_state[0])
+            keys_ids_list.append(key)
+        run_state = list(self._restore_dicts(runstate.JOB_RUN_STATE, keys_ids_list).values())
```
nemacysts, line 185: suggestion: 
```python
keys = [jobrun.get_job_run_id(job_name, run_num) for run_num in run_nums]
run_state = list(self._restore_dicts(runstate.JOB_RUN_STATE, keys).values())
```

(it's nice to not include the variable type in the name)

tron/serialize/runstate/statemanager.py
```diff
@@ -160,15 +174,16 @@ def restore(self, job_names, skip_validation=False):
...
+        runs.extend(run_state)
```
nemacysts, line 186: question: could we return `run_state` here?

tron/serialize/runstate/statemanager.py
```diff
@@ -14,6 +15,8 @@
 from tron.serialize.runstate.yamlstore import YamlStateStore
 from tron.utils import observer
 
+# import threading
```
nemacysts, line 18: if we go with this, reminder to remove the commented code before final review :)

nemacysts (commented):
i think the approach here makes sense - you can ignore all the comments prefixed with `question/suggestion:` for now :)

tron/serialize/runstate/dynamodb_state_store.py
EmanElsaban: yea, my code is giving one chance/attempt to get keys and unprocessed keys for a chunk. I'm not sure why we give 10 attempts? why not more and why not less?

EmanElsaban: hmm rn though if we don't get the keys by first attempt we miss on them. I'm basically looping through the list of chunks size and attempting to get key's response on the first time

nemacysts: 10 was likely picked as an arbitrary # if i had to guess :)

my comment was more around this doing `len(cand_keys_list)` attempts now (if i'm reading this right), which would be a very small number in small clusters and potentially large number in big clusters

tron/serialize/runstate/statemanager.py
EmanElsaban: yup, it works i got the inspiration from this example: https://docs.python.org/3/library/concurrent.futures.html#threadpoolexecutor-example . Basically I mark each future with its job name because it makes it easy in the code below to index the map and get the "job_name" which is later used to index "jobs" and save under each job name ["runs"], the result of this func "_restore_runs_for_job" 
```
jobs[results[result]]["runs"] = result.result()
```

tron/serialize/runstate/statemanager.py
EmanElsaban: yup, was leaving this for potential "threading.Lock()" usage

tron/serialize/runstate/statemanager.py
EmanElsaban: Nice good catch, I think I overlooked this when trying to have this new code behave the same way this func was written before

tron/serialize/runstate/dynamodb_state_store.py
```diff
@@ -47,38 +50,50 @@ def restore(self, keys) -> dict:
         vals = self._merge_items(first_items, remaining_items)
         return vals
 
-    def _get_items(self, keys: list) -> object:
-        items = []
+    def chunk_keys(self, keys: list) -> list:
```
nemacysts, line 53:
```suggestion
    def chunk_keys(self, keys: list[str]) -> list[list[str]]:
```

(i think)

tron/serialize/runstate/dynamodb_state_store.py
```diff
@@ -47,38 +50,50 @@ def restore(self, keys) -> dict:
...
+            # chunks of at most 100 keys will be in this list as there could be smaller chunks
+            cand_keys_chunks.append(keys[i : min(len(keys), i + 100)])
+        return cand_keys_chunks
+
+    def _get_items(self, table_keys: list) -> object:
+        items = []
+        # let's avoid potentially mutating our input :)
+        cand_keys_list = copy.copy(table_keys)
+        attempts_to_retrieve_keys = 0
+        while attempts_to_retrieve_keys < MAX_ATTEMPTS and len(cand_keys_list) != 0:
```
nemacysts, line 67: fwiw: we're missing the old behavior of throwing an exception after we've exhausted retries

tron/serialize/runstate/statemanager.py
```diff
@@ -146,11 +148,22 @@ def restore(self, job_names, skip_validation=False):
...
+        with concurrent.futures.ThreadPoolExecutor(max_workers=3) as executor:
+            # start the threads and mark each future with it's job name
+            # this is useful so that we can index the job name later to add the runs to the jobs dictionary
+            results = {
+                executor.submit(self._restore_runs_for_job, job_name, job_state): job_name
+                for job_name, job_state in jobs.items()
+            }
+            for result in concurrent.futures.as_completed(results):
+                jobs[results[result]]["runs"] = result.result()
+        # TODO: clean the Mesos code below
```
nemacysts, line 166: i'm pretty sure that it's safe to delete that restore :p

tron/serialize/runstate/statemanager.py
```diff
@@ -160,16 +173,20 @@ def restore(self, job_names, skip_validation=False):
...
-            else:
-                runs.append(run_state[0])
-        return runs
+        keys = [jobrun.get_job_run_id(job_name, run_num) for run_num in run_nums]
+        job_runs_restored_states = self._restore_dicts(runstate.JOB_RUN_STATE, keys)
+        run_state = copy.copy(job_runs_restored_states)
+        for key, value in run_state.items():
+            if value == {}:
+                log.error(f"Failed to restore {key}, no state found for it!")
+                job_runs_restored_states.pop(key)
```
nemacysts, line 184:
```suggestion
        for run_id, state in run_state.items():
            if state == {}:
                log.error(f"Failed to restore {run_id}, no state found for it!")
                job_runs_restored_states.pop(run_id)
```
might be a little more self-documenting (although, you'll want to verify that i'm reading this right and that these are indeed valid renames :p)

tron/serialize/runstate/statemanager.py
```diff
@@ -160,16 +173,20 @@ def restore(self, job_names, skip_validation=False):
...
+
+        run_state = list(job_runs_restored_states.values())
+        # We need to sort below otherwise the runs will not be in order
+        run_state.sort(key=lambda x: x["run_num"], reverse=True)
+        return run_state
```
nemacysts, line 189:
```suggestion
        runs = list(job_runs_restored_states.values())
        # We need to sort below otherwise the runs will not be in order
        runs.sort(key=lambda x: x["run_num"], reverse=True)
        return runs
```

would more closely match the previous naming

tron/serialize/runstate/dynamodb_state_store.py
EmanElsaban: it's a list of lists of dicts according to the stack value, will update it with that

tron/serialize/runstate/dynamodb_state_store.py
EmanElsaban: I added it in the new commit

tron/serialize/runstate/dynamodb_state_store.py
```diff
@@ -50,7 +51,7 @@ def restore(self, keys) -> dict:
         vals = self._merge_items(first_items, remaining_items)
         return vals
 
-    def chunk_keys(self, keys: list) -> list:
+    def chunk_keys(self, keys: List[dict]) -> List[List[dict]]:
```
nemacysts, line 54: non-blocking suggestion:
if we wanted to get more advanced, i think we can do some typing shenanigans so that this looks something like
```
def chunk_keys(self, keys: List[dict[K, V]]) -> List[List[dict[K, V]]]:
```
or honestly, we could be a little more generic here with something like
```
def chunk_keys(self, keys: Sequence[T]) -> List[List[T]]:
```
since this should work even with a list/tuple/etc

nemacysts (commented):
(asking internally what the rollout plan is first before approving)

tron/serialize/runstate/dynamodb_state_store.py
EmanElsaban: Will do
```
chunk_keys(self, keys: Sequence[T]) -> List[Sequence[T]]
```
cause if i do list[list[t]], I get 
```
tron/serialize/runstate/dynamodb_state_store.py: note: In member "chunk_keys" of class "DynamoDBStateStore":
tron/serialize/runstate/dynamodb_state_store.py:64:16: error: Incompatible return value type (got "List[Sequence[T]]", expected "List[List[T]]")  [return-value]
            return cand_keys_chunks
```

tron/core/job_scheduler.py
```diff
@@ -25,13 +25,15 @@ def __init__(self, job):
...
     def restore_state(self, job_state_data, config_action_runner):
-        """Restore the job state and schedule any JobRuns."""
+        """Load the job state and schedule any JobRuns."""
         job_runs = self.job.get_job_runs_from_state(job_state_data)
         for run in job_runs:
             self.job.watch(run)
         self.job.runs.runs.extend(job_runs)
         log.info(f"{self} restored")
 
+        # Tron will recover any action run that has UNKNOWN status and k
```
bobokun, line 35: Is this an unfinished sentence?

tron/serialize/runstate/dynamodb_state_store.py
```diff
@@ -7,14 +9,19 @@
...
 
 import boto3  # type: ignore
 
 from tron.metrics import timer
 
 OBJECT_SIZE = 400000
 MAX_SAVE_QUEUE = 500
+MAX_ATTEMPTS = 10
 log = logging.getLogger(__name__)
+T = TypeVar("T")
```
bobokun, line 24: What's this used for?

bobokun (commented):
just a few clarifying questions

tron/core/job_scheduler.py
EmanElsaban: good catch, will fix it now

tron/serialize/runstate/dynamodb_state_store.py
EmanElsaban: In
```
chunk_keys(self, keys: Sequence[T]) -> List[Sequence[T]]
```

KaspariK (approved):

bobokun (approved):

EvanKrall (approved):

PR #951, opened 2024-05-08
Update task_proc to better handle killing tasks

nemacysts:
We've seen that task_proc/k8s will sometimes not correctly send events for pods that we try to kill ourselves (either because the pods are already gone or because the event is somehow missing data), so this task_proc version will send synthetic events when we call kill() to ensure that tron is in the correct state :)

### DIFF
--- a/tron/kubernetes.py
+++ b/tron/kubernetes.py
@@ -413,7 +413,7 @@ class KubernetesCluster:
             # workaround logic runs but k8s sends the correct event faster than we can send our synthetic
             # one and the hackiness of this is somewhat removed by the `event.raw` check - that should only
             # exclude our synthetic event.
-            if not event.platform_type == "killed" and event.raw is None:
+            if not (event.platform_type == "killed" and event.raw is None):
                 log.warning(f"Got event for unknown task ({task_id} not in {self.tasks.keys()}): {event}")
             return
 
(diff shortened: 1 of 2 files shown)


nemacysts:
This requires https://github.com/Yelp/task_processing/pull/214

tron/kubernetes.py
```diff
@@ -404,7 +404,17 @@ def _handle_task_event(self, event: Event) -> None:
...
+            # NOTE: we don't log killed events for tasks we don't know about, as we do some slightly
+            # funky things with these events: namely, we'll send our own synthetic killed event to
+            # work around some weird k8s event behavior we've seen in the past where the coalesced
+            # event that we get in the task_processing watch loop either doesn't have the correct state
+            # or is missing entirely. This is a bit of a hack, I'm sorry :(
+            # That said, without this we'd get somewhat annoying logspam in the tron logs whenever our
+            # workaround logic runs but k8s sends the correct event faster than we can send our synthetic
+            # one and the hackiness of this is somewhat removed by the `event.raw` check - that should only
+            # exclude our synthetic event.
+            if not event.platform_type == "killed" and event.raw is None:
```
jfongatyelp, line 416:
```suggestion
            if not (event.platform_type == "killed" and event.raw is None):
```
We want to exclude this warning for our own synthetic events right? If we'd done it w/o the () we'd have excluded any real event where event.raw !=None, wouldn't we?

nemacysts: whoops, good catch!

nemacysts:
(retested after the last change: still looks good :))

jfongatyelp (approved):

PR #959, opened 2024-05-13
Delete Mesos related exit codes and docker files - TRON-2187

EmanElsaban:
Deletes Mesos related exit codes, Mesos dockerfiles and ``docker_parameters``


nemacysts (approved):

tests/config/config_parse_test.py
nemacysts, line 1: imo, we can continue asserting what we print if we'd like, but the modified tests would be more useful if they also verified that the config deletes any extra keys as well

EmanElsaban: cool, will add the behaviour testing in addition to the message assertion

nemacysts (approved):

PR #960, opened 2024-05-13
Automate make release in Tron

bobokun:
This automates the release process similar to what we have in PaaSTA and creates automated releases and changelogs


nemacysts (approved):
ty!

docs/source/conf.py
```diff
@@ -58,9 +58,10 @@ def __getattr__(cls, name):
 # built documents.
 #
 # The short X.Y version.
-version = ".".join(str(s) for s in tron.__version_info__[:2])
+version = "LATEST_VERSION"
 # The full version, including alpha/beta/rc tags.
-release = tron.__version__
+release = "LATEST_RELEASE"
```
nemacysts, line 63: do these get replaced by something?

nemacysts (approved):
probably want to make sure that the man pages still work after this, but lgtm otherwise :)

docs/source/conf.py
bobokun: hmm good question. I was looking at the [Paasta docs](https://github.com/Yelp/paasta/blob/master/docs/source/conf.py#L53) for reference. I'll look into it a bit more to see how it gets replaced (or if it does)

bobokun: lmao guess not https://github.com/Yelp/paasta/commit/52e2b713288e7fffaffb48dc88dc08e4c9dbff74

nemacysts (approved):

PR #961, opened 2024-05-13
Delete Mesos ActionRun code

nemacysts:
We don't run Mesos clusters anymore (and haven't for a while), let's clean up all this cruft.


nemacysts:
tron-infrastage seems happy with this change :)

bobokun (approved):

EmanElsaban (approved):

tron/core/actionrun.py
EmanElsaban, line 1: I think actionrun.py is the last file that imports from tron.mesos, should we also delete in this PR mesos.py and mesos_test.py

EmanElsaban (approved):

PR #963, opened 2024-05-14
TRON-2186: Fix docs build on latest release

bobokun:
This should fix the docs building on the latest version.
I've also updated the changelog to point to the latest release and removed the `CHANGELOG.md` which hasn't been updated since 2019.
The additional requirement for `requirements.txt` is necessary in order to generate the api docs
I've removed any auto generated docs and build from the repo as well (They were excluded previously in .gitignore but was missed after moving the doc files to docs/source)


tox.ini
```diff
@@ -35,13 +35,15 @@ ignore = E501,E265,E241,E704,E251,W504,E231,W503,E203
 [testenv:docs]
 deps =
     --requirement={toxinidir}/requirements-docs.txt
+    --requirement={toxinidir}/requirements-minimal.txt
```
nemacysts, line 38: requirements-minimal.txt is not meant to be installed - if we need a real venv, we probably want to install from requirements.txt instead

and just curious: what error is this fixing?

bobokun: https://github.com/Yelp/Tron/actions/runs/9070754496/job/24923133131#step:7:333

nemacysts (approved):
imo, we could also get rid of the generated code docs from here + paasta without anyone noticing, but i guess it won't hurt to have these :p

PR #969, opened 2024-05-23
Handling exceptions thrown from threads span by ThreadPoolExecutor - TRON-2202

EmanElsaban:
This PR adds a try/except block to where we call result() to catch any exceptions that might happen in the threads span by ThreadPoolExecutor.

### Testing
I tested if with this change we now catch an exception thrown by "_restore_runs_for_job" and exit the trond process. This works so I added it to the BatchGetItem call as well where we call result()

### DIFF
--- a/tests/serialize/runstate/dynamodb_state_store_test.py
+++ b/tests/serialize/runstate/dynamodb_state_store_test.py
@@ -266,3 +266,19 @@ class TestDynamoDBStateStore:
                 store.restore(keys)
             except Exception:
                 assert_equal(mock_failed_read.call_count, 11)
+
+    def test_result_exception(self, store, small_object):
+        # This test is to check if the result() function raises an exception
+        keys = [store.build_key("thing", i) for i in range(3)]
+        value = pickle.loads(small_object)
+        pairs = zip(keys, (value for i in range(len(keys))))
+        store.save(pairs)
+        store._consume_save_queue()
+
+        mock_future = mock.MagicMock()
+        mock_future.result.side_effect = Exception("mocked exception")
+        with mock.patch("concurrent.futures.Future", return_value=mock_future):
+            with mock.patch("concurrent.futures.as_completed", return_value=[mock_future]):
+                with pytest.raises(Exception) as exec_info:
+                    store.restore(keys)
+                assert str(exec_info.value) == "mocked exception"
--- a/tests/serialize/runstate/dynamodb_state_store_test.py
+++ b/tests/serialize/runstate/dynamodb_state_store_test.py
@@ -267,18 +267,14 @@ class TestDynamoDBStateStore:
             except Exception:
                 assert_equal(mock_failed_read.call_count, 11)
 
-    def test_result_exception(self, store, small_object):
-        # This test is to check if the result() function raises an exception
+    def test_restore_exception_propagation(self, store, small_object):
+        # This test is to ensure that restore propagates exceptions upwards: see DAR-2328
         keys = [store.build_key("thing", i) for i in range(3)]
-        value = pickle.loads(small_object)
-        pairs = zip(keys, (value for i in range(len(keys))))
-        store.save(pairs)
-        store._consume_save_queue()
 
         mock_future = mock.MagicMock()
         mock_future.result.side_effect = Exception("mocked exception")
-        with mock.patch("concurrent.futures.Future", return_value=mock_future):
-            with mock.patch("concurrent.futures.as_completed", return_value=[mock_future]):
+        with mock.patch("concurrent.futures.Future", return_value=mock_future, autospec=True):
+            with mock.patch("concurrent.futures.as_completed", return_value=[mock_future], autospec=True):
                 with pytest.raises(Exception) as exec_info:
                     store.restore(keys)
                 assert str(exec_info.value) == "mocked exception"
--- a/tron/serialize/runstate/dynamodb_state_store.py
+++ b/tron/serialize/runstate/dynamodb_state_store.py
@@ -85,9 +85,9 @@ class DynamoDBStateStore:
                 # request otherwise
                 cand_keys_list = []
             for resp in concurrent.futures.as_completed(responses):
-                items.extend(resp.result()["Responses"][self.name])
-                # add any potential unprocessed keys to the thread pool
                 try:
+                    items.extend(resp.result()["Responses"][self.name])
+                    # add any potential unprocessed keys to the thread pool
                     if resp.result()["UnprocessedKeys"].get(self.name) and attempts_to_retrieve_keys < MAX_ATTEMPTS:
                         cand_keys_list.append(resp.result()["UnprocessedKeys"][self.name]["Keys"])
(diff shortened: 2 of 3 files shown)


tron/serialize/runstate/dynamodb_state_store.py
```diff
@@ -87,14 +87,18 @@ def _get_items(self, table_keys: list) -> object:
...
+                    if resp.result()["UnprocessedKeys"].get(self.name) and attempts_to_retrieve_keys < MAX_ATTEMPTS:
+                        cand_keys_list.append(resp.result()["UnprocessedKeys"][self.name]["Keys"])
+                    elif attempts_to_retrieve_keys >= MAX_ATTEMPTS:
+                        failed_keys = resp.result()["UnprocessedKeys"][self.name]["Keys"]
+                        error = Exception(
+                            f"tron_dynamodb_restore_failure: failed to retrieve items with keys \n{failed_keys}\n from dynamodb\n{resp.result()}"
+                        )
+                        raise error
+                except Exception as e:
+                    log.error(repr(e))
```
nemacysts, line 100: log.exception() will let us log a human-readable message and automatically include the trace back after said message

nemacysts: it also seems like we don't need a try-except here? we're catching the exception we're throwing and then re-raising it without doing anything else

tron/serialize/runstate/statemanager.py
```diff
@@ -161,7 +161,11 @@ def restore(self, job_names, skip_validation=False):
                 for job_name, job_state in jobs.items()
             }
             for result in concurrent.futures.as_completed(results):
-                jobs[results[result]]["runs"] = result.result()
+                try:
+                    jobs[results[result]]["runs"] = result.result()
+                except Exception as e:
+                    log.error(f"Failed to restore runs for {results[result]}: {e}")
+                    raise SystemExit(f"Error: {e}")
```
nemacysts, line 168: sys.exit() would be a little clearer here :)

tron/serialize/runstate/statemanager.py
```diff
@@ -161,7 +161,11 @@ def restore(self, job_names, skip_validation=False):
                 for job_name, job_state in jobs.items()
             }
             for result in concurrent.futures.as_completed(results):
-                jobs[results[result]]["runs"] = result.result()
+                try:
+                    jobs[results[result]]["runs"] = result.result()
+                except Exception as e:
+                    log.error(f"Failed to restore runs for {results[result]}: {e}")
```
nemacysts, line 167: something like `log.exception(f"Unable to restore state for {results[result]} - exiting to avoid corrupting data.")` might be a little clearer in logs

tron/serialize/runstate/statemanager.py
nemacysts: that said, i think this means that restore_runs_for_jobs() now has some dead code? i.e., the code for processing data that's just an empty dictionary

nemacysts: or is that still possible for jobs that we actually have no saved state for at all? which i guess might be jobs that were somehow created without ever saving data back to dynamo?

tron/serialize/runstate/dynamodb_state_store.py
EmanElsaban: this is mainly for any exception that might be raised from "result()" since we have seen that if we don't have it in a try/except block then we might not see that exception

EmanElsaban: hmm I think the try should also include ``items.extend(resp.result()["Responses"][self.name])`` since we start accessing result() here

tron/serialize/runstate/statemanager.py
```diff
@@ -161,7 +163,13 @@ def restore(self, job_names, skip_validation=False):
                 for job_name, job_state in jobs.items()
             }
             for result in concurrent.futures.as_completed(results):
-                jobs[results[result]]["runs"] = result.result()
+                try:
+                    jobs[results[result]]["runs"] = result.result()
+                except Exception as e:
+                    log.exception(
+                        f"Unable to restore state for {results[result]} - exiting to avoid corrupting data. Exception: {e}"
```
nemacysts, line 170:
```suggestion
                        f"Unable to restore state for {results[result]} - exiting to avoid corrupting data."
```

log.exception will print out the exception following this log message - no need to include it again :)

tron/serialize/runstate/dynamodb_state_store.py
```diff
@@ -85,16 +85,20 @@ def _get_items(self, table_keys: list) -> object:
...
+                    if resp.result()["UnprocessedKeys"].get(self.name) and attempts_to_retrieve_keys < MAX_ATTEMPTS:
+                        cand_keys_list.append(resp.result()["UnprocessedKeys"][self.name]["Keys"])
+                    elif attempts_to_retrieve_keys >= MAX_ATTEMPTS:
+                        failed_keys = resp.result()["UnprocessedKeys"][self.name]["Keys"]
+                        error = Exception(
+                            f"tron_dynamodb_restore_failure: failed to retrieve items with keys \n{failed_keys}\n from dynamodb\n{resp.result()}"
+                        )
+                        raise error
+                except Exception as e:
+                    log.exception(f"Error: {e}")
```
nemacysts, line 100:
```suggestion
                    log.exception(f"Encountered issues retrieving data from DynamoDB")
```

`log.exception` will already include the traceback, so we should write some more human-friendly text :)

tests/serialize/runstate/dynamodb_state_store_test.py
```diff
@@ -266,3 +266,19 @@ def test_retry_reading(self, store, small_object, large_object):
                 store.restore(keys)
             except Exception:
                 assert_equal(mock_failed_read.call_count, 11)
+
+    def test_result_exception(self, store, small_object):
+        # This test is to check if the result() function raises an exception
```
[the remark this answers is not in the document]
nemacysts: (this would make it a little clearer as to what we're testing)

tron/serialize/runstate/statemanager.py
```diff
@@ -161,7 +163,11 @@ def restore(self, job_names, skip_validation=False):
                 for job_name, job_state in jobs.items()
             }
             for result in concurrent.futures.as_completed(results):
-                jobs[results[result]]["runs"] = result.result()
+                try:
+                    jobs[results[result]]["runs"] = result.result()
+                except Exception:
+                    log.exception(f"Unable to restore state for {results[result]} - exiting to avoid corrupting data.")
+                    sys.exit(ExitCode.fail)
```
nemacysts, line 170: sorry, just noticed this ExitCode import - i'd probably just do a `sys.exit(1)` since cmd_utils is ostensibly meant for CLI utils like tronctl/tronview/etc

nemacysts (commented):
sorry, i missed one thing in my previous review - lgtm otherwise tho :)

tron/serialize/runstate/statemanager.py
EmanElsaban: Np, thank youu

PR #972, opened 2024-06-04
Use S3LogsReader with superregion and UTC timezone

yaroliakh:
Updating `yelp_clog` that comes with two changes in S3LogsReader:
- usage of `superregion` as a parameter to determine the prefix from where to fetch logs in S3
- usage of UTC dates as a standard timezone for logs stored in S3


nemacysts (approved):

jvperrin (approved):
Nice!

PR #975, opened 2024-06-07
support projected service account volumes

piax93:
~The equivalent of https://github.com/Yelp/paasta/pull/3888.~

~I really don't like the way I'm resolving the service name, but digging through the call graph I don't see a good place where I could infer that without going insane, so looking for suggestions there.~

Change of plan, I'm actually going to integration the whole logic into paasta-tools, and just support passing the projected volumes from the configurations here. This way I get rid of the "name of the service" problem, and can also use a single implementation for the "does this service need the token?" logic.


danielpops:
> I really don't like the way I'm resolving the service name

This is on par with how we did it in the 2019 implementation https://github.yelpcorp.com/services/opa_policy_manager/blob/0ef9d7fc1f193ac217458272695cdda26c0d1e8b/opa_policy_manager/paasta.py#L60 , so I'd say it's fine until it's not!

danielpops (approved):

tron/utils/authentication.py
```diff
@@ -0,0 +1,87 @@
...
+    if watcher:
+        watcher.reload_if_changed()
+    return cast(
+        ProjectedSAVolume,
+        staticconf.read("service_auth_token_settings", namespace=TOKEN_VOLUME_CONF_NAMESPACE, default={}),
+    )
+
+
+def get_projected_sa_volumes(service: str) -> List[ProjectedSAVolume]:
+    """Return projected service account volume, as a single elemenet list,
```
nemacysts, line 80:
```suggestion
    """Return projected service account volume, as a single element list,
```

tron/utils/authentication.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+def get_authenticating_services() -> Set[str]:
+    """Load list of services participating in authenticated traffic
+
+    :param str soa_dir: SOA configuration directory
+    :return: set of service names
+    """
+    global CACHED_AUTHENTICATING_SERVICES
+    soa_path = staticconf.read_string("soa_path", namespace=TRON_NAMESPACE, default=DEFAULT_SOA_PATH)
```
nemacysts, line 50: just curious: any scenarios where we'd want to change this path just for some tron masters?

nemacysts: (and same with `DEFAULT_PAASTA_CONFIG_PATH`)

nemacysts (commented):
seems generally fine to me - we tend to put most config in MASTER.yaml for Tron, but that might be a pain for this since i think we'd have to plumb things quite a way through Tron for values read from there to be readable here

that said, i can ship once we confirm that `service: X`-using configs won't be a problem for this (all my other comments are non-blocking or just me being curious)

tron/core/actionrun.py
```diff
@@ -347,6 +348,10 @@ def id(self):
     def name(self):
         return self.action_name
 
+    @property
+    def service_name(self):
+        return self.job_run_id.split(".", 1)[0]
```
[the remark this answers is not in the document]
piax93: I think we can treat the job configuration as authoritative for the service identity. I.e., if a job is defined in `soa-configs/some_service/...` the identity of the job is going to be "some_service" regardless of the image which is getting used.

(and, tbf, we cannot really do differently with the current implementation)

tron/utils/authentication.py
piax93: No. It's just fear of "we changed that file location from some $reason, but Tron is currently pinned, and we are are stuck".

tron/utils/authentication.py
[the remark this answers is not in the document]
piax93: Ah, interesting, I was indeed struggling to figure out from where the job configurations were originally sourced.

I agree it's probably fine, or I would argue better, to have this part of the configuration not frozen, so that retries can keep into account changes to authorization policies.

That said, if I just add support for SA volumes here, and then move the rest of the logic in paasta-tools (or rather, rely on what's getting put in place in https://github.com/Yelp/paasta/pull/3888), I can avoid jumping through a bunch of hoops to load configs in here, so I should probably do that.

danielpopskandor (approved):

nemacysts (approved):

PR #982, opened 2024-06-25
Add clarification around reverting changes that touch DynamoDB data

nemacysts:
A revert is essentially a delete, but I don't think many of us would treat one the same as an explicit delete given that generally reverting code is a safe operation :)

### DIFF
--- a/tron/config/config_parse.py
+++ b/tron/config/config_parse.py
@@ -4,6 +4,7 @@ contain a validated configuration.
 
 
 WARNING: it is *NOT* safe to delete classes that are being validated (or their attributes) if there are any references to them in DynamoDB! (See DAR-2328)
+NOTE: this means that reverting a change that adds a new attribute is not safe :)
 """
 import datetime
 import getpass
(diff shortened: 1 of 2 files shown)


EmanElsaban (approved):

cuza (approved):

tron/config/config_parse.py
```diff
@@ -4,6 +4,7 @@
 
 
 WARNING: it is *NOT* safe to delete classes that are being validated (or their attributes) if there are any references to them in DynamoDB! (See DAR-2328)
+NOTE: this means that reverting a change that adds a new attribute is not safe :)
```
bobokun, line 7: Maybe we can also add a note that says until TRON-2200 is complete?

nemacysts: good idea - done!

bobokun (approved):

cuza (approved):

EmanElsaban (approved):

PR #985, opened 2024-07-02
TRON-2210: Add a tool that can sync from pod state to tron w/ tronctl

jfongatyelp:
This adds a pretty simple script that will:

1. search for any completed pods in the tron namespace given a KUBECONFIG
2. attempt to match those pods with any existing action runs that are in a state that can be updated
3. update the state w/ `tronctl` if the pod & tron's states do not match.

Tron will only allow us to transition from a few known states, so this script only allows using `tronctl fail` or `tronctl succeed`  if the tron action run was either lost/unknown or 'stuck' at `running` or `starting`. 

Right now, I allow passing both a tron-url + tronctl-wrapper so that this could be run from outside the tron server itself, but defaults assume we'll be running on a local tron server. 

Still adding some tests but thought I should get feedback now in case I need to shift gears. 

## Validation
Currently, one reliable way to get an actual tron action run into an `unknown` state is to restart tron w/in a short period of the pod having been started. 

Forcing a few of these cases in tron-spark-pnw-devc, the script correctly updates the 'unknown' action runs' states:
```
Action state scheduled for paasta-contract-monitor.k8s_bad.41.sleep_with_retries not modifiable, no action taken
Action state failed for paasta-contract-monitor.k8s_bad.40.sleep not modifiable, no action taken
paasta-contract-monitor.k8s_bad.40.sleep_with_retries state unknown needs updating to fail
Dry-Run: Would run ['tronctl', 'fail', 'paasta-contract-monitor.k8s_bad.40.sleep_with_retries']
Action state failed for paasta-contract-monitor.k8s_bad.39.sleep not modifiable, no action taken
paasta-contract-monitor.k8s_bad.39.sleep_with_retries state unknown needs updating to fail
Dry-Run: Would run ['tronctl', 'fail', 'paasta-contract-monitor.k8s_bad.39.sleep_with_retries']
Action state failed for paasta-contract-monitor.k8s_bad.38.sleep not modifiable, no action taken
paasta-contract-monitor.k8s_bad.38.sleep_with_retries state running needs updating to fail
Dry-Run: Would run ['tronctl', 'fail', 'paasta-contract-monitor.k8s_bad.38.sleep_with_retries']
Action state failed for paasta-contract-monitor.k8s_bad.37.sleep not modifiable, no action taken
paasta-contract-monitor.k8s_bad.37.sleep_with_retries state running needs updating to fail
Dry-Run: Would run ['tronctl', 'fail', 'paasta-contract-monitor.k8s_bad.37.sleep_with_retries']
Action state failed for paasta-contract-monitor.k8s_bad.36.sleep not modifiable, no action taken
paasta-contract-monitor.k8s_bad.36.sleep_with_retries state running needs updating to fail
Dry-Run: Would run ['tronctl', 'fail', 'paasta-contract-monitor.k8s_bad.36.sleep_with_retries']
Action state cancelled for paasta-contract-monitor.k8s_bad.35.sleep not modifiable, no action taken
```

### DIFF
--- a/tools/sync_tron_state_from_k8s.py
+++ b/tools/sync_tron_state_from_k8s.py
@@ -7,6 +7,8 @@ Update tron state from k8s api if tron has not yet updated correctly
 This will search for completed pods in the cluster specified in the kubeconfig in the `tron` namespace and use tronctl to transition any whose states do not match.
 """
 import argparse
+import base64
+import hashlib
 import logging
 import subprocess
 import sys
@@ -35,6 +37,23 @@ TRON_MODIFIABLE_STATES = [
 ]
 
 log = logging.getLogger("sync_tron_from_k8s")
+
+
+# NOTE: Copied from paasta_tools.kubernetes_tools, if it changes there it must be updated here
+def limit_size_with_hash(name: str, limit: int = 63, suffix: int = 4) -> str:
+    """Returns `name` unchanged if it's length does not exceed the `limit`.
+    Otherwise, returns truncated `name` with it's hash of size `suffix`
+    appended.
+
+    base32 encoding is chosen as it satisfies the common requirement in
+    various k8s names to be alphanumeric.
+    """
+    if len(name) > limit:
+        digest = hashlib.md5(name.encode()).digest()
+        hash = base64.b32encode(digest).decode().replace("=", "").lower()
+        return f"{name[:(limit-suffix-1)]}-{hash[:suffix]}"
+    else:
+        return name
 
 
 def parse_args():
@@ -103,7 +122,7 @@ def get_tron_state_from_api(tron_server: str, num_runs: int = 100) -> List[Dict[
     )
 
     for job in jobs:
-        # What am I doing wrong here, why do I have to append /api
+        # Update job URL to be used with API instead of web
         url = f'/api{job["url"]}'
         log.debug(f'Fetching job {job["name"]} at {url}')
         job_runs = client.job(
@@ -120,16 +139,15 @@ def get_matching_pod(action_run: Dict[str, Any], pods: Dict[str, V1Pod]) -> Opti
     action_name = action_run["action_name"]
     job_name = action_run["job_name"]
     run_num = action_run["run_num"]
-
     service, job = job_name.split(".")
-    # TODO:  how to fetch k8s shortened instance name to match labels?
     instance_name = f"{job}.{action_name}"
+    sanitized_instance_name = limit_size_with_hash(instance_name)
     matching_pods = sorted(
         [
             pod
             for pod in pods.values()
             if pod.metadata.labels["paasta.yelp.com/service"] == service
-            and pod.metadata.labels["paasta.yelp.com/instance"] == instance_name
+            and pod.metadata.labels["paasta.yelp.com/instance"] == sanitized_instance_name
             and pod.metadata.labels["tron.yelp.com/run_num"] == run_num
         ],
         # If action has retries, there will be multiple pods w/ same job_run; we only want the latest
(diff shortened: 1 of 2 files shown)


tools/sync_tron_state_from_k8s.py
```diff
@@ -0,0 +1,202 @@
...
+    )
+
+    for job in jobs:
+        # What am I doing wrong here, why do I have to append /api
+        url = f'/api{job["url"]}'
+        log.debug(f'Fetching job {job["name"]} at {url}')
+        job_runs = client.job(
+            url,
+            include_action_runs=True,
+            count=num_runs,  # TODO: fetch job run_limit and use that for count ?
```
nemacysts, line 112: not a bad idea :)

tools/sync_tron_state_from_k8s.py
```diff
@@ -0,0 +1,202 @@
...
+    # /jobs returns only the latest 5 runs, we'll need to request all runs instead ourselves
+    jobs = client.jobs(
+        include_job_runs=False,
+        include_action_runs=False,
+        include_action_graph=False,
+        include_node_pool=False,
+    )
+
+    for job in jobs:
+        # What am I doing wrong here, why do I have to append /api
```
nemacysts, line 106: not doing anything wrong - i'm not quite sure why this method doesn't automatically add in /api for us

tools/sync_tron_state_from_k8s.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+def get_matching_pod(action_run: Dict[str, any], pods: Dict[str, V1Pod]) -> Optional[V1Pod]:
+    """Given a tron action_run, try to find the right pod that matches."""
+    action_name = action_run["action_name"]
+    job_name = action_run["job_name"]
+    run_num = action_run["run_num"]
+
+    service, job = job_name.split(".")
+    # TODO:  how to fetch k8s shortened instance name to match labels?
```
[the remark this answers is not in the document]
nemacysts: i.e., copy the limit_size_with_hash code we call in https://sourcegraph.com/github.com/Yelp/paasta@7c82a62aad1e5055095a921727c17ee9079df663/-/blob/paasta_tools/tron_tools.py?L938

nemacysts (approved):

tools/sync_tron_state_from_k8s.py
jfongatyelp: Turns out we don't actually include run_limit in job/job run API responses :/  so this would need to pull in from other sources (either config api or reading the configs some other way), so I think I'm going to leave it as is for now (we can always run w/ a higher num_run if we need to).

nemacysts: yea, we can probably read from /nail/etc/services - but it's probably fine to leave as-is, i can't think of anything getting anywhere near their `run_limit` in terms of concurrently running jobs

tools/sync_tron_state_from_k8s.py
```diff
@@ -0,0 +1,220 @@
...
+    )
+
+
+def get_desired_state_from_pod(pod: V1Pod) -> str:
+    k8s_state = pod.status.phase
+    return POD_STATUS_TO_TRON_STATE.get(k8s_state, "NoMatch")
+
+
+def update_tron_from_pods(
+    jobs: List[Dict[str, Any]], pods: Dict[str, V1Pod], tronctl_wrapper: str = "tronctl", do_work: bool = True
```
nemacysts, line 168:
```suggestion
    jobs: List[Dict[str, Any]], pods: Dict[str, V1Pod], tronctl_wrapper: str = "tronctl", do_work: bool = False
```

would be nice just in case (and would match the script default)

nemacysts (approved):

tools/sync_tron_state_from_k8s.py
jfongatyelp: oops, no idea why I would default to true there, thanks for catching this!

PR #988, opened 2024-07-05
TRON-2208: Add toggle in tron config to disable retries on LOST k8s jobs

cuza:
## Summary

This PR addresses [TRON-2208](http://y/TRON-2208) by adding a toggle in the Tron configuration to disable automatic retries for Kubernetes jobs that are marked as "LOST". This is particularly useful for non-idempotent jobs where retries can be dangerous.

### Configuration
- **`tron/config/config_parse.py`**: Added `non_retryable_exit_codes` to the Kubernetes configuration schema.
- **`tron/config/schema.py`**: Updated the schema to include `non_retryable_exit_codes`.

### Core Logic
- **`tron/core/actionrun.py`**: 
  - Modified the `_exit_unsuccessful` method to check for `non_retryable_exit_codes` and skip retries if the exit code is in this list.
  - Refactored `non_retryable_exit_codes` to use a list instead of a tuple.
- **`tron/kubernetes.py`**: 
  - Updated the `handle_event` method to set the exit code to `EXIT_KUBERNETES_TASK_LOST` for lost tasks.
  - Added `non_retryable_exit_codes` to the Kubernetes configuration.
- **`tron/utils/exitcode.py`**: Added a new exit code `EXIT_KUBERNETES_TASK_LOST` with a value of `-12`.

### Tests
- **`tests/config/config_parse_test.py`**: Updated tests to include `non_retryable_exit_codes`.
- **`tests/kubernetes_test.py`**: Updated tests to check for the new exit code and ensure the correct behavior when a task is marked as lost.

### Miscellaneous
- **`.gitignore`**: Added `.fleet` to the ignore list.
- **`requirements-dev-minimal.txt`**: Added `debugpy`.
- **`requirements-dev.txt`**: Added `debugpy`.

## Motivation

Given that "LOST" means Tron has lost track of a pod it already thought it had started for a job, attempting to retry/start a replacement can be dangerous for non-idempotent jobs. In the current state, these will consume retries, but with some of our EKS migration methods, LOST tasks are more likely. Therefore, we should have a way to temporarily pause retries on these.

## Testing

- Unit tests have been updated and added to cover the new configuration option and its effects on job retry behavior.

## Related Issues

- [TRON-2208](http://y/TRON-2208): Add toggle in Tron config to disable retries on LOST k8s jobs

## Notes

- Ensure to update your Tron configuration to include the new `non_retryable_exit_codes` option if you wish to use this feature.
- This change is backward compatible; if the new configuration option is not set, the default behavior remains unchanged.

### DIFF
--- a/requirements-dev-minimal.txt
+++ b/requirements-dev-minimal.txt
@@ -1,4 +1,5 @@
 asynctest
+debugpy
 flake8
 mock
 mypy
--- a/tests/kubernetes_test.py
+++ b/tests/kubernetes_test.py
@@ -447,7 +447,6 @@ def test_handle_event_lost(mock_kubernetes_task):
         )
     )
 
-    assert mock_kubernetes_task.is_unknown
     assert mock_kubernetes_task.exit_status == exitcode.EXIT_KUBERNETES_TASK_LOST
 
 
--- a/tests/kubernetes_test.py
+++ b/tests/kubernetes_test.py
@@ -447,7 +447,7 @@ def test_handle_event_lost(mock_kubernetes_task):
         )
     )
 
-    assert mock_kubernetes_task.is_unknown
+    assert mock_kubernetes_task.exit_status == exitcode.EXIT_KUBERNETES_TASK_LOST
 
 
 def test_create_task_disabled():
--- a/tests/kubernetes_test.py
+++ b/tests/kubernetes_test.py
@@ -447,6 +447,7 @@ def test_handle_event_lost(mock_kubernetes_task):
         )
     )
 
+    assert mock_kubernetes_task.is_unknown
     assert mock_kubernetes_task.exit_status == exitcode.EXIT_KUBERNETES_TASK_LOST
 
 
--- a/tron/core/actionrun.py
+++ b/tron/core/actionrun.py
@@ -1314,9 +1314,7 @@ class KubernetesActionRun(ActionRun, Observer):
 
         return "\n".join(msgs)
 
-    def _exit_unsuccessful(
-        self, exit_status=None, retry_original_command=True, is_lost_task=False
-    ) -> Optional[Union[bool, ActionCommand]]:
+    def _exit_unsuccessful(self, exit_status=None, retry_original_command=True) -> Optional[Union[bool, ActionCommand]]:
 
         k8s_cluster = KubernetesClusterRepository.get_cluster()
         disable_retries_on_lost = False if not k8s_cluster else k8s_cluster.disable_retries_on_lost
@@ -1329,7 +1327,7 @@ class KubernetesActionRun(ActionRun, Observer):
         if self.last_attempt is not None:
             self.last_attempt.exit(exit_status)
         if self.retries_remaining is not None:
-            if disable_retries_on_lost and is_lost_task:
+            if disable_retries_on_lost and exit_status == exitcode.EXIT_KUBERNETES_TASK_LOST:
                 log.info(f"{self} skipping auto-retries due to disable_retries_on_lost being enabled.")
             else:
                 if self.retries_remaining > 0:
@@ -1345,7 +1343,7 @@ class KubernetesActionRun(ActionRun, Observer):
             return self._done("fail", exit_status)
 
     def handle_action_command_state_change(
-        self, action_command: KubernetesTask, event: str, event_data=None
+        self, action_command: ActionCommand, event: str, event_data=None
     ) -> Optional[Union[bool, ActionCommand]]:
         """
         Observe ActionCommand state changes and transition the ActionCommand state machine to a new state.
@@ -1361,7 +1359,7 @@ class KubernetesActionRun(ActionRun, Observer):
         if event == ActionCommand.EXITING:
             if action_command.exit_status is None:
                 # This is different from SSHActionRun - allows retries to happen, if configured
-                return self._exit_unsuccessful(None, is_lost_task=action_command.is_lost_task)
+                return self._exit_unsuccessful(None)
 
             if not action_command.exit_status:
                 return self.success()
--- a/tron/core/actionrun.py
+++ b/tron/core/actionrun.py
@@ -590,7 +590,9 @@ class ActionRun(Observable):
 
         return self._done("fail", exit_status)
 
-    def _exit_unsuccessful(self, exit_status=None, retry_original_command=True) -> Optional[Union[bool, ActionCommand]]:
+    def _exit_unsuccessful(
+        self, exit_status=None, retry_original_command=True, non_retryable_exit_codes=[]
+    ) -> Optional[Union[bool, ActionCommand]]:
         if self.is_done:
             log.info(
                 f"{self} got exit code {exit_status} but already in terminal " f'state "{self.state}", not retrying',
@@ -599,13 +601,16 @@ class ActionRun(Observable):
         if self.last_attempt is not None:
             self.last_attempt.exit(exit_status)
         if self.retries_remaining is not None:
-            if self.retries_remaining > 0:
-                self.retries_remaining -= 1
-                return self.restart(original_command=retry_original_command)
+            if exit_status in non_retryable_exit_codes:
+                log.info(f"{self} skipping auto-retries due to non-retryable exit code was recieved.")
             else:
-                log.info(
-                    f"Reached maximum number of retries: {len(self.attempts)}",
-                )
+                if self.retries_remaining > 0:
+                    self.retries_remaining -= 1
+                    return self.restart(original_command=retry_original_command)
+                else:
+                    log.info(
+                        f"Reached maximum number of retries: {len(self.attempts)}",
+                    )
         if exit_status is None:
             return self._done("fail_unknown", exit_status)
         else:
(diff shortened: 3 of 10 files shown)


tron/core/actionrun.py
```diff
@@ -1314,6 +1314,34 @@ def kill(self, final: bool = True) -> Optional[str]:
...
+                    self.retries_remaining -= 1
+                    return self.restart(original_command=retry_original_command)
+                else:
+                    log.info(
+                        f"Reached maximum number of retries: {len(self.attempts)}",
+                    )
+        if exit_status is None:
+            return self._done("fail_unknown", exit_status)
+        else:
+            return self._done("fail", exit_status)
```
jfongatyelp, line 1343: Might be easier for us to maintain in the long run to have a condition catch if disable_retries_on_lost & exit_status ==EXIT_KUBERNETES_TASK_LOST and then handle that case as we want, but then depend on super()._exit_unsuccessful so we don't have to recreate ActionRun's retry handling here?

Alternatively, you could add a new param to ActionRun config classes that includes the list of 'non-retryable' exit codes (if any; our default would be None/empty list), and then we'd only need to override that param for KubernetesActionRun (preferably from k8s_config as well), rather than override `_exit_unsuccessful`.  

It means we wouldn't need a special toggle in config just for 'disabling retries on LOST', the toggle would just be setting this list to be `[EXIT_KUBERNETES_TASK_LOST]` instead of `[]` for example

cuza: editing `super()._exit_unsuccessful` like this is a lot simpler to maintain, but it means that to enable the toggle we'll need to release a new tron version everywhere instead of just changing tron config. 

Is this a behavior that we want???


```python
NON_RETRYABLE_EXIT_CODES: Set[int] = {exitcode.EXIT_KUBERNETES_TASK_LOST}

    def _exit_unsuccessful(self, exit_status=None, retry_original_command=True) -> Optional[Union[bool, ActionCommand]]:
        if self.is_done:
            log.info(
                f"{self} got exit code {exit_status} but already in terminal " f'state "{self.state}", not retrying',
            )
            return None
        if self.last_attempt is not None:
            self.last_attempt.exit(exit_status)
        if self.retries_remaining is not None:
            if exit_status in NON_RETRYABLE_EXIT_CODES:
                log.info(f"{self} skipping auto-retries due to non-retryable exit code was recieved.")
            else:
                if self.retries_remaining > 0:
                    self.retries_remaining -= 1
                    return self.restart(original_command=retry_original_command)
                else:
                    log.info(
                        f"Reached maximum number of retries: {len(self.attempts)}",
                    )
        if exit_status is None:
            return self._done("fail_unknown", exit_status)
        else:
            return self._done("fail", exit_status)
```

jfongatyelp: >  it means that to enable the toggle we'll need to release a new tron version everywhere instead of just changing tron config.

Yup we definitely don't want that!

That's where i was suggesting
> and then we'd only need to override that param for KubernetesActionRun (**preferably from k8s_config as well**)

Or this config doesn't even have to be k8s-specific; we can just have a full list and it can contain any exit codes (meaning we could combine from different executor-specific codes too, if we wanted), then the top level ActionRun would pull from that config.

nemacysts: i'm a fan of the non-retryable exit code list - it seems like it'd be more usable past this migration and should hopefully be less code too

nemacysts:
sidenote: can we add some additional context to the PR description? it can still reference the ticket, but having more context in `git blame` (without having to switch to a browser and go to jira) is always nice :)

tron/core/actionrun.py
nemacysts: (we can also use srv-configs for this list - it doesn't necessarily have to be sourced from MASTER.yaml :))

cuza:
> sidenote: can we add some additional context to the PR description? it can still reference the ticket, but having more context in `git blame` (without having to switch to a browser and go to jira) is always nice :)

yeah of course, this is just a draft PR to make sure that we agreed on the implementation 😄

tests/kubernetes_test.py
```diff
@@ -447,7 +447,7 @@ def test_handle_event_lost(mock_kubernetes_task):
         )
     )
 
-    assert mock_kubernetes_task.is_unknown
+    assert mock_kubernetes_task.exit_status == exitcode.EXIT_KUBERNETES_TASK_LOST
```
jfongatyelp, line 450: You might want to keep the old assert as well, since we do still want to verify 'lost' k8s events lead to marking the task as `unknown`.

tron/core/actionrun.py
```diff
@@ -599,13 +601,16 @@ def _exit_unsuccessful(self, exit_status=None, retry_original_command=True) -> O
         if self.last_attempt is not None:
             self.last_attempt.exit(exit_status)
         if self.retries_remaining is not None:
-            if self.retries_remaining > 0:
-                self.retries_remaining -= 1
-                return self.restart(original_command=retry_original_command)
+            if exit_status in non_retryable_exit_codes:
+                log.info(f"{self} skipping auto-retries due to non-retryable exit code was recieved.")
```
jfongatyelp, line 605: not a blocker, extremely minor rewording for readability
```suggestion
                log.info(f"{self} skipping auto-retries, received non-retryable exit code ({exit_status}).")
```

tron/core/actionrun.py
```diff
@@ -1314,6 +1319,19 @@ def kill(self, final: bool = True) -> Optional[str]:
 
         return "\n".join(msgs)
 
+    def _exit_unsuccessful(
+        self, exit_status=None, retry_original_command=True, non_retryable_exit_codes=()
+    ) -> Optional[Union[bool, ActionCommand]]:
+
+        k8s_cluster = KubernetesClusterRepository.get_cluster()
+        non_retryable_exit_codes = () if not k8s_cluster else k8s_cluster.non_retryable_exit_codes
```
jfongatyelp, line 1327: Same note here about the tuple

```suggestion
        non_retryable_exit_codes = [] if not k8s_cluster else k8s_cluster.non_retryable_exit_codes
```

tron/core/actionrun.py
```diff
@@ -590,7 +590,9 @@ def fail(self, exit_status=None):
 
         return self._done("fail", exit_status)
 
-    def _exit_unsuccessful(self, exit_status=None, retry_original_command=True) -> Optional[Union[bool, ActionCommand]]:
+    def _exit_unsuccessful(
+        self, exit_status=None, retry_original_command=True, non_retryable_exit_codes=()
```
jfongatyelp, line 594:
```suggestion
        self, exit_status=None, retry_original_command=True, non_retryable_exit_codes=[]
```

Since we're validating that non_retryable_exit_codes is a list w/ build_list_of_type... in our validator, we should probably set the default here to empty list instead of empty tuple? In practice, would probably have no effect but would help future us when debugging and prevent us wondering when we turned this into a tuple :p

tron/kubernetes.py
```diff
@@ -281,10 +282,12 @@ def __init__(
         default_volumes: Optional[List[ConfigVolume]] = None,
         pod_launch_timeout: Optional[int] = None,
         watcher_kubeconfig_paths: Optional[List[str]] = None,
+        non_retryable_exit_codes: Optional[Tuple[int, ...]] = (),
```
jfongatyelp, line 285: I dont think we really _want_ this to be a Tuple; the validator validates that we have a _list_ of this type instead. 

If your mypy was telling you you needed this to be a Tuple instead, it is likely simply because you are setting defaults everywhere to tuples rather than the empty list :D

tron/kubernetes.py
```diff
@@ -621,6 +624,7 @@ def recover(self, task: KubernetesTask) -> None:
 class KubernetesClusterRepository:
     # Kubernetes config
     kubernetes_enabled: bool = False
+    kubernetes_non_retryable_exit_codes: Tuple[int, ...] = ()
```
jfongatyelp, line 627: Again, I think we want a list.

Additionally, I think it might be nicer to allow this as `Optional` in the same way our other configs are set here (tho this would require us to set a default properly if we get `None` in `KubernetesCluster` above, similar to how default_volumes/pod_launch_timeout are set)

jfongatyelp (commented):
Looks like a nice option, would be nice to see a test where k8s_options.non_retryable_exit_codes was non-empty and checking our code does the right thing tho

tests/kubernetes_test.py
cuza: sadly the way we're checking unknow is by checking exit code 

``` python
  def is_unknown(self):
      return self.exit_status is None
```

tron/core/actionrun.py
cuza: I did initially a list but changed into a tuple since build_list_of_type returns tuple

https://github.com/Yelp/Tron/blob/b8e91c32c8f78e7b6788da98cd7f79e1d3b098c7/tron/config/config_utils.py#L193

tests/kubernetes_test.py
jfongatyelp: Ohh interesting; i didn't realize assigning the LOST's an exit code changed this behavior, thanks. 

Did you happen to glance through all the other places where we use `is_unknown`/unknown to check this change won't change any expected behavior there?

tron/core/actionrun.py
jfongatyelp: Hrm, that **is** confusing since we also use `build_list_of_type_validator` for e.g. `secret_volumes`, and then type that as a `List` in https://github.com/Yelp/Tron/blob/master/tron/core/action.py#L36 !

jfongatyelp: OK, reading the build_list_of_type cmd again: the `validator()` will return a tuple of validators, but the actual _values_ it is validating should be lists, and it validates it as such here: https://github.com/Yelp/Tron/blob/b8e91c32c8f78e7b6788da98cd7f79e1d3b098c7/tron/config/config_utils.py#L189C15-L189C25 .  

Therefore, our default values for these config options should be lists, as I understand it.

tests/kubernetes_test.py
cuza: It should not be a problem since we're only marking lost jobs with the special lost exit code

tron/core/actionrun.py
```diff
@@ -599,13 +601,16 @@ def _exit_unsuccessful(self, exit_status=None, retry_original_command=True) -> O
         if self.last_attempt is not None:
             self.last_attempt.exit(exit_status)
         if self.retries_remaining is not None:
-            if self.retries_remaining > 0:
-                self.retries_remaining -= 1
-                return self.restart(original_command=retry_original_command)
+            if exit_status in non_retryable_exit_codes:
+                log.info(f"{self} skipping auto-retries, received non-retryable exit code ({exit_status}).")
```
nemacysts, line 605: @cuza think we should set `self.retries_remaining = 0` here as well? i was making sure that we didn't have a getter/property for retries_remaining that we used elsewhere and noticed that `retries_remaining` ends up getting stored in the serialized data that we use for persistence - i'm not sure what happens on a restart with that atm though

tests/kubernetes_test.py
nemacysts: [is_broken()](https://sourcegraph.yelpcorp.com/Yelp/Tron@d51984ef27faf1d743881a8c578f02e1ddd60c72/-/blob/tron/core/actionrun.py?L724) uses the `is_unknown` property and it looks like there's some [jobrun](https://sourcegraph.yelpcorp.com/Yelp/Tron@d51984ef27faf1d743881a8c578f02e1ddd60c72/-/blob/tron/core/jobrun.py?L259) code that uses `is_broken` to gate starting new action runs - but i'm not sure if this change impacts that in a meaningful way yet

nemacysts (approved):
i don't think there's any major blockers - it's probably fine to run this in non-prod for a bit

i'm not 100% sure about my 2 non-requirements.txt comments, but we should be able to sus any issues out by actually running this since the case we're handling here is pretty rare anyway :)

requirements-dev-minimal.txt
```diff
@@ -1,4 +1,5 @@
 asynctest
+debugpy
```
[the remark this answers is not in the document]
cuza: VSCode Python Debugguer :) 

https://github.com/microsoft/debugpy

tron/core/actionrun.py
cuza: Oh that's not a bad idea. I initially made `retries_remaining` but deleted cuz I thought it was overkill. I was not thinking about what would happen if we restart tho

nemacysts (approved):

PR #989, opened 2024-07-09
TRON-2195: Support watcher_kubeconfig_paths

jfongatyelp:
Requires https://github.com/Yelp/task_processing/pull/213 & new version of taskproc to be released.

This adds the support to tron's MASTER config's `k8s_options` to allow the new `old_kubeconfig_paths` option and passes it along to task_processing's kubernetes_pod_executor.

### DIFF
--- a/requirements-minimal.txt
+++ b/requirements-minimal.txt
@@ -21,7 +21,7 @@ pytimeparse
 pytz
 PyYAML>=5.1
 requests
-task_processing[mesos_executor,k8s]>=1.1.0
+task_processing[mesos_executor,k8s]>=1.2.0
 Twisted>=19.7.0
 urllib3>=1.24.2
 Werkzeug>=0.15.3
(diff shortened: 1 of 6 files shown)


cuza:
don't we need to add validation and a default value for this in `tron/config/config_parse.py`?

jfongatyelp:
> don't we need to add validation and a default value for this in `tron/config/config_parse.py`?

Thanks! That's probably a good call to maintain parity w/ all the rest of our configs.  That said I've always been a bit confused why we specify defaults both in config_parse and in the schema classes themselves. Currently this still works fine w/o errors (verified on one of our test tron servers w/ setting + unsetting `watcher_kubeconfig_paths` and updating via paasta_setup_tron_namespace) since we are just using the ConfigKubernetes schema class I've updated in config_parse, and so it gets the default I've specified there.

nemacysts (approved):

nemacysts (approved):

PR #991, opened 2024-07-13
Support multiple clusters in sync_tron_from_k8s

jfongatyelp:
If an action supports retries then we migrate to a new cluster while a job is ongoing, some attempts for this action could live on the old cluster and others on the new. 

In order to catch this situation, sync_tron_from_k8s needs to be able to inspect pods in both old & new clusters to determine the correct tron state. 

This simply allows specifying multiple kubeconfigs and gathering _all_ pods from all clusters in order to compare against potentially-updateable actions.

This has been verified to work with multiple kubeconfigs for both eks and non-eks infrastage, and I've verified the metadata.creationTimestamps appear to be in UTC so will be consistent across clusters so we are comparing timestamps appropriately. 

(Note that this means running from the devbox is not quite as convenient since our kubeconfigs are shared in a single file; I figured just forcing folks to run this somewhere with multiple files [such as the servers themselves, or by asking folks to extract each cluster to its own kubeconfig] would be easier than adding some custom method of specifying file+context combinations in the args)

### DIFF
--- a/tools/sync_tron_state_from_k8s.py
+++ b/tools/sync_tron_state_from_k8s.py
@@ -91,7 +91,7 @@ def parse_args():
 
     # We can only have multiple kubeconfigs, or multiple contexts with a single config
     if len(args.kubeconfig_path) > 1 and args.kubecontext:
-        parser.error("You can only specify a single --kubeconfig-path if specifying multiple --kubecontext arguments.")
+        parser.error("You can only specify a single --kubeconfig-path if specifying --kubecontext arguments.")
 
     # tron's base level is critical, not info, adjust accoringly
     if args.verbose:
--- a/tools/sync_tron_state_from_k8s.py
+++ b/tools/sync_tron_state_from_k8s.py
@@ -58,7 +58,12 @@ def limit_size_with_hash(name: str, limit: int = 63, suffix: int = 4) -> str:
 
 def parse_args():
     parser = argparse.ArgumentParser()
-    parser.add_argument("--kubeconfig-path", dest="kubeconfig_path", help="KUBECONFIG path")
+    parser.add_argument(
+        "--kubeconfig-path",
+        dest="kubeconfig_path",
+        help="KUBECONFIG path; multiple can be specified to find pods in multiple clusters",
+        nargs="+",
+    )
     parser.add_argument(
         "--do-work",
         dest="do_work",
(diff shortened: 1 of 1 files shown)


nemacysts (approved):
we could potentially hardcode our paasta.conf kubeconfig (we puppet that everywhere, right?) and then we only need to have folks specify a context name, but this is perfectly fine :)

jfongatyelp:
> we could potentially hardcode our paasta.conf kubeconfig (we puppet that everywhere, right?) and then we only need to have folks specify a context name, but this is perfectly fine :)

Nah, I don't think we puppet paasta.conf everywhere, just devboxes. But that said, it seems simple enough to have 2 kind of 'modes', so I've added another arg so folks can _either_ provide a single kubeconfig + multiple kubecontexts (paasta.conf) or multiple kubeconfigs w/ no kubecontext. 

This was verified to work and return the same results w/ 2 separate infrastage+eks-infrastage kubeconfigs as well as using paasta.conf w/ 2 separate kubecontexts.

tools/sync_tron_state_from_k8s.py
```diff
@@ -77,6 +89,10 @@ def parse_args():
     parser.add_argument("-v", "--verbose", dest="verbose", action="store_true", default=False, help="Verbose logging")
     args = parser.parse_args()
 
+    # We can only have multiple kubeconfigs, or multiple contexts with a single config
+    if len(args.kubeconfig_path) > 1 and args.kubecontext:
+        parser.error("You can only specify a single --kubeconfig-path if specifying --kubecontext arguments.")
```
nemacysts, line 94: i'm assuming this will also do a sys.exit() or something similar?

tools/sync_tron_state_from_k8s.py
```diff
@@ -97,7 +113,10 @@ def parse_args():
     return args
 
 
-def fetch_pods(kubeconfig_path: str) -> Dict[str, V1Pod]:
+def fetch_pods(kubeconfig_path: str, kubecontext: Optional[str]) -> Dict[str, V1Pod]:
+    if kubecontext:
+        # KubeClient only uses the environment variable
```
nemacysts, line 118: whoops, sorry - i should have also allowed for passing this through as an arg

nemacysts (approved):

tools/sync_tron_state_from_k8s.py
jfongatyelp: Yup that's been my experience, it's using https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.error : 
```
$ ./tools/sync_tron_state_from_k8s.py --kubeconfig-path /etc/kubernetes/paasta.conf kubeconfig_eks_infrastage.conf --kubecontext pnw-devc infrastage
usage: sync_tron_state_from_k8s.py [-h] [--kubeconfig-path KUBECONFIG_PATH [KUBECONFIG_PATH ...]] [--kubecontext [KUBECONTEXT [KUBECONTEXT ...]]] [--do-work] [--tron-url TRON_URL]
                                   [--tronctl-wrapper TRONCTL_WRAPPER] [-n NUM_RUNS] [-v]
sync_tron_state_from_k8s.py: error: You can only specify a single --kubeconfig-path if specifying --kubecontext arguments.
$ echo $?
2
```

PR #995, opened 2024-09-05
Fix negative value check for non-retryable exit codes in Kubernetes configuration

cuza:
This pull request fixes a bug in the Kubernetes configuration where a negative value was not properly checked for non-retryable exit codes.


nemacysts (approved):
might be nice to add a test in `tests/config/config_parse_test.py` - but this code is pretty simple

jfongatyelp:
> might be nice to add a test in `tests/config/config_parse_test.py` - but this code is pretty simple

++, would definitely be nice

jfongatyelp (approved):

PR #997, opened 2024-09-17
Write JSON state

KaspariK:
## What
We store job and job run state as pickles. We would like to not do that. This is part of the path to not doing that by writing state data as JSON alongside our current pickles. Restoring from JSON will follow in another PR.

## Why
In DAR-2328, the removal of the mesos-related code led to resetting of tron jobs state resulting in job runs starting from "0". The pickles weren't being unpickled correctly as this relies on classes that were deleted in the mesos code.


tron/core/actionrun.py
```diff
@@ -279,32 +302,36 @@ class ActionRun(Observable):
...
+        command_config: action.ActionCommandConfig,
+        parent_context: Optional[CommandContext] = None,
+        output_path: Optional[filehandler.OutputPath] = None,
+        cleanup: bool = False,
+        start_time: Optional[datetime.datetime] = None,
+        end_time: Optional[datetime.datetime] = None,
+        run_state: str = SCHEDULED,
+        exit_status: Optional[int] = None,
+        attempts: Optional[list] = None,  # TODO: list of...ActionCommandConfig?
+        action_runner: Optional[Union[NoActionRunnerFactory, SubprocessActionRunnerFactory]] = None,
```
nemacysts, line 317: maybe one day we'll either remove `NoActionRunnerFactory` (i don't think we use it outside of tests) or we'll add an abstract base class or something with the expected interface so that we don't need a union here :p

tron/actioncommand.py
```diff
@@ -195,6 +201,15 @@ def __eq__(self, other):
...
         return not self == other
 
+    @staticmethod
+    def to_json(state_data: dict) -> str:
+        return json.dumps(
+            {
+                "status_path": state_data["status_path"],
+                "exec_path": state_data["exec_path"],
+            }
+        )
```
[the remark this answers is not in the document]
nemacysts: oh, i see - a lot of these have `state_data()` properties! i think that  that makes this less pressing since the calls will (i assume) look like `SomeClass.to_json(some_object.state_data)

but it might still be nicer to switch since for things that do implement state_data() this would look like
```suggestion
    def to_json(self) -> str:
        state_data = self.state_data
        
        return json.dumps(
            {
                "status_path": state_data["status_path"],
                "exec_path": state_data["exec_path"],
            }
        )
```


and for classes that don't have a `state_data` property we'd be in the scenario described above and would benefit from this no longer being a staticmethod :)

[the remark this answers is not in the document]
nemacysts: (another benefit would also be avoiding any potential circular imports in the future since we won't need to import classes just to call to_json() :)

tron/serialize/runstate/dynamodb_state_store.py
```diff
@@ -134,6 +154,7 @@ def _merge_items(self, first_items, remaining_items) -> dict:
         return deserialized_items
 
     def save(self, key_value_pairs) -> None:
+        log.debug(f"Adding to save queue: {key_value_pairs}")
```
nemacysts, line 157: would this be too noisy in production?

tron/actioncommand.py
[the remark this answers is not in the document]
nemacysts: oh, i see - when this gets called in tron/serialize/runstate/dynamodb_state_store.py we just have a key and a dict ;_;

tron/serialize/runstate/dynamodb_state_store.py
```diff
@@ -152,19 +181,20 @@ def _consume_save_queue(self):
...
-                    key, val = self.save_queue.popitem(last=False)
+                    key, (pickled_val, json_val) = self.save_queue.popitem(last=False)
+                log.debug(f"Processing save for {key} with a value of {pickled_val}")
                 # Remove all previous data with the same partition key
                 # TODO: only remove excess partitions if new data has fewer
                 self._delete_item(key)
-                if val is not None:
-                    self[key] = pickle.dumps(val)
+                if pickled_val is not None:
+                    self.__setitem__(key, pickle.dumps(pickled_val), json_val)
```
nemacysts, line 190: just curious: how come we need `__setitem__` now?

nemacysts: ah, i see - i think we probably want to keep the existing signature since `__setitem__` is an established "dunder" method in python

what we'd probably need to do here instead is to type that the value for `__setitem__` is going to be a tuple :)

tron/actioncommand.py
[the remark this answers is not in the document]
nemacysts: hmm, maybe this is better for a post-pickle-deletion refactor where we update the simplified code to pass around the actual objects rather than state_data dicts

tron/serialize/runstate/dynamodb_state_store.py
```diff
@@ -173,6 +203,17 @@ def _consume_save_queue(self):
         else:
             self.save_errors = 0
 
+    def get_type_from_key(self, key: str) -> str:
+        return key.split()[0]
+
+    def _serialize_item(self, key: str, state: Dict[str, Any]) -> str:
```
nemacysts, line 209: if this only accepts two values for `key`, i think we can write `key: Literal[runstate.JOB_STATE, runstate.JOB_RUN_STATE]` rather than `key: str`

tron/utils/crontab.py
```diff
@@ -37,34 +42,38 @@ class FieldParser:
...
+        sorted_groups = sorted(groups, key=lambda x: (isinstance(x, str), x))
         if has_last:
-            groups.append("LAST")
-        return groups
+            sorted_groups.append("LAST")
+        if not sorted_groups:
+            return None
+        if all(isinstance(x, int) for x in sorted_groups):
+            return sorted_groups  # type: ignore
+        return sorted_groups  # type: ignore
```
nemacysts, line 68: could we add comments for these ignores?

tron/serialize/runstate/dynamodb_state_store.py
```diff
@@ -212,19 +264,28 @@ def __setitem__(self, key: str, val: bytes) -> None:
...
+                                index * OBJECT_SIZE : min(index * OBJECT_SIZE + OBJECT_SIZE, len(pickled_val))
+                            ],
                         },
                         "num_partitions": {
                             "N": str(num_partitions),
                         },
+                        "json_val": {
+                            "S": json_val[index * OBJECT_SIZE : min(index * OBJECT_SIZE + OBJECT_SIZE, len(json_val))]
+                        },
+                        "num_json_val_partitions": {
```
nemacysts, line 277: i always thought that this num_partitions stuff was something that dynamo required 🤣 - i guess this is just for our usage (to know how many partitions we're using per-item?)?

nemacysts (commented):
ty for adding all these types as well!

tron/serialize/runstate/dynamodb_state_store.py
KaspariK: Yeah, exactly. We have key+index for our compound key, but we use num_partitions in a funny multi-step restore thing where we get all the first partitions, then for every one of those we try _get_remaining_partitions (which annoyingly will need to handle JSON having more partitions).

I debated at least adding a check before this to see if num_partitions was greater than 1, but just never got around to it. This might be worth a refactor after we've depickled things.

tron/serialize/runstate/dynamodb_state_store.py
KaspariK: I did have that, but was getting `Variable not allowed in type expression`. I guess in this case I should have done `Literal[job_state", "job_run_state"]` or maybe just pretended it's an enum and added `# type: ignore`?

tron/serialize/runstate/dynamodb_state_store.py
KaspariK: Err probably. I added a few of these while testing because our current logging only has a generic `_consume_save_queue saved X items in Ys`, I don't think I really intended these to make it into the PR 😐. Removed.

tron/core/actionrun.py
KaspariK: Added a ticket for some noble soul in a future unhack 😛

tron/utils/crontab.py
KaspariK: Added types. This file was one of those dominoes that came up when adding types elsewhere. I tried to use my whole brain to do this one because it's kind of a spooky change, but I think I got it. Infrastage testing looks good, but it's not exhaustive

tron/utils/crontab.py
nemacysts, line 1: we talked about this a bit in our 1:1, but i think it's fine to type: ignore things a bit and deal with fixing any type issues here for a later PR if we're not feeling particularly confident in these changes :)

tron/actioncommand.py
```diff
@@ -195,6 +202,15 @@ def __eq__(self, other):
     def __ne__(self, other):
         return not self == other
 
+    @staticmethod
+    def to_json(state_data: dict) -> str:
```
nemacysts, line 206: from our 1:1: should we (or do we need to) handle failures in to_json? e.g., non-serializable data, missing data in the state_data dict leading to a KeyError?

if we want the json writing to be optional until we're reading from the json fields, we probably also want to make sure that json serialization failures don't result in losing pickle'd writes :)

nemacysts (commented):
generally lgtm - we can hopefully refactor things a bit to make this even nicer after all the pickle logic is gone :p

i think my only remaining question is do we need to handle json failures in any specific way

(and i'll leave it up to you whether or not you'd like to merge the crontab changes)

tron/utils/crontab.py
KaspariK: For the sake of posterity I kept this, but tried to maintain the logic as much as possible and added a few tests for good measure.

tron/actioncommand.py
```diff
@@ -195,6 +203,22 @@ def __eq__(self, other):
...
+                {
+                    "status_path": state_data["status_path"],
+                    "exec_path": state_data["exec_path"],
+                }
+            )
+        except KeyError as e:
+            log.error(f"Missing key in state_data: {e}")
+            return None
+        except Exception as e:
+            log.error(f"Error serializing SubprocessActionRunnerFactory to JSON: {e}")
```
nemacysts, line 219: doing something like the below suggestion here and in the other to_json()
```suggestion
        except KeyError:
            log.exception(f"Missing key in state_data:")
            return None
        except Exception:
            log.exception(f"Error serializing SubprocessActionRunnerFactory to JSON:")
```
might be nice since it'll include the full traceback

tron/serialize/runstate/dynamodb_state_store.py
```diff
@@ -152,19 +177,20 @@ def _consume_save_queue(self):
...
                 if val is not None:
-                    self[key] = pickle.dumps(val)
+                    self[key] = (pickle.dumps(val), json_val)
                 # reset errors count if we can successfully save
                 saved += 1
             except Exception as e:
-                error = "tron_dynamodb_save_failure: failed to save key" f'"{key}" to dynamodb:\n{repr(e)}'
+                error = "tron_dynamodb_save_failure: failed to save key " f'"{key}" to dynamodb:\n{repr(e)}'
                 log.error(error)
+                # Add items back to the queue if we failed to save
```
nemacysts, line 191: thoughts on adding back to the save queue with just the pickled data? if there's any issues writing the json, a second json-less attempt might still succeed right?

nemacysts (approved):
i have one last question/thought around the save queue handling, but otherwise i think this is good to start testing in our non-prod envs

we should probably setup some log-based alerting (or maybe use the prometheus integration?) for whenever json writing errors happen since they're definitely things we'll want to handle ASAP as part of the pickle-removal process

tron/serialize/runstate/dynamodb_state_store.py
KaspariK: It's a solid idea. I don't think we see `tron_dynamodb_save_failure` very often so it's a reasonable fallback. Would be nice to catch any partitioning issues.

KaspariK:
> we should probably setup some log-based alerting (or maybe use the prometheus integration?) for whenever json writing errors happen since they're definitely things we'll want to handle ASAP as part of the pickle-removal process

When you say prom integration are you thinking some counter like json_serialization_errors.inc() on failed serialization, then alerting on that?

nemacysts:
> When you say prom integration are you thinking some counter like json_serialization_errors.inc() on failed serialization, then alerting on that?

@KaspariK yup - exactly!

nemacysts (approved):

PR #1002, opened 2024-10-21
Use logreader rathern than vector-reader

nemacysts:
This CLI is being renamed, so let's account for that


jvperrin (approved):
Thanks!

yalinglee (approved):
Thank you!

PR #1003, opened 2024-10-24
Add TSC support to Tron

nemacysts:
This requires a task_processing bump to allow for starting pods with Topology Spread Constraints (TSCs) as well as a PaaSTA one (that will be merged after this is fully deployed) to start configuring this.

NOTE: since this is adding new data that will be serialized, reverting this will not be as straightforward as reverting this commit: we would need to leave the config classes laying around so that we can still deserialize the pickles

abridged yaml of a pod launched with this: https://fluffy.yelpcorp.com/i/BV1gwjFrFGFn88cR8SPmpMwspTlmxGmj.html


nemacysts:
requires https://github.com/Yelp/task_processing/pull/220

jfongatyelp (approved):

requirements-minimal.txt
nemacysts, line 1: this change really shouldn't be necessary (and check-requirements is fine internally without this) - but i couldn't figure out what's going on 

(and in any case, hopefully we can de-pickle things in the near future and actually delete all the unused mesos code)

jfongatyelp (approved):

EmanElsaban (approved):

PR #1004, opened 2024-10-31
Use total_seconds for timedeltas. Log job and jobrun names/runnums on serialization

KaspariK:
I think total_seconds is the easiest to convert back to timedelta later. I added some logging that might be noisier than we want longterm, but should make debugging these a lot easier.

Recreated the errors in infrastage and this does appear to resolve things.

Sidenote, I'm happy to leave this unmerged while doing more testing in dev/stage to maybe minimize the number of patch releases.

### DIFF
--- a/tron/core/actionrun.py
+++ b/tron/core/actionrun.py
@@ -761,7 +761,7 @@ class ActionRun(Observable, Persistable):
                     "attempts": [ActionRunAttempt.to_json(attempt) for attempt in state_data["attempts"]],
                     "retries_remaining": state_data["retries_remaining"],
                     "retries_delay": state_data["retries_delay"].total_seconds()
-                    if state_data["retries_delay"]
+                    if state_data["retries_delay"] is not None
                     else None,
                     "action_runner": action_runner_json,
                     "executor": state_data["executor"],
--- a/tron/core/actionrun.py
+++ b/tron/core/actionrun.py
@@ -740,6 +740,7 @@ class ActionRun(Observable, Persistable):
     @staticmethod
     def to_json(state_data: dict) -> Optional[str]:
         """Serialize the ActionRun instance to a JSON string."""
+
         action_runner = state_data.get("action_runner")
         if action_runner is None:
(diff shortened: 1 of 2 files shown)


tron/core/actionrun.py
```diff
@@ -759,7 +760,9 @@ def to_json(state_data: dict) -> Optional[str]:
                     "exit_status": state_data["exit_status"],
                     "attempts": [ActionRunAttempt.to_json(attempt) for attempt in state_data["attempts"]],
                     "retries_remaining": state_data["retries_remaining"],
-                    "retries_delay": state_data["retries_delay"],
+                    "retries_delay": state_data["retries_delay"].total_seconds()
+                    if state_data["retries_delay"]
+                    else None,
```
nemacysts, line 765:
```suggestion
                    "retries_delay": state_data["retries_delay"].total_seconds()
                    if state_data["retries_delay"] is not None
                    else None,
```

i think we'd want something like the above - unless there's no difference between None and 0 when deserializing back into a timedelta later :)

nemacysts (approved):
i think it's fine to have multiple patch releases - it'll make it easier to track the impact of changes while we iterate in non-prod

tron/core/actionrun.py
KaspariK: Ah, true. I think it's ultimately the same in restart, but there is indeed a semantic difference.

tron/serialize/runstate/dynamodb_state_store.py
```diff
@@ -204,8 +204,10 @@ def get_type_from_key(self, key: str) -> str:
     def _serialize_item(self, key: Literal[runstate.JOB_STATE, runstate.JOB_RUN_STATE], state: Dict[str, Any]) -> Optional[str]:  # type: ignore
         try:
             if key == runstate.JOB_STATE:
+                log.info(f"Serializing Job: {state.get('job_name')}")
```
KaspariK, line 207: https://www.youtube.com/watch?v=us3glFag2C4

PR #1013, opened 2024-12-20
Try to load config_name_mapping from disk just once

nemacysts:
There's enough files (and enough YAML in these files) that the IO/YAML parsing takes a significant amount of time.

While it's nice to always load from the source-of-truth (i.e., the files on-disk) - it's not worth paying the performance penalty (especially at the scale we're seeing internally) for a negligible benefit.

Locally (with a large test config), this results in a ~5x improvement in timings. More concretely, my test configs were taking ~30ish seconds *without* this diff and ~6ish seconds *with* this diff.

### DIFF
--- a/tron/config/manager.py
+++ b/tron/config/manager.py
@@ -193,7 +193,7 @@ class ConfigManager:
             # TODO: consider storing the hash alongside the config so that we only calculate
             # hashes once?
             return hash_digest(
-                yaml.dumps(
+                yaml.dump(
                     self.get_config_name_mapping()[name],
                     # ensure that the keys are always in a stable order
                     sort_keys=True,
--- a/tron/config/manager.py
+++ b/tron/config/manager.py
@@ -2,6 +2,7 @@ import hashlib
 import logging
 import os
 from copy import deepcopy
+from typing import Union
 
 from tron import yaml
 from tron.config import config_parse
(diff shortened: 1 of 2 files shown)


jfongatyelp (dismissed):
I see there were some things calculating a hash for these configs (reading the files listed from the manifestfile), and those seem to still be reading directly from disk and not the cache -- since we're writing back to the file on every update, that should all be fine but do we need to worry about getting out of sync there?

nemacysts:
> I see there were some things calculating a hash for these configs (reading the files listed from the manifestfile), and those seem to still be reading directly from disk and not the cache -- since we're writing back to the file on every update, that should all be fine but do we need to worry about getting out of sync there?

@jfongatyelp good question - i think i looked at this when making these changes, but PTO has wiped all memory of what i thought. let me dig into this again!

jfongatyelp (dismissed):

tron/config/manager.py
```diff
@@ -165,6 +184,25 @@ def get_hash(self, name) -> str:
         """Return a hash of the configuration contents for name."""
         if name not in self:
             return self.DEFAULT_HASH
+
+        if name in self.get_config_name_mapping():
+            # unfortunately, we have the parsed dict in memory.
+            # rather than hit the disk to get the raw string - let's convert
+            # the in-memory dict to a yaml string and hash that to save a couple
+            # ms (in testing, ~3ms over loading from disk and ~1ms over dumping to json :p)
+            # TODO: consider storing the hash alongside the config so that we only calculate
```
jfongatyelp, line 193: good idea! ++

nemacysts:
@jfongatyelp sorry - i didn't realize i had a typo leftover after one last attempt at timing json vs yaml vs disk hashing!

jfongatyelp (approved):

PR #1014, opened 2024-12-21
Disable the tooltip animation that is causing a strange problem where you can't get a tooltip when entering from the top of an object

KaspariK:
## Context
Working my way through some UI bugs that I've stumbled upon. This was pretty tricky to solve because we use an ancient version of bootstrap. 

>  We choose to fix Tron. We choose to fix Tron... We choose to fix Tron in this decade and do the other things, not because they are easy, but because they are hard! 
> \- John F. Kennedy

## Solution
- [Before](https://fluffy.yelpcorp.com/i/RH33zkzJmSZrrqXn8rgQDmGg6nm8DzT5.gif)
- [After](https://fluffy.yelpcorp.com/i/QB5pnhZzHCqNdK4598HLznJ9RNJ1pFQT.gif)

### DIFF
--- a/tronweb/coffee/views.coffee
+++ b/tronweb/coffee/views.coffee
@@ -63,7 +63,9 @@ window.isSelected = (current, value) ->
     if current == value then "selected" else ""
 
 window.makeTooltips = (root) ->
-    root.find('.tt-enable').tooltip()
+    root.find('.tt-enable').tooltip({
+        animation: false
+    })
 
 
 window.formatName = (name) =>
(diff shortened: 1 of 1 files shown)


tronweb/coffee/views.coffee
```diff
@@ -63,7 +63,9 @@ window.isSelected = (current, value) ->
     if current == value then "selected" else ""
 
 window.makeTooltips = (root) ->
-    root.find('.tt-enable').tooltip()
+    root.find('.tt-enable').tooltip({
+        animation: false
```
nemacysts, line 67: i'd probably add a comment re: why we're disabling animation here:
```suggestion
        # we're seeing a weird issue where attempting to get the tooltip from an 
        # object doesn't work if you enter the object from the top (???) - for some
        # reason, disabling animations fixes the issue
        animation: false
```

nemacysts (approved):

nemacysts (approved):

PR #1015, opened 2024-12-27
Add read_json to be a part of master.yaml config

EmanElsaban:
This adds the read_json to be a config in master.yaml instead of srv-configs. It makes it easier to set this option per Tron Master.

### Testing
Tested this by setting read_json in master.yaml and checked if the json deserialization was enabled or not

### DIFF
--- a/tests/serialize/runstate/statemanager_test.py
+++ b/tests/serialize/runstate/statemanager_test.py
@@ -70,6 +70,7 @@ class TestPersistentStateManager(TestCase):
         self.store.build_key.side_effect = lambda t, i: f"{t}{i}"
         self.buffer = StateSaveBuffer(1)
         self.manager = PersistentStateManager(self.store, self.buffer)
+        self.read_json = False
 
     def test__init__(self):
         assert_equal(self.manager._impl, self.store)
@@ -83,7 +84,6 @@ class TestPersistentStateManager(TestCase):
 
     def test_restore(self):
         job_names = ["one", "two"]
-        read_json = False
         with mock.patch.object(self.manager, "_restore_dicts", autospec=True,) as mock_restore_dicts, mock.patch.object(
             self.manager,
             "_restore_runs_for_job",
@@ -99,7 +99,7 @@ class TestPersistentStateManager(TestCase):
 
             restored_state = self.manager.restore(job_names)
             assert mock_restore_dicts.call_args_list == [
-                mock.call(runstate.JOB_STATE, job_names, read_json),
+                mock.call(runstate.JOB_STATE, job_names, self.read_json),
             ]
             assert len(mock_restore_runs.call_args_list) == 2
             assert restored_state == {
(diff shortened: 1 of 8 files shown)


tests/serialize/runstate/statemanager_test.py
```diff
@@ -83,6 +83,7 @@ def test_keys_for_items(self):
 
     def test_restore(self):
         job_names = ["one", "two"]
+        read_json = False
```
nemacysts, line 86: we can probably move this to the setup function (https://github.com/Yelp/Tron/blob/df1dc3520e4fd63f880ac1c23e66b8529d24c627/tests/serialize/runstate/statemanager_test.py#L68) so that we don't need to set it for every test :)

tron/serialize/runstate/statemanager.py
```diff
@@ -133,13 +133,13 @@ def restore(self, job_names):
...
+    def _restore_runs_for_job(self, job_name, job_state, read_json: bool = False):
         """Restore the state for the runs of each job"""
         run_nums = job_state["run_nums"]
         keys = [jobrun.get_job_run_id(job_name, run_num) for run_num in run_nums]
-        job_runs_restored_states = self._restore_dicts(runstate.JOB_RUN_STATE, keys)
-        runs = copy.copy(job_runs_restored_states)
-        for run_id, state in runs.items():
+        job_runs_restored_states = self._restore_dicts(runstate.JOB_RUN_STATE, keys, read_json)
+        restored_states_copy = copy.copy(job_runs_restored_states)
+        for run_id, state in restored_states_copy.items():
```
nemacysts, line 142: ignorable suggestion: 
```suggestion
        all_job_runs = copy.copy(job_runs_restored_states)
        for run_id, state in all_job_runs.items():
```
the suggested naming would make it clearer what this holds :)

nemacysts (approved):
suggestions are non-blocking :)

tron/serialize/runstate/statemanager.py
```diff
@@ -154,10 +154,10 @@ def _keys_for_items(self, item_type, names):
         keys = (self._impl.build_key(item_type, name) for name in names)
         return dict(zip(keys, names))
 
-    def _restore_dicts(self, item_type: str, items: List[str]) -> Dict[str, dict]:
+    def _restore_dicts(self, item_type: str, items: List[str], read_json: bool = False) -> Dict[str, dict]:
         """Return a dict mapping of the items name to its state data."""
         key_to_item_map = self._keys_for_items(item_type, items)
-        key_to_state_map = self._impl.restore(key_to_item_map.keys())
+        key_to_state_map = self._impl.restore(key_to_item_map.keys(), read_json)
```
jfongatyelp, line 160: I saw you added the `read_json` param to the dynamodb_state_store, but it would be cool if we added the json support to the `shelve` store's `restore` too (if we're still planning to support non-dynamo state for development anyway)

jfongatyelp (approved):

tron/serialize/runstate/statemanager.py
EmanElsaban: yea I agree. I think we can add this change in a separate pr since it will add JSON support when writing as well the shelve objects since I think currently they are pickled objects

jfongatyelp (approved):

PR #1016, opened 2025-01-06
U/kkasp/tron 2239 migrate pickle only state to json

KaspariK:
Script for working with pickle/json keys. Supports outputting, converting, and deleting.

TODO: probably worth incorporating some of luis' state restore script?

### DIFF
--- a/tools/pickles_to_json.py
+++ b/tools/pickles_to_json.py
@@ -347,20 +347,20 @@ Actions:
   convert           Convert pickled state data to JSON format and update the DynamoDB table.
   dump-pickle       Load and print the pickles for specified keys.
   dump-json         Load and print JSON data for specified keys.
-  delete-keys       Delete the specified keys from the DynamoDB table.
+  delete       Delete the specified keys from the DynamoDB table.
 Examples:
-  Validate pickles (dry run, write deprecated keys to deprecated_keys.txt):
-    pickles_to_json.py --table-name infrastage-tron-state --table-region us-west-1 convert --all --dry-run --deprecated-keys-output deprecated_keys.txt --tron-api tron-infrastage
-  Convert all pickles to JSON (without dry run):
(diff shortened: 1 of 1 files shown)


tools/pickles_to_json.py
```diff
@@ -0,0 +1,399 @@
...
+    """
+    Load and print JSON data for the given list of keys.
+    :param source_table: The DynamoDB table resource.
+    :param keys: A list of keys for which to load and print JSON data.
+    """
+    for key in keys:
+        dump_json_key(source_table, key)
+
+
+def delete_keys(source_table: ServiceResource, keys: List[str]) -> None:
```
KaspariK, line 122: Based on discussion with @nemacysts and @EmanElsaban we can probably do something here where we look for jobs that Tron doesn't know about and use that as the test for a valid delete

tools/pickles_to_json.py
```diff
@@ -0,0 +1,534 @@
...
+    delete_keys_parser.add_argument(
+        "--all",
+        action="store_true",
+        help="Apply the action to all keys in the table.",
+    )
+    delete_keys_parser.add_argument(
+        "--dry-run",
+        action="store_true",
+        help="Simulate the deletion without making any changes to the DynamoDB table",
+    )
```
nemacysts, line 474: i'm personally a fan of dry-run by default, but that's just me :p

nemacysts (dismissed):

tools/pickles_to_json.py
KaspariK: Good call.

I took the easier (lazy) path here with an --execute flag and just passed `not args.execute` to the existing dry_run logic. I am down to rewrite the conversion functions a bit to avoid the negative logic, but only if you comment that 😛

EmanElsaban (approved):

nemacysts (approved):

PR #1017, opened 2025-01-08
Add default behaviour for jobs that dont have some keys and handle the non-existent of some json_vals

EmanElsaban:
In this PR, we add the .get to some keys that we have seen error out when running the migration script in stage and that's because these jobs were too old and they didnt have some of these keys specified. If a key errors out, then there is no json val written for that action run. Thus, this PR sets a default value for those keys. It also adds a try/except to catch the scenario where a json_val doesnt exist for a key, otherwise I think it would fail silently and reset jobs to 0.

Note: I didn't change all keys to use .get instead, I wanted to change whatever was necessary as I think some keys should still error out on missing key if they are not optional? thoughts on this.

### DIFF
--- a/tron/core/action.py
+++ b/tron/core/action.py
@@ -115,6 +115,13 @@ class ActionCommandConfig(Persistable):
             return obj
 
         try:
+            # NOTE: you'll notice that there's a lot of get() accesses of state_data for
+            # pretty common fields - this is because ActionCommandConfig is used by more
+            # than one type of ActionRun (Kubernetes, Mesos, SSH) and these generally look
+            # different. Alternatively, some of these fields are used by KubernetesActionRun
+            # but are relatively new and older runs do not have data for them.
+            # Once we get rid of the SSH and Mesos code as well as older runs in DynamoDB,
+            # we'll likely be able to clean this up.
             return json.dumps(
                 {
                     "command": state_data["command"],
--- a/tron/core/action.py
+++ b/tron/core/action.py
@@ -117,37 +117,42 @@ class ActionCommandConfig(Persistable):
         try:
             return json.dumps(
                 {
-                    "command": state_data["command"],
-                    "cpus": state_data["cpus"],
-                    "mem": state_data["mem"],
-                    "disk": state_data["disk"],
-                    "cap_add": state_data["cap_add"],
-                    "cap_drop": state_data["cap_drop"],
+                    "command": state_data.get("command"),
+                    "cpus": state_data.get("cpus"),
+                    "mem": state_data.get("mem"),
+                    "disk": state_data.get("disk"),
+                    "cap_add": state_data.get("cap_add", []),
+                    "cap_drop": state_data.get("cap_drop", []),
                     "constraints": [
-                        serialize_namedtuple(constraint) for constraint in state_data["constraints"]
+                        serialize_namedtuple(constraint) for constraint in state_data.get("constraints", [])
                     ],  # convert each ConfigConstraint to dictionary, so it would be a list of dicts
-                    "docker_image": state_data["docker_image"],
+                    "docker_image": state_data.get("docker_image"),
                     "docker_parameters": [
-                        serialize_namedtuple(parameter) for parameter in state_data["docker_parameters"]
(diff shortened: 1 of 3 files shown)


tron/core/action.py
```diff
@@ -117,37 +117,42 @@ def serialize_namedtuple(obj):
...
-                    "mem": state_data["mem"],
-                    "disk": state_data["disk"],
-                    "cap_add": state_data["cap_add"],
-                    "cap_drop": state_data["cap_drop"],
+                    "command": state_data.get("command"),
+                    "cpus": state_data.get("cpus"),
+                    "mem": state_data.get("mem"),
+                    "disk": state_data.get("disk"),
+                    "cap_add": state_data.get("cap_add", []),
+                    "cap_drop": state_data.get("cap_drop", []),
```
nemacysts, line 125: we have a default if cap_drop isn't provided in task_proc - but i think to take advantage of that we'd have to not pass through a value if we didn't get one, which would probably be a bit tricky

i sorta wonder if we should instead default to re-adding the default cap drops here

although: i'm kinda surpised that there's stuff missing these to begin with

tron/core/action.py
```diff
@@ -117,37 +117,42 @@ def serialize_namedtuple(obj):
         try:
             return json.dumps(
                 {
-                    "command": state_data["command"],
-                    "cpus": state_data["cpus"],
-                    "mem": state_data["mem"],
-                    "disk": state_data["disk"],
-                    "cap_add": state_data["cap_add"],
-                    "cap_drop": state_data["cap_drop"],
+                    "command": state_data.get("command"),
```
nemacysts, line 120: everything should have a command - sounds like maybe we have some invalid data stored in dynamo?

tron/core/action.py
EmanElsaban: IIRC I have seen a job that was either missing cap_add or cap_drop. There is a lot of ancient data in the db that had features no longer supported . I see in actioncommandconfig theyre set to the config if it exists or empty list
```
            cap_add=config.cap_add or [],
            cap_drop=config.cap_drop or [],
```

KaspariK (dismissed):

tron/core/action.py
```diff
@@ -124,30 +124,35 @@ def serialize_namedtuple(obj):
                     "cap_add": state_data["cap_add"],
                     "cap_drop": state_data["cap_drop"],
                     "constraints": [
-                        serialize_namedtuple(constraint) for constraint in state_data["constraints"]
+                        serialize_namedtuple(constraint) for constraint in state_data.get("constraints", [])
```
nemacysts, line 127: i'd maybe add a comment above the `return json.dumps(` line that says something to the effect of
```Python
# NOTE: you'll notice that there's a lot of get() accesses of state_data for
# pretty common fields - this is because ActionCommandConfig is used by more
# than one type of ActionRun (Kubernetes, Mesos, SSH) and these generally look
# different. Alternatively, some of these fields are used by KubernetesActionRun
# but are relatively new and older runs do not have data for them.
# Once we get rid of the SSH and Mesos code, we'll likely be able to clean this
# up.
```

nemacysts (approved):
imo, it'd be nice to add some explanatory comments here since this seems likely to surprise future yelpers that aren't us (or even us in the future after we forget everything :p) without some breadcrumbs :)

tron/core/actionrun.py
```diff
@@ -811,12 +811,12 @@ def to_json(state_data: dict) -> Optional[str]:
                     "job_run_id": state_data["job_run_id"],
                     "action_name": state_data["action_name"],
                     "state": state_data["state"],
-                    "original_command": state_data["original_command"],
+                    "original_command": state_data.get("original_command"),
```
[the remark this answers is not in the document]
EmanElsaban: I've seen both fail for a job from what i recall before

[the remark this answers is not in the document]
nemacysts: if we're not too pressed for time, it'd be nice to look at those jobs and see if they're junk data - but otherwise a `# TODO: figure out why we've seen cases of this - this shouldn't happen` seems fine to me as well :)

nemacysts (approved):

KaspariK (approved):

PR #1018, opened 2025-01-08
Update delete_item logic to handle json partitions

KaspariK:
We need to update the logic for deleting items to account for a difference in partitions between pickles and json

### DIFF
--- a/tests/serialize/runstate/dynamodb_state_store_test.py
+++ b/tests/serialize/runstate/dynamodb_state_store_test.py
@@ -273,7 +273,7 @@ class TestDynamoDBStateStore:
 
     def test_delete_item_with_json_partitions(self, store, small_object, large_object):
         key = store.build_key("job_state", "test_job")
-        value = small_object
+        value = large_object
 
         store.save([(key, value)])
         store._consume_save_queue()
--- a/tests/serialize/runstate/dynamodb_state_store_test.py
+++ b/tests/serialize/runstate/dynamodb_state_store_test.py
@@ -262,8 +262,8 @@ class TestDynamoDBStateStore:
         value = large_object
         pairs = list(zip(keys, (value for i in range(len(keys)))))
         store.save(pairs)
-
         store._consume_save_queue()
+
         for key, value in pairs:
             store._delete_item(key)
         for key, value in pairs:
--- a/tests/serialize/runstate/dynamodb_state_store_test.py
+++ b/tests/serialize/runstate/dynamodb_state_store_test.py
@@ -263,11 +263,36 @@ class TestDynamoDBStateStore:
         pairs = list(zip(keys, (value for i in range(len(keys)))))
         store.save(pairs)
 
+        store._consume_save_queue()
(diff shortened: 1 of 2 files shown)


tests/serialize/runstate/dynamodb_state_store_test.py
```diff
@@ -263,11 +263,36 @@ def test_delete_item(self, store, small_object, large_object):
         pairs = list(zip(keys, (value for i in range(len(keys)))))
         store.save(pairs)
 
+        store._consume_save_queue()
         for key, value in pairs:
```
nemacysts, line 267: ignorable nit:
```suggestion
        store.save(pairs)
        store._consume_save_queue()
        
        for key, value in pairs:
```
would be more consistent with the previous test (and make it clearer that the save()/_consume_save_queue() are coupled

tests/serialize/runstate/dynamodb_state_store_test.py
```diff
@@ -263,11 +263,36 @@ def test_delete_item(self, store, small_object, large_object):
...
+    def test_delete_item_with_json_partitions(self, store, small_object, large_object):
+        key = store.build_key("job_state", "test_job")
+        value = small_object
+
+        store.save([(key, value)])
+        store._consume_save_queue()
+
+        num_partitions, num_json_val_partitions = store._get_num_of_partitions(key)
+        assert num_partitions > 0
+        assert num_json_val_partitions > 0
```
nemacysts, line 283: just curious: how come we're not asserting an exact # of partitions?

tests/serialize/runstate/dynamodb_state_store_test.py
```diff
@@ -263,11 +263,36 @@ def test_delete_item(self, store, small_object, large_object):
...
         for key, value in pairs:
             store._delete_item(key)
-
         for key, value in pairs:
-            assert_equal(store._get_num_of_partitions(key), 0)
+            num_partitions, num_json_val_partitions = store._get_num_of_partitions(key)
+            assert_equal(num_partitions, 0)
+            assert_equal(num_json_val_partitions, 0)
+
+    def test_delete_item_with_json_partitions(self, store, small_object, large_object):
```
KaspariK, line 274: Hmm, maybe it would make more sense to test with the larger object here

nemacysts (approved):
just left some non-blocking suggestions/questions :)

tests/serialize/runstate/dynamodb_state_store_test.py
KaspariK: I think it potentially more finicky/brittle and ultimately all I care about here is that we have partitions. Partition amounts might change as we update the max amount back to 400kb post-pickle + I've fiddled with the size of large_object a few times while working on other parts of Tron.

nemacysts: gotcha - makes sense :)

KaspariK:
I'll add that last comment as a todo lol

PR #1020, opened 2025-01-09
Bump task_processing to v1.3.1 for capability fix

nemacysts:
Just like Yelp/paasta#3972 and Yelp/paasta#3973, we need to ensure that there are no duplicates between cap_add and cap_drop - otherwise, the cap_drop entry will "win" and the duplicate capability will not be added.


EmanElsaban (approved):

KaspariK (approved):
I support this

PR #1023, opened 2025-01-15
Add dynamodb retry config for throttling and other errors. Add exponential backoff and jitter for unprocessed keys. Fix edge case where we succesfully process keys on our last attempt but still fail

KaspariK:
1. Use Boto3 retries (see [Retries - Boto3](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/retries.html))
2. Backoff on getting unprocessed keys

### DIFF
--- a/tests/serialize/runstate/dynamodb_state_store_test.py
+++ b/tests/serialize/runstate/dynamodb_state_store_test.py
@@ -345,7 +345,7 @@ class TestDynamoDBStateStore:
                 ):
                     store.restore(keys)
             except Exception:
-                assert_equal(mock_failed_read.call_count, 11)
+                assert_equal(mock_failed_read.call_count, 10)
 
     def test_restore_exception_propagation(self, store, small_object):
         # This test is to ensure that restore propagates exceptions upwards: see DAR-2328
--- a/tests/serialize/runstate/dynamodb_state_store_test.py
+++ b/tests/serialize/runstate/dynamodb_state_store_test.py
@@ -334,13 +334,13 @@ class TestDynamoDBStateStore:
     @pytest.mark.parametrize(
         "attempt, expected_delay",
         [
-            (1, 0.5),
-            (2, 1.0),
-            (3, 2.0),
-            (4, 4.0),
-            (5, 8.0),
-            (6, 10.0),
-            (7, 10.0),
+            (1, 1),
+            (2, 2),
+            (3, 4),
+            (4, 8),
+            (5, 10),
+            (6, 10),
+            (7, 10),
         ],
     )
     def test_calculate_backoff_delay(self, store, attempt, expected_delay):
--- a/tests/serialize/runstate/dynamodb_state_store_test.py
+++ b/tests/serialize/runstate/dynamodb_state_store_test.py
@@ -8,7 +8,6 @@ from moto import mock_dynamodb2
 from moto.dynamodb2.responses import dynamo_json_dump
 
 from testifycompat import assert_equal
-from testifycompat.assertions import assert_in
 from tron.serialize.runstate.dynamodb_state_store import DynamoDBStateStore
 from tron.serialize.runstate.dynamodb_state_store import MAX_UNPROCESSED_KEYS_RETRIES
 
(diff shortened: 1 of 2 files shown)


tron/serialize/runstate/dynamodb_state_store.py
```diff
@@ -106,20 +128,33 @@ def _get_items(self, table_keys: list) -> object:
...
+                attempts += 1
+                # Exponential backoff for retrying unprocessed keys
+                exponential_delay = min(base_delay_seconds * (2 ** (attempts - 1)), max_delay_seconds)
+                # Full jitter (i.e. from 0 to exponential_delay) will help minimize the number and length of calls
+                jitter = random.uniform(0, exponential_delay)
+                delay = jitter
+                log.warning(
+                    f"Attempt {attempts}/{MAX_UNPROCESSED_KEYS_RETRIES} - Retrying {len(cand_keys_list)} unprocessed keys after {delay:.2f}s delay."
+                )
+                time.sleep(delay)
```
KaspariK, line 151: What to do about this lil guy?

KaspariK: !8ball we should use a restore thread

tests/serialize/runstate/dynamodb_state_store_test.py
```diff
@@ -307,45 +310,56 @@ def test_retry_saving(self, store, small_object, large_object):
...
-        pairs = zip(keys, (value for i in range(len(keys))))
+        pairs = zip(keys, [value] * len(keys))
         store.save(pairs)
+        store._consume_save_queue()
+
+        # Mock random.uniform to return the upper limit of the range so that we are simulating max jitter
+        def side_effect_random_uniform(a, b):
+            return b
+
+        mock_random_uniform.side_effect = side_effect_random_uniform
```
nemacysts, line 340: i think
```suggestion
        mock_random_uniform.side_effect = lambda a, b: b
```
would work too - but either way is fine (having a nested function definition in a test usually looks a little off to me since it's sometimes a sign that said function could be a mock and a lambda makes it a lot clearer that this is a throwaway)

tests/serialize/runstate/dynamodb_state_store_test.py
```diff
@@ -8,7 +8,9 @@
 from moto.dynamodb2.responses import dynamo_json_dump
 
 from testifycompat import assert_equal
+from testifycompat.assertions import assert_in
```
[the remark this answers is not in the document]
nemacysts: (and related, I'd probably replace the new `assert_equal` calls with `assert X == Y`)

tests/serialize/runstate/dynamodb_state_store_test.py
```diff
@@ -307,45 +310,56 @@ def test_retry_saving(self, store, small_object, large_object):
...
-            try:
-                with mock.patch("tron.config.static_config.load_yaml_file", autospec=True), mock.patch(
-                    "tron.config.static_config.build_configuration_watcher", autospec=True
-                ):
-                    store.restore(keys)
-            except Exception:
-                assert_equal(mock_failed_read.call_count, 11)
+            with pytest.raises(Exception) as exec_info, mock.patch(
+                "tron.config.static_config.load_yaml_file", autospec=True
+            ), mock.patch("tron.config.static_config.build_configuration_watcher", autospec=True):
```
nemacysts, line 349: could these be merged with the outer context manager?

tron/serialize/runstate/dynamodb_state_store.py
nemacysts: yea, we should probably try to figure out a non-blocking way to do this or have this run in a separate thread - if we get to the worst case of 5 attempts and this is running on the reactor thread, we'll essentially block all of tron from doing anything for 20s

nemacysts: although, actually - this is probably fine since we do all sorts of blocking stuff in restore and aren't expecting tron to be usable/do anything until we've restored everything

...so maybe this is fine?

tron/serialize/runstate/dynamodb_state_store.py
```diff
@@ -106,20 +128,33 @@ def _get_items(self, table_keys: list) -> object:
...
                     log.exception("Encountered issues retrieving data from DynamoDB")
                     raise e
-            attempts_to_retrieve_keys += 1
+            if cand_keys_list:
+                attempts += 1
+                # Exponential backoff for retrying unprocessed keys
+                exponential_delay = min(base_delay_seconds * (2 ** (attempts - 1)), max_delay_seconds)
+                # Full jitter (i.e. from 0 to exponential_delay) will help minimize the number and length of calls
+                jitter = random.uniform(0, exponential_delay)
+                delay = jitter
```
nemacysts, line 147: did we mean to add the exponential delay + the jitter? or is waiting for a random time between 0 and the expected delay what we wanted? (i.e., is waiting 0 seconds fine?)

tests/serialize/runstate/dynamodb_state_store_test.py
```diff
@@ -307,45 +310,56 @@ def test_retry_saving(self, store, small_object, large_object):
...
+
+        # We also need to verify that sleep was called with expected delays
+        expected_delays = []
+        base_delay_seconds = 0.5
+        max_delay_seconds = 10
+        for attempt in range(1, MAX_UNPROCESSED_KEYS_RETRIES + 1):
+            expected_delay = min(base_delay_seconds * (2 ** (attempt - 1)), max_delay_seconds)
+            expected_delays.append(expected_delay)
+        actual_delays = [call.args[0] for call in mock_sleep.call_args_list]
+        assert_equal(actual_delays, expected_delays)
```
nemacysts, line 362: i'd maybe extract the exponential backoff logic in `tron/serialize/runstate/dynamodb_state_store.py` to a function so that we can write a more targeted test for that and simplify this to checking if we called that function the right amount of times

(mostly 'cause I generally try to avoid for loops/calculations inside tests :p)

tron/serialize/runstate/dynamodb_state_store.py
nemacysts: oh, i see - boto is doing it's own exponential backoff?

imo, we can likely skip adding any jitter - each dynamodb table only has a single reader/writer (tron) so there's not much of a risk of a thundering herd scenario :)

tron/serialize/runstate/dynamodb_state_store.py
```diff
@@ -106,20 +128,33 @@ def _get_items(self, table_keys: list) -> object:
...
-                    # add any potential unprocessed keys to the thread pool
-                    if resp.result()["UnprocessedKeys"].get(self.name) and attempts_to_retrieve_keys < MAX_ATTEMPTS:
-                        cand_keys_list.extend(resp.result()["UnprocessedKeys"][self.name]["Keys"])
-                    elif attempts_to_retrieve_keys >= MAX_ATTEMPTS:
-                        failed_keys = resp.result()["UnprocessedKeys"][self.name]["Keys"]
-                        error = Exception(
-                            f"tron_dynamodb_restore_failure: failed to retrieve items with keys \n{failed_keys}\n from dynamodb\n{resp.result()}"
-                        )
-                        raise error
+                    result = resp.result()
```
EmanElsaban, line 131: I wonder if we should also print the response when we get into the exception block to also have an idea on why we got unprocessed keys and why we exceeded the attempts

EmanElsaban: so maybe we add it here
```
                except Exception as e:
                    log.exception("Encountered issues retrieving data from DynamoDB")
                    raise e
```

tests/serialize/runstate/dynamodb_state_store_test.py
```diff
@@ -294,7 +296,8 @@ def test_delete_item_with_json_partitions(self, store, small_object, large_objec
             vals = store.restore([key])
         assert key not in vals
 
-    def test_retry_saving(self, store, small_object, large_object):
+    @mock.patch("time.sleep", return_value=None)
```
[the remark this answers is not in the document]
KaspariK: Alright, not what you commented on at all, but upon closer inspection this test isn't really testing much. I'll rewrite this

tron/serialize/runstate/dynamodb_state_store.py
KaspariK: I was hesitant to dump the response because it can get pretty large. After a lot of reading I've landed on logging ResponseMetadata on ClientError. This should capture what we care about

See https://fluffy.yelpcorp.com/i/qWG1tRPrFt40M6pPr3lLkXnCSbJJBFhd.html

tron/serialize/runstate/dynamodb_state_store.py
```diff
@@ -106,20 +132,33 @@ def _get_items(self, table_keys: list) -> object:
...
+                    f"Attempt {attempts}/{MAX_UNPROCESSED_KEYS_RETRIES} - "
+                    f"Retrying {len(cand_keys_list)} unprocessed keys after {delay}s delay."
+                )
+                time.sleep(delay)
+        if cand_keys_list:
+            error = Exception(
+                f"tron_dynamodb_restore_failure: failed to retrieve items with keys \n{cand_keys_list}\n from dynamodb after {MAX_UNPROCESSED_KEYS_RETRIES} retries."
+            )
+            log.error(error)
+            raise error
```
nemacysts, line 161: i'd maybe do something like
```suggestion
            msg = f"tron_dynamodb_restore_failure: failed to retrieve items with keys \n{cand_keys_list}\n from dynamodb after {MAX_UNPROCESSED_KEYS_RETRIES} retries."
            log.error(msg)
            
            raise KeyError(msg)
```
as it looks a little funky to log an Exception object, and a KeyError would give more info :)

(or we could have a custom exception defined for this)

tron/serialize/runstate/dynamodb_state_store.py
```diff
@@ -337,25 +376,21 @@ def __setitem__(self, key: str, value: Tuple[bytes, str]) -> None:
...
-                        )
-                        log.error(f"Failed to save partition for key: {key}, error: {repr(e)}")
-                        raise e
-                    else:
-                        log.warning(f"Got error while saving {key}, trying again: {repr(e)}")
+                    timer(
+                        name="tron.dynamodb.setitem",
+                        delta=time.time() - start,
+                    )
+                    log.error(f"Failed to save partition for key: {key}, error: {repr(e)}")
```
nemacysts, line 392: i know this is old code being moved around, so you can leave this as-is but
```suggestion
                    log.exception(f"Failed to save partition for key: {key}")
```
would include the full traceback for us automatically :)

nemacysts: although: it looks like there's a behavior change here?

might be worth adding a comment here (or in the docstring) that this function will not retry on its own and that it's the callers responsibility to do so)

tron/serialize/runstate/dynamodb_state_store.py
```diff
@@ -106,20 +132,33 @@ def _get_items(self, table_keys: list) -> object:
...
+                    log.exception(f"ClientError during batch_get_item: {e.response}")
+                    raise
+                except Exception:
                     log.exception("Encountered issues retrieving data from DynamoDB")
-                    raise e
-            attempts_to_retrieve_keys += 1
+                    raise
+            if cand_keys_list:
+                attempts += 1
+                delay = self._calculate_backoff_delay(attempts)
```
nemacysts, line 150: fwiw, I think it's probably fine to rely on the built-in backoff from boto - there shouldn't be anything else touching these dynamo tables other than tron, so we don't really need any jitter to avoid a thundering herd scenario :)

tron/serialize/runstate/dynamodb_state_store.py
```diff
@@ -83,12 +102,19 @@ def chunk_keys(self, keys: Sequence[T]) -> List[Sequence[T]]:
             cand_keys_chunks.append(keys[i : min(len(keys), i + 100)])
         return cand_keys_chunks
 
+    def _calculate_backoff_delay(self, attempt: int) -> int:
+        base_delay_seconds = 1
+        max_delay_seconds = 10
+        delay: int = min(base_delay_seconds * (2 ** (attempt - 1)), max_delay_seconds)
```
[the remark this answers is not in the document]
nemacysts: i guess the attempt=0 case would really just lead to a .5s jitter...which is fine?

nemacysts (dismissed):
generally looks good to me - i left some minor suggestions, but i don't think any of them are blocking atm

tron/serialize/runstate/dynamodb_state_store.py
KaspariK: There are two levels of backoff, basically. There is the built-in retry config that catches something like throttling, and then there is our own backoff based on unprocessedkeys. This seems to be the suggested approach based on the warning in: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/dynamodb/client/batch_get_item.html

It's not 100% clear to me that the retry config will handle unprocessedkeys

nemacysts: (sorry, missed this reply!)
ah, gotcha - might be helpful to add a comment about that here as i can definitely see myself coming back to this in a year and wondering why we have our own backoff when the boto comment further above says we're using the built-in retries :p

(not a blocker tho)

tron/serialize/runstate/dynamodb_state_store.py
```diff
@@ -83,12 +103,21 @@ def chunk_keys(self, keys: Sequence[T]) -> List[Sequence[T]]:
...
+
     def _get_items(self, table_keys: list) -> object:
         items = []
         # let's avoid potentially mutating our input :)
         cand_keys_list = copy.copy(table_keys)
-        attempts_to_retrieve_keys = 0
-        while len(cand_keys_list) != 0:
+        attempts = 0
+
+        while len(cand_keys_list) != 0 and attempts < MAX_UNPROCESSED_KEYS_RETRIES:
```
nemacysts, line 120: just posting this here for future us: we'll probably want to refactor this at some point to not consume an attempt if we're making progress (i.e., we got at least one key back) and we're simply seeing dynamodb send us partial responses (unless we wanna take a hard line with what our data sizes are such that we can always get a full chunk back at any time) and only consume an attempt if we're doing an error-caused retry

nemacysts (approved):

nemacysts (approved):

PR #1029, opened 2025-02-04
Add pagination and search functionality to job collection API

cuza:
Implement pagination and search capabilities for the job collection API, allowing users to retrieve jobs in a paginated format and filter jobs based on search queries.

### DIFF
--- a/tests/api/resource_test.py
+++ b/tests/api/resource_test.py
@@ -223,8 +223,24 @@ class TestJobCollectionResource(WWWTestCase):
     def test_render_GET(self):
         self.resource.get_data = MagicMock()
         result = self.resource.render_GET(REQUEST)
-        assert_call(self.resource.get_data, 0, False, False, True, True)
+        assert_call(self.resource.get_data, 0, False, False, True, True, None, None)
         assert "jobs" in result
+
+    def test_render_SEARCH(self):
+        request = build_request(query="test")
+        job1 = MagicMock()
+        job1.get_name.return_value = "testjob1"
+        job2 = MagicMock()
+        job2.get_name.return_value = "othertestjob"
+        job3 = MagicMock()
+        job3.get_name.return_value = "otherjob"
+        self.resource.job_collection.get_jobs.return_value = [job1, job2, job3]
+        result = self.resource.render_SEARCH(request)
+        assert "jobs" in result
+        assert len(result["jobs"]) == 2
+        job_names = [job["name"] for job in result["jobs"]]
+        assert "testjob1" in job_names
+        assert "othertestjob" in job_names
 
     def test_getChild(self):
         child = self.resource.getChild(b"testname", mock.Mock())
--- a/tron/api/resource.py
+++ b/tron/api/resource.py
@@ -5,6 +5,7 @@ view current state, event history and send commands to trond.
 import collections
 import datetime
 import logging
+import math
 import traceback
 
 import staticconf
@@ -310,10 +311,16 @@ class JobCollectionResource(resource.Resource):
         include_action_runs=False,
         include_action_graph=True,
         include_node_pool=True,
+        page=None,
+        page_size=None,
     ):
+        jobs = self.job_collection.get_jobs()
+        if page is not None and page_size is not None:
+            start = (page - 1) * page_size
+            jobs = jobs[start : start + page_size]
         return adapter.adapt_many(
             adapter.JobAdapter,
-            self.job_collection.get_jobs(),
+            jobs,
             include_job_run,
             include_action_runs,
             include_action_graph,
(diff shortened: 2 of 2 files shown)


tron/api/resource.py
```diff
@@ -350,13 +357,45 @@ def render_GET(self, request):
...
+                    response={"error": "page out of range"},
+                    code=http.BAD_REQUEST,
+                )
+            response = dict(jobs=jobs, pagination=dict(page=page, page_size=page_size, total=total))
+        else:
+            response = dict(jobs=jobs)
+        return respond(request=request, response=response)
+
+    @AsyncResource.bounded
+    def render_SEARCH(self, request):
```
nemacysts, line 393: my understanding is that the `render_*` functions are named after the HTTP request method used for the request that's being handled and that that SEARCH is not standard HTTP method and that we probably want QUERY? that said, not sure if that's still a draft RFC or if it's actually part of the standard

but just to confirm: the intent is that we'll have the troncli/ui use SEARCH/QUERY verbs to hit this new function?

tron/api/resource.py
```diff
@@ -350,13 +357,45 @@ def render_GET(self, request):
...
+                    request=request,
+                    response={"error": "page and page_size must be positive integers"},
+                    code=http.BAD_REQUEST,
+                )
+            if page > math.ceil(total / page_size):
+                return respond(
+                    request=request,
+                    response={"error": "page out of range"},
+                    code=http.BAD_REQUEST,
+                )
```
nemacysts, line 386: might be nice to have a validate function so that we can do something like:

```Python
request_ok, request_error = self.job_collection.validate_paging_params(page, page_size)
if not request_ok:
    return respond(
        request=request
        response={"error": request_error},
        code=http.BAD_REQUEST,
    )
```

tron/api/resource.py
```diff
@@ -350,13 +357,45 @@ def render_GET(self, request):
...
+        page = requestargs.get_integer(request, "page")
+        page_size = requestargs.get_integer(request, "page_size")
+
+        jobs = self.get_data(
+            include_job_runs,
+            include_action_runs,
+            include_action_graph,
+            include_node_pool,
+            page,
+            page_size,
```
nemacysts, line 370: we're using page/page_size here before validating them - is that intentional?

tron/api/resource.py
```diff
@@ -310,10 +311,16 @@ def get_data(
         include_action_runs=False,
         include_action_graph=True,
         include_node_pool=True,
+        page=None,
+        page_size=None,
     ):
+        jobs = self.job_collection.get_jobs()
+        if page is not None and page_size is not None:
+            start = (page - 1) * page_size
+            jobs = jobs[start : start + page_size]
```
nemacysts, line 320:
```suggestion 
        if page is not None and page_size is not None:
            jobs = self.job_collection.get_jobs_by_page(
                page,
                page_size,
            )
        else:
            jobs = self.job_collection.get_jobs()
```
might make testing easier (since then we can test the paging logic without testing the rest of `get_data`)

tests/api/resource_test.py
```diff
@@ -223,9 +223,25 @@ def setup_resource(self):
     def test_render_GET(self):
         self.resource.get_data = MagicMock()
         result = self.resource.render_GET(REQUEST)
-        assert_call(self.resource.get_data, 0, False, False, True, True)
+        assert_call(self.resource.get_data, 0, False, False, True, True, None, None)
         assert "jobs" in result
 
+    def test_render_SEARCH(self):
+        request = build_request(query="test")
+        job1 = MagicMock()
```
nemacysts, line 231: nit: we should probably set the `spec` on these to whatever the real type of these is so that the tests better match reality (e.g., so that the mocks fail should we ever remove `get_name` from the actual type, etc)

tron/api/resource.py
```diff
@@ -350,13 +357,45 @@ def render_GET(self, request):
...
+                )
+            if page > math.ceil(total / page_size):
+                return respond(
+                    request=request,
+                    response={"error": "page out of range"},
+                    code=http.BAD_REQUEST,
+                )
+            response = dict(jobs=jobs, pagination=dict(page=page, page_size=page_size, total=total))
+        else:
+            response = dict(jobs=jobs)
```
nemacysts, line 389: imo, inlining the dicts rather than calling the dict constructor would be more modern/easier to read - that said, we have pretty mixed usage of dict/{} in this file already so not a blocker

tron/api/resource.py
```diff
@@ -350,13 +357,45 @@ def render_GET(self, request):
...
+        return respond(request=request, response=response)
+
+    @AsyncResource.bounded
+    def render_SEARCH(self, request):
+        query = requestargs.get_string(request, "query")
```
nemacysts, line 394: do we need to do any validation on the `query` data?

nemacysts (commented):
i'm assuming this is related to the hackathon efforts last year to come up with a new tron ui, but adding additional context into the description would be nice (in case i'm wrong or for folks that don't have that knowledge)

and sidenote: how are we envisioning the search function being used? i noticed that it's not currently being paged - is that an oversight or something for future us to handle?

tron/api/resource.py
cuza: Yes your understanding is correct and I did added `SEARCH` as a custom http verb and it is not part of the standard, I don't even think it's part of a draft. It did't felt wrong when I was doing it but also all of this code is from a hackathon project so I was more focus on getting something built fast.

I'm more than happy to take suggestions on what endpoint we can drop the search at or we can have a discussion about if it makes sense to have the custom http verb

tron/api/resource.py
cuza: Great catch!!! this is a good point!!! most of the page validation happens a bit after on the code, handling exceptions here makes more sense.

tron/api/resource.py
cuza: Search is currently very basic we could maybe limit the length of the string 🤷🏽 we also don't use it in a way that could lead to SQL injections

tron/api/resource.py
cuza: TIL: [HTTP SEARCH verb](https://www.ietf.org/archive/id/draft-ietf-httpbis-safe-method-w-body-00.html) is a thing

cuza:
> i'm assuming this is related to the hackathon efforts last year to come up with a new tron ui, but adding additional context into the description would be nice (in case i'm wrong or for folks that don't have that knowledge)
> 
> and sidenote: how are we envisioning the search function being used? i noticed that it's not currently being paged - is that an oversight or something for future us to handle?

All your assumptions are correct:
 - The pagination is to be able to have a more snappy/faster tron ui.
 - The search was added cuz if we paginate tronweb main dashboard it will be hard to find a tronjob
   - Search does not have pagination since I left it for future me to implement. I also was not very sure if we wanted/needed search so keep it basic as a PoC to see if I liked the idea.

tron/api/resource.py
```diff
@@ -350,13 +357,45 @@ def render_GET(self, request):
...
+
+        if page is not None and page_size is not None:
+            total = len(self.job_collection.get_jobs())
+            if page < 1 or page_size < 1:
+                return respond(
+                    request=request,
+                    response={"error": "page and page_size must be positive integers"},
+                    code=http.BAD_REQUEST,
+                )
+            if page > math.ceil(total / page_size):
```
KaspariK, line 381: nit, but the case where we have 0 jobs should still return a page (with no jobs).

PR #1034, opened 2025-03-18
fix service name extraction logic

piax93:
It was never going to be a nice implementation, but my first attempt is just so WRONG, I'm ashamed of myself, mpiano--

### DIFF
--- a/tron/api/auth.py
+++ b/tron/api/auth.py
@@ -51,13 +51,12 @@ class AuthorizationFilter:
         token = (request.getHeader("Authorization") or "").strip()
         token = token.split()[-1] if token else ""  # removes "Bearer" prefix
         url_path = request.path.decode()
-        service = url_path.split("/")[-1].split(".", 1)[0] if "/jobs/" in url_path else None
         auth_outcome = self._is_request_authorized_impl(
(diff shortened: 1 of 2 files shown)


KaspariK (dismissed):

tron/api/auth.py
```diff
@@ -104,3 +103,18 @@ def _is_request_authorized_impl(
...
+        Example:
+        /api/jobs/someservice.instance/110/run -> someservice
+
+        :param str path: request path
+        :return: service name, or None if not found
+        """
+        if not path.startswith("/api/jobs/"):
+            return None
+        path_parts = path.split("/")
+        return path_parts[3].split(".", 1)[0] if len(path_parts) > 3 and path_parts[3] else None
```
nemacysts, line 120: thoughts on something like
```
def extract_service_name(url: str) -> Optional[str]:
    if match := re.search(r'/api/jobs/([^/.]+)', url):
        return match.group(1)
    return None
```
to remove the multiple splits/list indexings?

nemacysts (approved):
seems fine to me, but i will say i'm usually not a fan of multiple splits/list indexings :p

nemacysts (approved):

PR #1041, opened 2025-04-08
fix build_url_request when method is None

piax93:
I got too used to mypy catching this sort of stuff for me...

### DIFF
--- a/tron/commands/client.py
+++ b/tron/commands/client.py
@@ -54,7 +54,7 @@ def build_url_request(uri, data, headers=None, method=None):
     headers = headers or default_headers
     enc_data = urllib.parse.urlencode(data).encode() if data else None
     # Currently implementing auth only for management actions (i.e. POST requests)
-    if os.getenv("TRONCTL_API_AUTH") and (data or method.upper() == "POST"):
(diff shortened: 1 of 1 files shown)


nemacysts (approved):
yea, i've been meaning to enable mypy on more of tron...

tron/commands/client.py
```diff
@@ -54,7 +54,7 @@ def build_url_request(uri, data, headers=None, method=None):
     headers = headers or default_headers
     enc_data = urllib.parse.urlencode(data).encode() if data else None
     # Currently implementing auth only for management actions (i.e. POST requests)
-    if os.getenv("TRONCTL_API_AUTH") and (data or method.upper() == "POST"):
+    if os.getenv("TRONCTL_API_AUTH") and (data or (method and method.upper() == "POST")):
```
jfongatyelp, line 57: Sorry i've been lagging behind all the PRs and changes with your auth work, but just curious why do we need to send a token on some non-`POST` requests (where `data` is not empty)?

piax93: Because this thing uses urllib, and if data != None that will switch to a POST automatically even if the method is not provided.

jfongatyelp: TIL thanks!

Qmando (approved):

PR #1042, opened 2025-04-10
Reduce transact size to avoid cap

KaspariK:
There is no world where we can write 100 items of our chunk size without hitting the cap.


EmanElsaban (dismissed):

nemacysts (approved):
(for after we merge this: think it's possible to make this configurable so that we don't need a tron release?)

jfongatyelp (dismissed):

PR #1043, opened 2025-04-23
Drop under the limit a bit more for maximum safety

KaspariK:
The current limit is probably ok, but is a bit close for comfort. The impact of this should be negligible.


nemacysts (dismissed):
i think it'd be nice to have this come from config so that if there's any incidents before we finish removing the pickle code, we can drop this further without needing to go through a release or shenanigans

(but we've dropped this number quite a bit already, so we'll probably be fine here for a while?)

PR #1044, opened 2025-04-30
use refreshable tokens for api auth

piax93:
The new version adds an option to use an authentication flow with refresh tokens, which will allow `tronctl` commands to be run unattended for over the set duration of an identity token (which is fixed to 1 hour). Refresh tokens are valid for 90 days, which should be more than enough for all use cases.


nemacysts:
@piax93 just curious: can this be merged/released at any time or does this require doing any puppet/etc shenanigans in preparation?

piax93:
@nemacysts It can be released already. The Puppet shenanigans were already done.

PR #1046, opened 2025-05-14
U/kkasp/tron 2391 aggregate metrics

KaspariK:
We frequently find ourselves wanting better metrics around these things.

I'd been fiddling with some visualizations in https://grafana.yelpcorp.com/d/delvfqhtpfgu8e/scratch?orgId=1&from=now-6h&to=now&timezone=browser, but I think we'll just take anything worthwhile and add it to existing dashboards.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -51,7 +51,7 @@ oauthlib==3.1.0
 parso==0.7.0
 pexpect==4.7.0
 pickleshare==0.7.5
-prometheus-client==0.20.0
+prometheus-client==0.21.1
 prompt-toolkit==3.0.38
 psutil==5.6.6
 ptyprocess==0.6.0
--- a/tron/core/actionrun.py
+++ b/tron/core/actionrun.py
@@ -18,6 +18,7 @@ from twisted.internet.base import DelayedCall
 
 from tron import command_context
 from tron import node
+from tron import prom_metrics
 from tron.actioncommand import ActionCommand
 from tron.actioncommand import NoActionRunnerFactory
 from tron.actioncommand import SubprocessActionRunnerFactory
@@ -541,6 +542,9 @@ class ActionRun(Observable, Persistable):
 
         if not self.machine.check("start"):
             return False
+
+        executor = self.executor or "ssh"
+        prom_metrics.tron_action_runs_created_counter.labels(executor=str(executor)).inc()
 
         if len(self.attempts) == 0:
             log.info(f"{self} starting")
--- a/tron/core/job_scheduler.py
+++ b/tron/core/job_scheduler.py
@@ -3,6 +3,7 @@ import logging
 import humanize
 from twisted.internet import reactor
 
+from tron import prom_metrics
 from tron.core import recovery
 from tron.core.job import Job
 from tron.core.jobrun import JobRun
@@ -162,6 +163,15 @@ class JobScheduler(Observer):
         if not self.job.allow_overlap and any(self.job.runs.get_active(node)):
             self._queue_or_cancel_active(job_run)
             return
+
+        # If we're actually going to run the job (not cancel or queue it)
+        if (
+            not run_queued
+            and job_run.is_scheduled
+            and (self.job.allow_overlap or not any(self.job.runs.get_active(node)))
+        ):
+            # Increment job run counter when we're actually starting a run
+            prom_metrics.tron_job_runs_created_counter.inc()
 
         job_run.start()
         self.schedule_termination(job_run)
--- a/tron/mcp.py
+++ b/tron/mcp.py
@@ -1,10 +1,15 @@
 import logging
 import time
 from contextlib import contextmanager
+from typing import Optional
+
+from prometheus_client import Gauge
+from prometheus_client import Histogram
 
 from tron import actioncommand
 from tron import command_context
 from tron import node
+from tron import prom_metrics
 from tron.config import manager
 from tron.config.schema import MASTER_NAMESPACE
 from tron.core.job import Job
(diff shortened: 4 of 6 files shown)


tron/prom_metrics.py
```diff
@@ -1,17 +1,111 @@
...
+#
+# Total Startup Time (tron_last_startup_duration_seconds)
+#   |-- _load_config()
+#   |-- Total Restore Time (tron_last_restore_duration_seconds)
+#   |     |-- state_watcher.restore() (tron_last_dynamodb_data_retrieval_duration_seconds_gauge)
+#   |     |-- jobs.restore_state() (tron_last_job_state_application_duration_seconds_gauge)
+#   |     |-- overhead within restore_state()
+#   |-- jobs.run_queue_schedule()
+#   |-- initial_setup()
+tron_restore_duration_seconds_histogram = Histogram(
```
KaspariK, line 75: It's possible we don't care too much about histogram versions of these, but it is nice to see an immediate change in grafana instead of relying on splunk shenanigans later on.

requirements.txt
```diff
@@ -51,7 +51,7 @@ oauthlib==3.1.0
 parso==0.7.0
 pexpect==4.7.0
 pickleshare==0.7.5
-prometheus-client==0.20.0
+prometheus-client==0.21.1
```
KaspariK, line 54: Not explicitly necessary, but I figured why not

tron/mcp.py
```diff
@@ -79,23 +96,55 @@ def _load_config(self, reconfigure=False, namespace_to_reconfigure=None):
                 namespace_to_reconfigure=namespace_to_reconfigure,
             )
 
+    def _update_metrics(self):
```
nemacysts, line 99:
```suggestion
    def _update_metrics(self) -> None:
```
non-blocking nit since i'm trying to add types to all of tron :p

tron/mcp.py
```diff
@@ -20,15 +25,27 @@
 
 
 @contextmanager
-def timer(function_name: str):
-    start = time.time()
+def timer(
+    operation_name: str,
+    histogram_metric: Optional[Histogram] = None,
+    gauge_metric: Optional[Gauge] = None,
+):
```
nemacysts, line 32:
```suggestion
) -> None:
```
non-blocking nit since i'm trying to add types to all of tron :p

nemacysts (approved):
i have some suggestions, but they're all technically optional :p

tron/core/actionrun.py
```diff
@@ -542,6 +543,9 @@ def start(self, original_command=True) -> Optional[Union[bool, ActionCommand]]:
         if not self.machine.check("start"):
             return False
 
+        executor = self.executor or "ssh"
+        prom_metrics.tron_action_runs_created_counter.labels(executor=str(executor)).inc()
```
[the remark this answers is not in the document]
EmanElsaban: good point, I assume by the description here 
```
tron_action_runs_created_counter = Counter(
    "tron_action_runs_created", "Total number of ActionRuns created", ["executor"]
)

```
is for total number of ActionRuns actually created , so we could leave it here?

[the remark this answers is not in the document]
EmanElsaban: I wonder if we should have another metric for "tron_valid_action_runs" that would only count the valid ones in addition to the total number of actionruns created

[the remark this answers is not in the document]
EmanElsaban: I added an additional metric "tron_action_runs_valid_counter" , it would be nice to be able to distinguish the valid action runs from the total ones created

tron/core/actionrun.py
```diff
@@ -542,6 +543,9 @@ def start(self, original_command=True) -> Optional[Union[bool, ActionCommand]]:
         if not self.machine.check("start"):
             return False
 
+        executor = self.executor or "ssh"
```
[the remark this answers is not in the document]
EmanElsaban: yea not sure why its needed, it should be defined in the constructor for ActionRun class
``
        self.executor = executor
``

tron/core/job_scheduler.py
```diff
@@ -163,6 +164,15 @@ def run_job(self, job_run, run_queued=False):
...
             return
 
+        # If we're actually going to run the job (not cancel or queue it)
+        if (
+            not run_queued
+            and job_run.is_scheduled
+            and (self.job.allow_overlap or not any(self.job.runs.get_active(node)))
+        ):
+            # Increment job run counter when we're actually starting a run
+            prom_metrics.tron_job_runs_created_counter.inc()
```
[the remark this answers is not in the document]
EmanElsaban: moved it to start in jobrun.py

tron/core/actionrun.py
```diff
@@ -564,6 +566,8 @@ def start(self, original_command=True) -> Optional[Union[bool, ActionCommand]]:
             self.fail(exitcode.EXIT_INVALID_COMMAND)
             return None
 
+        # Increment the valid actionRuns created
+        prom_metrics.tron_action_runs_valid_counter.labels(executor=str(executor)).inc()
```
nemacysts, line 570: this isn't necessarily guaranteed to give us the count of valid actionruns: https://github.com/Yelp/Tron/blob/eeb416ec3bf865ac77438c94de3aaa05cd7345bf/tron/core/actionrun.py#L1282 has a couple cases where we won't create a task for various reasons, for example

my suggestion was more along the lines of having the submit_command() function for the various ActionRun subclasses increment this counter rather than doing that here (or just living with the potential mismatch for now by having a single counter that we increment here (without trying to get too deep into whether or not start() actually started a task)

tron/mcp.py
```diff
@@ -20,15 +25,27 @@
...
     finally:
-        end = time.time()
-        log.info(f"Execution time for function {function_name}: {end-start}")
+        end_time = time.time()
+        duration = end_time - start_time
+        if histogram_metric:
+            histogram_metric.observe(duration)
+        if gauge_metric:
+            gauge_metric.set(duration)
+        log.info(f"Execution time for {operation_name}: {duration:.2f}s")
```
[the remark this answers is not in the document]
KaspariK: Hmmm, maybe a first slice is just having this in prom_metrics.py and moving it into a timing util down the line if we care to? At least this way it's close to the metrics it's designed to work with

tron/core/actionrun.py
```diff
@@ -542,6 +543,9 @@ def start(self, original_command=True) -> Optional[Union[bool, ActionCommand]]:
         if not self.machine.check("start"):
             return False
 
+        # Increment all the actionRuns created
+        prom_metrics.tron_action_runs_created_counter.labels(executor=self.executor).inc()
+
```
nemacysts, line 548: we should remove this inc() now, right?

tron/core/actionrun.py
```diff
@@ -597,6 +606,9 @@ def _done(self, target, exit_status=0) -> Optional[bool]:
             self.end_time = timeutils.current_time()
             if self.last_attempt is not None and self.last_attempt.end_time is None:
                 self.last_attempt.exit(exit_status, self.end_time)
+
+            prom_metrics.tron_action_runs_completed_counter.labels(executor=self.executor, outcome=target).inc()
```
nemacysts, line 610: thoughts on also adding exit_status? not sure if that would be useful or not

nemacysts (dismissed):

PR #1047, opened 2025-05-19
Attempt to type more of Tron

nemacysts:
There's still over a thousand failures with the desired end-state mypy configuration, but this should get us closer to a fully typed Tron


nemacysts:
lemme try to do this more incrementally

PR #1048, opened 2025-05-20
add support for api auth via vault

piax93:
This is on the same exact line of https://github.com/Yelp/paasta/pull/4053, since I found a case of tronctl being used in an unattended way.

I'll soon integrate the logic from `get_vault_auth_token` into a function in vault-tools, so that complexity goes to live somewhere else, but for the time being I did some simply copy-paste.

### DIFF
--- a/tron/commands/authentication.py
+++ b/tron/commands/authentication.py
@@ -1,0 +1,70 @@
+import os
+from typing import cast
+
+from botocore.credentials import InstanceMetadataFetcher
+from botocore.credentials import InstanceMetadataProvider
+
+from tron.commands.cmd_utils import get_client_config
+
+try:
+    from vault_tools.paasta_secret import get_client as get_vault_client  # type: ignore
+    from vault_tools.paasta_secret import get_vault_url
+    from vault_tools.paasta_secret import get_vault_ca
+    from okta_auth import get_and_cache_jwt_default  # type: ignore
+except ImportError:
+
+    def get_vault_client(url: str, capath: str) -> None:
+        pass
+
+    def get_vault_url(ecosystem: str) -> str:
+        return ""
+
+    def get_vault_ca(ecosystem: str) -> str:
+        return ""
+
+    def get_and_cache_jwt_default(client_id: str) -> str:
+        return ""
+
+
+def get_current_ecosystem() -> str:
+    """Get current ecosystem from host configs, defaults to dev if no config is found"""
+    try:
+        with open("/nail/etc/ecosystem") as f:
+            return f.read().strip()
+    except OSError:
+        pass
+    return "devc"
+
+
+def get_sso_auth_token() -> str:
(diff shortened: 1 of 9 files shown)


danielpopskandor (approved):

danielpopskandor (approved):

danielpops (dismissed):

tron/commands/authentication.py
```diff
@@ -0,0 +1,70 @@
...
+        return ""
+
+    def get_vault_ca(ecosystem: str) -> str:
+        return ""
+
+    def get_and_cache_jwt_default(client_id: str) -> str:
+        return ""
+
+
+def get_current_ecosystem() -> str:
```
nemacysts, line 29: could probably lru_cache this :)

tron/commands/authentication.py
```diff
@@ -0,0 +1,70 @@
+import os
+from typing import cast
+
+from botocore.credentials import InstanceMetadataFetcher
+from botocore.credentials import InstanceMetadataProvider
+
+from tron.commands.cmd_utils import get_client_config
+
+try:
+    from vault_tools.paasta_secret import get_client as get_vault_client  # type: ignore
```
nemacysts, line 10: hmm, i assume this is 'cause we don't have a py.typed marker in vault_tools and thus mypy isn't parsing the types?

nemacysts: or because `hvac` is not typed?

nemacysts: (might be worth commenting the `# type: ignore`s with the reasoning behind them - e.g., `# type: ignore  # foo 'cause bar`)

tron/commands/authentication.py
```diff
@@ -0,0 +1,70 @@
...
+            return f.read().strip()
+    except OSError:
+        pass
+    return "devc"
+
+
+def get_sso_auth_token() -> str:
+    """Generate an authentication token for the calling user from the Single Sign On provider, if configured"""
+    client_id = get_client_config().get("auth_sso_oidc_client_id")
+    return get_and_cache_jwt_default(client_id, refreshable=True) if client_id else ""  # type: ignore
```
nemacysts, line 42: we could probably replace the ignore with a cast() - but this is fine too

nemacysts (approved):
i think ideally we'd do some pod identity shenanigans to avoid building anything on top of the ssh executors since those are slated for deletion -  but we'll likely need to deal with the long tail of stuff still on ssh executors once we've moved the bulk of the workloads off of batch boxes/ssh executor (i.e., spark), and it probably wouldn't be too hard to swap things out for your alternative approach(es): i.e., add tronctl into service images + an IAM role to associate to the action that we can allow in the API authz or even your `external-publish` meta-executor idea

nemacysts (approved):

PR #1049, opened 2025-05-27
fix some annoying warnings

piax93:
Hadn't noticed, but changes from https://github.com/Yelp/Tron/pull/1048 introduced some annoying warnings for the CLI, and this couple of tweaks should fix them.

```
/opt/venvs/tron/lib/python3.8/site-packages/requests/__init__.py:102: RequestsDependencyWarning: urllib3 (1.26.5) or chardet (5.2.0)/charset_normalizer (2.0.12) doesn't match a supported version!
  warnings.warn("urllib3 ({}) or chardet ({})/charset_normalizer ({}) doesn't match a supported "
:0: UserWarning: You do not have a working installation of the service_identity module: 'No module named 'service_identity''.  Please install it from <https://pypi.python.org/pypi/service_identity> and make sure all of its dependencies are satisfied.  Without the service_identity module, Twisted can perform only rudimentary TLS client hostname verification.  Many valid certificate/hostname mappings may be rejected.
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -84,7 +84,7 @@ task-processing==1.3.5
 traitlets==5.0.0
 Twisted==22.10.0
 typing-extensions==4.5.0
-urllib3==1.26.5
+urllib3==1.25.10
 wcwidth==0.1.7
 websocket-client==0.56.0
 Werkzeug==2.2.3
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -84,7 +84,7 @@ task-processing==1.3.5
 traitlets==5.0.0
 Twisted==22.10.0
 typing-extensions==4.5.0
-urllib3==1.26.5
+urllib3==1.25.10
```
nemacysts, line 87: it probably wouldn't be too hard to update `requests`, but we can also punt on this :p

nemacysts (approved):

PR #1055, opened 2025-06-18
Add disk eviction error code and check logic

KaspariK:
## Problem
Tron currently thinks any pods evicted for disk reasons are spot terminated, which confuses folks who don't know what to look for. The new error -13 (spooky 🎺💀) will catch disk evictions.

## Solution
- Updated the current logic to look for an eviction message
- Fixed some typos

Tested in infrastage with a job that gets evicted and it works. I was also planning to include a hack to get this to the UI but I think I'll leave that to another time because it's kind of gross.

### DIFF
--- a/tron/kubernetes.py
+++ b/tron/kubernetes.py
@@ -92,10 +92,8 @@ class KubernetesTask(ActionCommand):
         """
         Update internal resource utilization statistics of all tronjobs running for this task's Tron master.
         """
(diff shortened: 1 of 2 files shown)


nemacysts (approved):
one small nit - but lgtm otherwise (other than handle_event() needing to be refactored, but that was true beforehand :p

nemacysts (approved):

PR #1057, opened 2025-06-20
Allow transitioning to success/fail from any terminal state

nemacysts:
This is helpful when folks have handled failures outside of Tron (or have done something outside of Tron that nominally means that the Tron status is now incorrect) OR to work-around any weirdness with Tron states (e.g., something funky happened in k8s and Tron never saw the event)

NOTE: this PR changes a bunch of extra files - but that's just due to mypy following more code paths after I typed the ActionRunController `__init__` so that I could go-to-def in my editor :p 
Additionally, kkasp and I have occasionally been runing into an state where `check-requirements` says a mismatched version of `cryptography` is installed - I finally got around to debugging this and realized this stems from our very hacky extra_requirements_yelp.txt shenanigans: `pyopenssl` has lower/upper-bounds on `cryptography` but, due to the ordering we install things in, `check-requirements` might run before we've installed `pyopenssl` and downgraded `cryptography` (and pass) or after (and fail)

The real changes are confined to
`tron/core/action.py::ActionRun::STATE_MACHINE` (and a now-deleted ActionRun tests that is no longer valid)


requirements.txt
```diff
@@ -16,7 +16,7 @@ cffi==1.12.3
 cfn-lint==0.24.4
 charset-normalizer==2.0.12
 constantly==15.1.0
-cryptography==39.0.1
+cryptography==41.0.5
```
nemacysts, line 19: this is the version we've actually been using internally

requirements.txt
```diff
@@ -63,9 +63,10 @@ pycparser==2.19
 pyformance==0.4
 Pygments==2.13.0
 pymesos==0.3.9
+pyopenssl==24.2.1
```
nemacysts, line 66: moved from extra_requirements_yelp and upgraded a tad to include type stubs

requirements.txt
```diff
@@ -63,9 +63,10 @@ pycparser==2.19
...
 pyrsistent==0.15.4
 pysensu-yelp==0.4.4
-PyStaticConfiguration==0.10.4
+PyStaticConfiguration==0.11.1
```
nemacysts, line 69: upgraded to include type stubs

tests/core/actionrun_test.py
```diff
@@ -1932,7 +1928,7 @@ def test_recover_done_no_change(
         last_attempt.kubernetes_task_ic = "test-kubernetes-task-id"
 
         assert not mock_k8s_action_run.recover()
-        assert mock_cluster_repo.get_cluster.call_count == 0
+        assert mock_cluster_repo.get_cluster.call_count != 0
```
nemacysts, line 1931: not sure why this test just started failing in this branch, but this has been very wrong until now

tron/core/actionrun.py
```diff
@@ -257,47 +257,64 @@ class ActionRun(Observable, Persistable):
     WAITING = "waiting"
     UNKNOWN = "unknown"
 
-    default_transitions = dict(fail=FAILED, success=SUCCEEDED)
+    default_transitions = {"fail": FAILED, "success": SUCCEEDED}
```
nemacysts, line 260: i took the liberty of making these inline dicts to make it easier to update things (and 'cause it's 2025 - we shouldn't be using the dict() constructor anymore)

PR #1058, opened 2025-06-23
Drop all traces of bionic

nemacysts:
We don't need to be building for bionic anymore, so let's get rid of anything that was previously used just for bionic.


tox.ini
```diff
@@ -45,18 +45,7 @@ commands=
...
     sphinx-build -b html -d docs/_build docs/source docs/_build/html
 
-[testenv:example-cluster]
-whitelist_externals=docker-compose
-deps=
-    docker-compose
-commands=
-    docker-compose -f example-cluster/docker-compose.yml build playground
-    docker-compose -f example-cluster/docker-compose.yml run -p 8089:8089 playground
-    docker-compose -f example-cluster/docker-compose.yml down
```
nemacysts, line 55: this was cruft since the docker-compose file was empty

PR #1060, opened 2025-06-23
Start running mypy over all of Tron

nemacysts:
We previously only mypy'd a small subset of (mostly) fully typed files, but we should probably start expanding the checked files - as well as incrementally tightening up out mypy settings.

This PR enables type-checking all of Tron with the existing settings as well as includes the disabled checks that I'll be attmepting to work through during our unhack this week.


nemacysts:
of course this has failures in GHA but not locally -_-

requirements.txt
```diff
@@ -26,7 +26,7 @@ executing==1.2.0
 future==0.18.3
 google-auth==1.23.0
 http-parser==0.9.0
-humanize==0.5.1
+humanize==4.10.0
```
nemacysts, line 29: this is only used for logging, but afaict, there's not breaking changes

requirements.txt
```diff
@@ -64,8 +64,8 @@ pyformance==0.4
 Pygments==2.13.0
 pymesos==0.3.9
 pyrsistent==0.15.4
-pysensu-yelp==0.4.4
-PyStaticConfiguration==0.10.4
+pysensu-yelp==1.0.3
+PyStaticConfiguration==0.11.1
```
nemacysts, line 68: bumped for included type hints

yelp_package/extra_requirements_yelp.txt
```diff
@@ -15,7 +15,6 @@ pyjwt==2.9.0               # required by okta-auth
 pyopenssl==23.2.0          # vault-tools dependency
 saml-helper==2.5.3         # required by okta-auth
 service-identity==24.2.0   # vault-tools dependency
-simplejson==3.19.2         # required by tron CLI
```
nemacysts, line 18: we don't support python2, we don't need simplejson anymore and can just rely on the stdlib

tron/utils/logreader.py
```diff
@@ -8,19 +8,29 @@
...
 
 from tron.config.static_config import get_config_watcher
 from tron.config.static_config import NAMESPACE
 
 try:
-    from logreader.readers import S3LogsReader
+    # mypy: allow-unused-ignore
+    from logreader.readers import S3LogsReader  # type: ignore[import-not-found]  # internal-only package
+
+    # mypy: no-unused-ignore
```
nemacysts, line 21: the joy of internal packages in an OSS repo

requirements.txt
nemacysts: (and no breaking changes from what i can see in the commit history)

tron/utils/logreader.py
```diff
@@ -15,12 +15,22 @@
...
 except ImportError:
     s3reader_available = False
 
+    class S3LogsReader:  # type: ignore[no-redef]  # stub class for internal-only package
+        def __init__(self, superregion: str):
+            raise ImportError("logreader (internal Yelp package) is not available - unable to display logs.")
+
+        def get_log_reader(self, log_name: str, start_datetime: datetime.datetime, end_datetime: datetime.datetime):
+            raise NotImplementedError("logreader (internal Yelp package) is not available - unable to display logs.")
+
```
EvanKrall, line 33: We're just defining this so that mypy doesn't complain about it not existing?

EvanKrall (approved):

tron/config/manager.py
```diff
@@ -193,11 +195,14 @@ def get_hash(self, name) -> str:
             # TODO: consider storing the hash alongside the config so that we only calculate
             # hashes once?
             return hash_digest(
-                yaml.dump(
-                    self.get_config_name_mapping()[name],
-                    # ensure that the keys are always in a stable order
-                    sort_keys=True,
-                ),
+                cast(
+                    Union[str, bytes],
```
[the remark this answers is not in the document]
nemacysts: uhhh, lemme try this out again - iirc, i think it was something do to with hash_digest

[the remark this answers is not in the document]
nemacysts: neat - i think the issue was with a corrupted/outdated venv or mypy cache: this doesn't seem to be necessary anymore!

EvanKrall (approved):

PR #1063, opened 2025-06-26
U/kkasp/tron 2240 delete pickles

KaspariK:
## Problem
We've been saving both pickles and JSON as part of our "safe migration" off of pickle data. Before we can remove pickles (and ultimately delete Mesos + hella old things) we need to stop saving pickles. This change removes the writing pipeline for pickles.

## Solution
Generally, just stop writing pickles. This includes:
- Removing read_json toggles from everywhere except the actual config schema
- Upping the partition sizes now that we only write half the data we used to

Some additional "just 'cause" changes:
- Updating moto to a version that has actual transact_write_items support so we stop rolling our own
- Refactor basically all tests for DynamoDB state
- Add some more explicit failures to the restore path that _should_ prevent Tron amnesia state (run nums resetting to 0)

**Note:** I'll test restoring on some more tables, but so far so good. I think it's pretty reviewable as is. Ideally we do some more read_json cleanup, but for now this is a safe approach. One annoying thing is writing to json_val and num_json_partitions, but that's just the way the cookie crumbles.

**Extra note:** This will not be rollback safe (we can technically rollback with restored data, but generally I think it's best to consider this release not pleasantly undone). Full steam ahead 🚂.

### DIFF
--- a/tron/serialize/runstate/dynamodb_state_store.py
+++ b/tron/serialize/runstate/dynamodb_state_store.py
@@ -1,8 +1,10 @@
 import concurrent.futures
 import copy
+import json
 import logging
 import math
 import os
+import sys
 import threading
 import time
 from collections import defaultdict
@@ -176,37 +178,56 @@ class DynamoDBStateStore:
 
             remaining_items = [
                 {"key": {"S": str(item["key"]["S"])}, "index": {"N": str(i)}}
-                # we start from 1 since we already have the 0th partition from get_first_partitions KKASP: confirm?
+                # we start from 1 since we already have the 0th partition from get_first_partitions
                 for i in range(1, num_partitions)
             ]
             keys_for_remaining_items.extend(remaining_items)
         return self._get_items(keys_for_remaining_items)
 
     def _merge_items(self, first_items, remaining_items) -> dict:
-        """Helper to merge multi-partition data into a single entry."""
+        """
+        Reassembles partitioned data and calls the deserializer for each object.
+
+        DynamoDB items that exceed the OBJECT_SIZE limit are split into multiple
+        partitions and stored as separate items in the table. Each partition
+        shares the same primary key but has a unique sort key ('index').
+
+        This function prepares the raw data for final deserialization by:
+        1.  Grouping all fetched partitions by their primary key.
+        2.  For each key, sorting the partitions by their index and concatenating
+            their data into a single, complete JSON string.
+        3.  Passing this final JSON string to `_deserialize_item`, which
+            handles the conversion to a specific Python state object.
+        """
         items = defaultdict(list)
         json_items: DefaultDict[str, str] = defaultdict(str)
-        # item = job run, each job run can have multiple rows and each row is called a partition
-        # Merge all items based their keys and deserialize their values
+
         if remaining_items:
             first_items.extend(remaining_items)
         for item in first_items:
             key = item["key"]["S"]
             items[key].append(item)
+
+        # Concatenate all partitions for each key
         for key, item in items.items():
             item.sort(key=lambda x: int(x["index"]["N"]))
-            for json_val in item:
+            for partition_item in item:
                 try:
-                    json_items[key] += json_val["json_val"]["S"]
-                except Exception:
-                    # KKASP: what should we do here? If we can't read the json_val we are cooked
-                    log.exception(f"json_val not found for key {key}")
+                    json_items[key] += partition_item["json_val"]["S"]
+                except KeyError:
+                    log.critical(
+                        f"Partition item for key '{key}' at index {partition_item['index']['N']} is missing 'json_val'. Shutting down due to data corruption."
+                    )
+                    sys.exit(1)
+
+        # Deserialize the concatenated JSON strings into Python objects
         try:
             deserialized_items = {k: self._deserialize_item(k, val) for k, val in json_items.items()}
-        except Exception as e:
-            # KKASP: what should we do here? If we can't read the json_val we are cooked
-            log.exception(f"Error deserializing JSON items: {repr(e)}")
-            raise
+        except (json.JSONDecodeError, ValueError):
+            log.critical(
+                "Failed to deserialize concatenated JSON string. This indicates a truncated or malformed record. Shutting down due to data corruption."
+            )
+            sys.exit(1)
 
         return deserialized_items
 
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -36,17 +29,12 @@ ipython-genutils==0.2.0
...
-jsonpointer==2.0
-jsonschema==3.2.0
 kubernetes==26.1.0
 lockfile==0.12.2
 MarkupSafe==2.1.1
 matplotlib-inline==0.1.3
-mock==3.0.5
-moto==1.3.13
+more-itertools==10.5.0
+moto==2.0.8
```
KaspariK, line 37: This gets us transact_write_items. There is probably value to going a bit further since this is still quite old and some of these older versions have older limits (e.g. a newer version has support for 100 items in a transaction, which we recently removed from Tron, though I'm sure there are other things :shrug:).

tests/serialize/runstate/dynamodb_state_store_test.py
```diff
@@ -1,108 +1,50 @@
...
+        # Pass `stopping=True` to prevent the background save thread from running during tests
+        state_store = DynamoDBStateStore(table_name, "us-west-2", stopping=True)
+        yield state_store
 
 
 @pytest.fixture
-def small_object():
-    yield {
+def job_state_object():
+    return {"enabled": True, "run_nums": [0, 1]}
```
nemacysts, line 42: it's nice to always yield in fixtures since that's one less change to make if we ever need some post-fixture cleanup - but this is also fine as-is :)

tests/serialize/runstate/dynamodb_state_store_test.py
```diff
@@ -115,221 +57,135 @@ def small_object():
...
-        "runs": [],
         "cleanup_run": None,
         "manual": False,
-        "large_data": [i for i in range(1_000_000)],
+        "runs": [base_run] * 5000,  # gives us a ~2000KB JobRun, ensuring at least a few partitions.
     }
 
 
-@pytest.mark.usefixtures("store", "small_object", "large_object")
+@pytest.mark.usefixtures("store", "job_state_object", "small_jobrun_state_object", "large_jobrun_state_object")
```
nemacysts, line 93: just curious (in case you know): do we need pytest.mark.usefixtures here? seems like we actually refer to the fixtures in our tests and, outside of the `store` fixture, the other fixtures only return data

tests/serialize/runstate/dynamodb_state_store_test.py
```diff
@@ -115,221 +57,135 @@ def small_object():
...
-        ]
-        store.save(delete)
-        store._consume_save_queue()
+    @pytest.mark.parametrize(
+        "object_fixture_name",
+        ["small_jobrun_state_object", "large_jobrun_state_object"],
+    )
+    def test_save_and_restore(self, store, object_fixture_name, request):
+        """Verify that objects of different sizes can be saved and restored correctly."""
+        state_object = request.getfixturevalue(object_fixture_name)
```
nemacysts, line 101:
```suggestion
    @pytest.mark.parametrize(
        "state_object",
        [small_jobrun_state_object, large_jobrun_state_object],
    )
    def test_save_and_restore(self, store, state_object, request):
        """Verify that objects of different sizes can be saved and restored correctly."""
```

disclaimer: i haven't test that - but shouldn't we be able to use the fixtures directly here rather than using the `getfixturevalue` indirection?

tests/serialize/runstate/dynamodb_state_store_test.py
```diff
@@ -115,221 +57,135 @@ def small_object():
...
-        store.save(pairs)
+        assert restored_vals[key] == state_object
+
+        num_partitions = store._get_num_of_partitions(key)
+        if object_fixture_name == "small_jobrun_state_object":
+            assert num_partitions == 1
+        else:
+            # For large objects, we expect multiple partitions, the number of partitions can vary depending on our object size
+            # and the DynamoDB partition size limit. We assert that there is more than one partition.
+            assert num_partitions > 1
```
nemacysts, line 118: we should try to avoid conditionals in tests - that said, not sure how to handle this specific case without having one of the parameterized values to pass through be a lambda or something :p

tests/serialize/runstate/dynamodb_state_store_test.py
```diff
@@ -345,33 +201,32 @@ def test_retry_saving(
...
-        unprocessed_value = {
+    def test_retry_reading(self, store, table_name):
+        """Verify that reading from DynamoDB retries when there are unprocessed keys."""
+        key = store.build_key("job_state", 0)
+        unprocessed_keys_response = {
             "Responses": {},
             "UnprocessedKeys": {
-                store.name: {
-                    "Keys": [{"key": {"S": store.build_key("job_state", 0)}, "index": {"N": "0"}}],
+                table_name: {
```
nemacysts, line 212: it's probably more future-proof to keep using `store.name` (and would remove a fixture from this test) - but this isn't the end of the world either :p

tron/serialize/runstate/dynamodb_state_store.py
```diff
@@ -81,7 +77,8 @@ def build_key(self, type, iden) -> str:
         """
         return f"{type} {iden}"
 
-    def restore(self, keys, read_json: bool = False) -> dict:
+    # KKASP: removed read_json so I think we can remove this parameter from the config
```
nemacysts, line 80:
```suggestion
```

tron/serialize/runstate/dynamodb_state_store.py
```diff
@@ -171,71 +168,67 @@ def _get_first_partitions(self, keys: list):
...
-                deserialized_items = {k: self._deserialize_item(k, val) for k, val in json_items.items()}
-            except Exception as e:
-                log.exception(f"Error deserializing JSON items: {repr(e)}")
-                # if reading from json failed we want to try reading the pickled data instead
-                read_json = False
+            for partition_item in item:
+                try:
+                    json_items[key] += partition_item["json_val"]["S"]
+                except KeyError:
+                    log.critical(
```
nemacysts, line 218: does log.critical print the traceback if it exists? if not, might be nice to change this to log.exception() to ensure that folks have a line number to jump to from the traceback :)

tron/serialize/runstate/dynamodb_state_store.py
```diff
@@ -248,12 +241,13 @@ def save(self, key_value_pairs) -> None:
                     time.sleep(5)
                     continue
                 with self.save_lock:
+                    # KKASP: what is this again?
```
nemacysts, line 244:
```suggestion
                    # TODO: what is this again?
```

nemacysts: and isn't this the delete shenanigans we do in tron?

PR #1067, opened 2025-07-23
Default sliders to show all items in model

nemacysts:
Defaulting to 10 doesn't actually save any time since we've fetched all the data anyway.

This will make pages longer, but it will be less confusing to folks when they get alerts for failed/unknown/etc jobs that are past the visible runs with the previous slider configuration.

### DIFF
--- a/tronweb/coffee/views.coffee
+++ b/tronweb/coffee/views.coffee
@@ -246,7 +246,8 @@ class module.SliderView extends Backbone.View
 
     initialize: (options) ->
         options = options || {}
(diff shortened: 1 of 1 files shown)


nemacysts:
(i tested this locally with a job i ran 50 times and loading the job page showed all 50 runs rather than just the last 10 by default - and there were no visible console errors)

KaspariK (approved):
Wonder what this looks like on those 1000 run_limit jobs...

KaspariK (approved):

PR #1071, opened 2025-08-28
Upgrade Tron to Python 3.10 - TRON-2435

EmanElsaban:
This PR upgrades Tron to Python 3.10


tests/serialize/runstate/dynamodb_state_store_test.py
```diff
@@ -125,7 +125,7 @@ def large_object():
         "runs": [],
         "cleanup_run": None,
         "manual": False,
-        "large_data": [i for i in range(1_000_000)],
+        "large_data": [i for i in range(10000)],
```
nemacysts, line 128: was this meant to be changed?

requirements-minimal.txt
```diff
@@ -11,6 +11,7 @@ ipdb
 ipython
 Jinja2>=3.1.2
 lockfile
+mock==3.0.5
```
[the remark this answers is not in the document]
EmanElsaban: ops I forgot to remove this from the minimal file

requirements-minimal.txt
```diff
@@ -11,6 +11,7 @@ ipdb
...
 moto
```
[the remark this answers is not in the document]
EmanElsaban: moved it to requirements-dev & requirments-dev-minimal files with its dependencies "responses" and "xmltodict" packages

tests/serialize/runstate/dynamodb_state_store_test.py
EmanElsaban: yea, I was getting this error
``
ClientError('An error occurred (ValidationException) when calling the TransactWriteItems operation: Item size has exceeded the maximum allowed size')
``
moto's mock enforces a 400kb item size limit and the 1000000 serializes to a json object greater than 400kb. I think older versions of moto didn't enforce this size limit and now that we upgraded moto it enforces it. So I reduced the number of integers we create here to be below that limit

nemacysts: gotcha! i assume that since the tests are still passing, then we're still ending up
splitting this object

tron/utils/crontab.py
```diff
@@ -46,20 +41,20 @@ class FieldParser:
     def normalize(self, source: str) -> str:
         return source.strip()
 
-    def get_groups(self, source: str) -> List[str]:
+    def get_groups(self, source: str) -> list[str]:
         return source.split(",")
 
-    def parse(self, source: str) -> Optional[Union[List[int], List[Union[int, str]]]]:
+    def parse(self, source: str) -> list[int] | list[int | str] | None:
```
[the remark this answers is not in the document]
EmanElsaban: okay will leave it as it is now then

nemacysts (dismissed):

tests/serialize/runstate/dynamodb_state_store_test.py
KaspariK: I suspect that by reducing the data to 10,000 integers, we're now creating an object that is smaller than our OBJECT_SIZE of 200KB. This would mean it no longer gets partitioned, and these tests are no longer verifying the multi-partition logic they were designed for.

Maybe we should also add an assert num_partitions(key) > 1 in these tests to guarantee we're actually testing the partitioning logic?

Also, outside the bounds of this PR, it might be a good time to rename these tests to something like test_save_object_requiring_partitioning. `more_than_4KB` is _technically_ correct but confusing 😄

nemacysts (dismissed):
lgtm - probably want @KaspariK to ship as well wrt the dynamo bits

KaspariK (approved):

PR #1072, opened 2025-08-28
TRON-2477: Do some more validation of TimeSpecs

jfongatyelp:
We ran into an infinite loop in our current TimeSpecification.get_match when asked for the next 30th of februrary, for example. 

Other similar impossible job schedules such as April 31 would have run into the same issue.  I considered changing this in the schedule parsing logic, but this will catch if (in the extremely slim chance) someone also specified an impossible date  using our complex scheduling format (`groc daily`, `value: 30th day in feb`). 

In this PR, we now do not allow an invalid TimeSpecificaation to be created, and when initially scheduling the next job runs, we check whether we get a valid TimeSpecification back, and if not, simply print an error in our logs and skip that job in our JobCollection.

In general, I do think our yelpsoa-configs validation should prevent us from even getting here again and will provide a better interface to job owners that show the invalid config, but if config on disk/in memory was ever able to get into a bad state like this, these changes would allow tron to continue gracefully, simply skipping that job with an invalid schedule. In addition, config change requests would succeed if provided with a new, valid schedule. 

In addition to the TimeSpecification tests, I've tested locally and ensured if tron starts w/ an invalid schedule, it simply skips:
```
2025-08-28 13:12:22,169 tron.core.job_collection ERROR Failed to build job 'MASTER.testjob0': Day 30 does not exist in any specified month ([2]). Skipping job.
```
One unintuitive behavior as a result of this change is that the job will also no longer be shown in tronweb, though it persists in config, which could potentially lead to confusion.  That said, I think trying to add the job but ensure everything down the line can properly handle+display appropriate warnings when it has an impossible schedule may be more trouble than this edge case is worth, tbh. I'm happy to give it a shot though if ppl think the opposite.

### DIFF
--- a/tron/core/job_collection.py
+++ b/tron/core/job_collection.py
@@ -43,8 +43,15 @@ class JobCollection:
         # NOTE: as this is a generator expression, we will only go through job configs
         # and build a scheduler for them once something iterates over us (i.e, once
         # `self.state_watcher.watch_all()` is called)
-        seq = (factory.build(config) for config in job_configs.values() if reconfigure_filter(config))
-        return map_to_job_and_schedule(filter(self.add, seq))
+        def safe_build(config):
+            try:
+                return factory.build(config)
+            except (ValueError, Exception) as e:
+                log.error(f"Failed to build job '{config.name}': {e}. Skipping job.")
+                return None
+
+        seq = (safe_build(config) for config in job_configs.values() if reconfigure_filter(config))
+        return map_to_job_and_schedule(filter(self.add, filter(lambda x: x is not None, seq)))
 
     def add(self, job_scheduler):
         return self.jobs.add(job_scheduler, self.update)
--- a/tests/utils/trontimespec_test.py
+++ b/tests/utils/trontimespec_test.py
@@ -3,6 +3,7 @@ import datetime
 import pytz
 
 from testifycompat import assert_equal
+from testifycompat import assert_raises
 from testifycompat import run
 from testifycompat import TestCase
 from tron.utils import trontimespec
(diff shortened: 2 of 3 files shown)


tron/core/job_collection.py
```diff
@@ -43,8 +43,15 @@ def reconfigure_filter(config):
         # NOTE: as this is a generator expression, we will only go through job configs
         # and build a scheduler for them once something iterates over us (i.e, once
         # `self.state_watcher.watch_all()` is called)
-        seq = (factory.build(config) for config in job_configs.values() if reconfigure_filter(config))
-        return map_to_job_and_schedule(filter(self.add, seq))
+        def safe_build(config):
+            try:
+                return factory.build(config)
+            except (ValueError, Exception) as e:
+                log.error(f"Failed to build job '{config.name}': {e}. Skipping job.")
```
nemacysts, line 50: i think ideally we'd be able to see that this has happened during a config update and show the failure then (so that we're aware), but having the job disappear from tronweb seems fine as well - someone will complain relatively quickly if this is impacting them :p 

that said - when this happens: do we lose the previous runs in tronweb upon fixing the misconfiguration?

nemacysts: i do also kinda wonder what happens if there's a dependency on a job with this misconfiguration

nemacysts (commented):
generally lgtm, just two questions

tron/core/job_collection.py
jfongatyelp: > i think ideally we'd be able to see that this has happened during a config update and show the failure then (so that we're aware), but having the job disappear from tronweb seems fine as well - someone will complain relatively quickly if this is impacting them :p

So when I attempt to upload a config w/ a bad schedule, this does fail and api returns an error (though doesn't seem to generate any logs about the failure on the server side 🤔 ), so at least setup_tron_namespace should alert us if this happens.  I'll try to see how I can make sure the error is logged tho.

> that said - when this happens: do we lose the previous runs in tronweb upon fixing the misconfiguration?

Great point, will run some more tests on my local run to verify

> i do also kinda wonder what happens if there's a dependency on a job with this misconfiguration

For this one, I'm suspecting we'll be ok since I thought the way we do inter-job 'dependencies' were really just based on trigger names, rather than having any actual tie between the jobs/jobruns?  But I will definitely also verify this.

jfongatyelp:
closing as #1108 solves this at the scheduling side rather than the timespecification side.

PR #1099, opened 2026-04-16
WIP: tronweb2 Vite + React

cuza:
**Addition of `tronweb2`**

* Added a new frontend application under `tronweb-src` using React, TypeScript, Vite, Tailwind CSS, and modern UI libraries. This includes core app scaffolding, routing, components (e.g., `AppShell`, `CommandPalette`, `DataTable`), configuration files, and build scripts.

**Integration and Serving**

* Updated `tron/api/resource.py` to serve the new `tronweb2` static files at the `/tronweb` path, alongside the existing `/web` frontend.

**Building and Packaging**

* Modified the `Makefile` to add build commands for `tronweb2`, ensure it is built during package creation, and clean up its build artifacts.
* Updated `MANIFEST.in` and `debian/install` to include `tronweb2` in source distributions and installation.
* Updated `setup.py` to include `tronweb2` as a package.


Makefile
```diff
@@ -59,7 +59,12 @@ coffee_%: docker_%
 		coffee -o tronweb/js/cs/ -c tronweb/coffee/ \
 	'
 
+tronweb2:
+	@echo "Building tronweb2"
+	cd tronweb-src && npm i && npm run build
```
cuza, line 64: Maybe for peace of mind we should also version control the lockfile and have this be
```suggestion
	cd tronweb-src && npm ci && npm run build
```

tronweb-src/src/lib/queries.ts
cuza, line 1: I wonder how bad of an idea is to have these pollings:

| Query | Interval |
| --- | --- |
| Status | 30s |
| Job list | 10s |
| Job detail | 5s |
| Job run | 5s |
| Action run (running/starting) | 2s |
| Action run (finished) | none |
| Configs | none |

Makefile
nemacysts: lockfiles++

piax93: there's the internal / public registry problem

.gitignore
```diff
@@ -52,3 +52,6 @@ debian/tron.postrm.debhelper
 debian/tron.preinst.debhelper
 debian/tron.prerm.debhelper
 debian/tron.substvars
+
+# tronweb2
```
nemacysts:
```suggestion
```

Makefile
cuza: ^^ this!!, but the hash should match on internal and external versions (if not we have a bigger problem) :)

piax93: hashes match, but lockfiles also have full URLs, and those won't

cuza: iirc npm ci does not throw errors as long as hashes match

cuza: I did not include the lockfile since I was more worried about exposing internal endpoints to the outside world on a public repo

PR #1100, opened 2026-05-19
Fix cron validation to reject trailing garbage in fields

sidtuladhar:
The regex pattern in FieldParser used .match() semantics without an end anchor, allowing invalid expressions like "5kevin" or "*/10abc" to pass validation silently. Adding a $ anchor ensures the entire field must match the pattern.

Fixes TRON-1761

### DIFF
--- a/tests/utils/crontab_test.py
+++ b/tests/utils/crontab_test.py
@@ -151,6 +151,28 @@ class TestInvalidInputs(TestCase):
         with assert_raises(ValueError):
             self.parser.parse("61")
 
+    def test_trailing_garbage_rejected(self):
+        with assert_raises(ValueError):
+            self.parser.parse("5kevin")
+
+    def test_trailing_text_after_range_rejected(self):
+        with assert_raises(ValueError):
+            self.parser.parse("5-10x")
+
+    def test_trailing_text_after_step_rejected(self):
+        with assert_raises(ValueError):
+            self.parser.parse("*/10abc")
+
+
+class TestInvalidCrontabLines(TestCase):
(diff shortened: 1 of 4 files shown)


mbrankin-art (dismissed):

nemacysts:
@sidtuladhar i've added some of the CIB folks as reviewers as well :)

nemacysts:
that said, i'm not 100% sure that this will prevent the issue we saw: `10 14 15-21 * 5` is technically a valid cron schedule, it's just not something that tron handles (i assume 'cause it was a pain to implement when this code was initially written)

could you test running what `paasta validate` runs (https://github.com/Yelp/paasta/blob/master/paasta_tools/tron_tools.py#L1353) on a yaml file that looks like the one that broke sync-deploy using this change?

sidtuladhar:
@nemacysts  right, it wouldn't prevent that. When i run `paasta validate` on the yaml that has the broken cron syntax for tron, there are no errors

sidtuladhar:
this prevents monthday and weekday in the parsing now

tests/utils/crontab_test.py
```diff
@@ -151,6 +151,28 @@ def test_invalid_expression(self):
         with assert_raises(ValueError):
             self.parser.parse("61")
 
+    def test_trailing_garbage_rejected(self):
+        with assert_raises(ValueError):
+            self.parser.parse("5kevin")
```
KaspariK, line 156: I don't think I've ever been called trailing garbage before

KaspariK (approved):

PR #1101, opened 2026-05-27
build for resolute

benbariteau:
Two surprising-ish things:
- PyPI index URL needs to be migrated since resolute does not have a platform-specific one
- resolute ships with coffeescript v2, which luckily only had one issue needing fixing


tronweb/coffee/models.coffee
```diff
@@ -122,6 +122,7 @@ class ConfigIndexEntry extends IndexEntry
 
 class CommandIndexEntry extends IndexEntry
     constructor: (@name, @job_name, @action_name) ->
+        super name
```
KaspariK, line 125: I wasn't aware we had this feature. As is tradition, this hasn't worked correctly for ~8 years, lol. `@name = name` is overwriting things with window.name, so we aren't surfacing commands in the search.

Would you mind removing `@name = name` and making `@name` a normal parameter alongside your `super name` fix?

i.e.:
```
constructor: (name, @job_name, @action_name) ->
    super name
```

benbariteau: Lol no problem

tronweb/coffee/models.coffee
```diff
@@ -122,7 +122,7 @@ class ConfigIndexEntry extends IndexEntry
 
 class CommandIndexEntry extends IndexEntry
     constructor: (@name, @job_name, @action_name) ->
```
KaspariK: Just missing that name needs to be a plain parameter
```suggestion
    constructor: (name, @job_name, @action_name) ->
```

benbariteau: I don't really know what that means, but sure lol

KaspariK (approved):

PR #1102, opened 2026-06-17
Remove package.json

kedarv:
v10 is rather ancient, let's pin the the oldest available LTS instead


nemacysts:
@kedarv any context behind this change/have you been able to test this locally?

kedarv:
> @kedarv any context behind this change/have you been able to test this locally?

Added context in the summary, but just pinning the minimum version to a supported node LTS version. I'm actually not sure why this repo needs a package.json file to begin with, so not sure how to test it 😅

nemacysts:
> > @kedarv any context behind this change/have you been able to test this locally?
> 
> Added context in the summary, but just pinning the minimum version to a supported node LTS version. I'm actually not sure why this repo needs a package.json file to begin with, so not sure how to test it 😅

ah gotcha, we have a funky Coffeescript-based UI - i was gonna say that this was required for that...but after some git history spelunking: i think this is dead code leftover from when we had some eslint hooks for a React-based UI re-write

...or in other words: i think we can probably just delete this lol

KaspariK:
> > @kedarv any context behind this change/have you been able to test this locally?
> 
> Added context in the summary, but just pinning the minimum version to a supported node LTS version. I'm actually not sure why this repo needs a package.json file to begin with, so not sure how to test it 😅

Yeah @nemacysts and I are chatting and we think this is likely a relic of a React TronWeb.

At any rate, it's not relevant to how we build Tron today (seems we use node 12.22 in the build though :p)

*edit: Node 12.22 is in the Jammy image and used by the distro coffeescript toolchain

kedarv:
@nemacysts @KaspariK sounds good, deleted the file instead.

KaspariK (approved):

PR #1103, opened 2026-07-08
Increase default k8s disk size from 1024 to 4096MiB

CharlesPeacock:
Context: Default disk size in Tron is 1gb. This routinely results in storage issues for certain jobs which then leads to onpoint issues that we manually deal with. In an effort to mitigate the frequency of this, we've decided to bump the default to 4096. 

Tests Passed

### DIFF
--- a/tron/kubernetes.py
+++ b/tron/kubernetes.py
@@ -32,7 +32,7 @@ if TYPE_CHECKING:
     from tron.serialize.runstate.statemanager import StateChangeWatcher
 
 DEFAULT_POD_LAUNCH_TIMEOUT_S = 300  # arbitrary number, same as Mesos offer timeout of yore
-DEFAULT_DISK_LIMIT = 1024.0  # arbitrary, same as what was chosen for Mesos-based Tronjobs
+DEFAULT_DISK_LIMIT = 4096.0  # arbitrary, same as what was chosen for Mesos-based Tronjobs
 
 KUBERNETES_TASK_LOG_FORMAT = "%(asctime)s %(name)s %(levelname)s %(message)s"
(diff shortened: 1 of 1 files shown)


tron/kubernetes.py
```diff
@@ -32,7 +32,7 @@
     from tron.serialize.runstate.statemanager import StateChangeWatcher
 
 DEFAULT_POD_LAUNCH_TIMEOUT_S = 300  # arbitrary number, same as Mesos offer timeout of yore
-DEFAULT_DISK_LIMIT = 1024.0  # arbitrary, same as what was chosen for Mesos-based Tronjobs
+DEFAULT_DISK_LIMIT = 4096.0  # arbitrary, same as what was chosen for Mesos-based Tronjobs
```
nemacysts, line 35: I *think* that we also send this from paasta (in https://github.com/Yelp/paasta/blob/2c55307319826defc68d145abe3c6a45ed31f133/paasta_tools/tron_tools.py#L314-L314) - i don't remember in what cases this default is used (or if it's a leftover from an initial version of the k8s code)

nemacysts: (i.e., we should probably change both)

CharlesPeacock: Great observation. Now that you mention it we are also declaring a default here that seems to affect all instance types. Not setting a value for the override in tron_tools:314 would make us fallback to 1024MiB of disk for a given Tron instance.

[https://github.com/Yelp/paasta/blob/2c55307319826defc68d145abe3c6a45ed31f133/paasta_tools/utils.py#L637](url)

It could be a good idea to change the three defaults altogether.
I will go ahead and create the pr in paasta if you agree.

KaspariK (approved):

PR #1104, opened 2026-07-09
Support relaying a datetime to `start` commands from tronctl

jfongatyelp:
Today, tronctl start and tronctl backfill commands accept date-only parameters because they use the `parse_date` helper which only knows the `%Y-%m-%d` format. 

However, this meant t hat we would never be able to initiate a run with a custom start _time_, e.g. to backfill a sub-daily (eg hourly) job. 

The `start` support in the api, however, doesn't seem to reduce the `run_time` param to just a date: 
resource.py's [render_POST](https://github.com/Yelp/Tron/blob/master/tron/api/resource.py#L280) -> requestarg's [get_datetime](https://github.com/Yelp/Tron/blob/master/tron/api/requestargs.py#L55C48-L55C59) -> [DATE_FORMAT](https://github.com/Yelp/Tron/blob/master/tron/api/requestargs.py#L5) which includes time (`"%Y-%m-%d %H:%M:%S"`)

This PR expands the supported formats in parse_date to some common datetime formats.  I also took the opportunity to move it to our shared cmd_utils rather than leaving it just in tronctl itself, and added some dumb tests. 

This change won't help `tronctl backfill` generate more than daily schedules, but at least allows folks to work around that limitation and send their own datetime thru `tronctl start`.

### DIFF
--- a/bin/tronctl
+++ b/bin/tronctl
@@ -13,6 +13,7 @@ import sys
 from collections import defaultdict
 from collections.abc import Callable
 from collections.abc import Generator
+from functools import partial
 from typing import Any
 from urllib.parse import urljoin
 
@@ -127,7 +128,7 @@ def parse_cli():
     # start
     cmd_parsers["start"].add_argument(
         "--run-date",
-        type=parse_date,
+        type=partial(parse_date, allow_datetime=True),
         dest="run_date",
         help="What the run-date should be set to",
     )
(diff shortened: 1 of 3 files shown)


KaspariK:
I think one catch, that's probably not a big deal, is `backfill` will also use this expanded date parser, right? So we'll accept datetimes but the command will stay date-based (like, timestamps are stripped in dryrun output, but real runs still submit the full datetime to the API).

Not saying anyone would do this but this would generate 1 run:
`tronctl backfill job --start-date 2026-07-08T18:00 --end-date 2026-07-09`

And this would generate 2:
`tronctl backfill job --start-date 2026-07-08 --end-date 2026-07-09`

jfongatyelp:
Ah interesting that's a good call, I was just assuming because of the start
date range generator, that would strip whatever times the user passed in to
backfill.

Unfortunate! I can just add an allow_datetime toggle to `parse_date` and
use that for `start` but off for `backfill` as a quick fix, but if we get
around to adding more schedule support for backfill we can start using date
times again.

KaspariK (approved):

PR #1106, opened 2026-07-14
Persist JobRun max_runtime deadlines across Tron restarts

CharlesPeacock:
EDIT August 2026: Decided to initially set 7 days as max runtime until the necessary yelpsoa-configs for long running jobs are set

From Slack:

> The change became quite larger than expected as there are a lot of conditions to test for, here is a summary:
> 
> The tests check for expired restoration, remaining-duration restoration, not-started runs, manual runs and cancellation during finalize() / cleanup()
> 
> The PR itself ensures a timer is created for each job run regardless of starting method (manual, scheduled, etc) and that the timer is properly cleaned up and job run is stopped when it expires.
> Since we persist start_time in DynamoDB for a given job run, we check the remaining duration of the timer if it has a max_runtime set. We get this from max_runtime - elapsed_time  and then check if a run has gone over max_runtime to stop the run or assign a new timer for its remaining duration.

Changes to default max_runtime as preparation to Airflow migration. Airflow max job runtime is 12 hours.

~~I chose the 36h default after making some calls to the Tron API and measuring the max runtime duration I could find, which hovered over 18 hours.~~ This should be a safe starting point to avoid prematurely killing any ongoing or future Tron jobs.

Tests passed.

### DIFF
--- a/tron/config/config_parse.py
+++ b/tron/config/config_parse.py
@@ -774,7 +774,7 @@ class ValidateJob(Validator):
         "enabled": True,
         "queueing": True,
         "allow_overlap": False,
-        "max_runtime": None,
+        "max_runtime": datetime.timedelta(hours=36),
         "monitoring": {},
         "time_zone": None,
         "expected_runtime": datetime.timedelta(hours=24),
--- a/tests/config/config_parse_test.py
+++ b/tests/config/config_parse_test.py
@@ -158,7 +158,7 @@ def make_job(**kwargs):
     kwargs.setdefault("run_limit", 50)
     kwargs.setdefault("all_nodes", False)
     kwargs.setdefault("cleanup_action", make_cleanup_action())
-    kwargs.setdefault("max_runtime")
+    kwargs.setdefault("max_runtime", datetime.timedelta(hours=36))
     kwargs.setdefault("allow_overlap", False)
     kwargs.setdefault("time_zone", None)
     kwargs.setdefault("expected_runtime", datetime.timedelta(0, 3600))


tron/config/config_parse.py
```diff
@@ -774,7 +774,7 @@ class ValidateJob(Validator):
         "enabled": True,
         "queueing": True,
         "allow_overlap": False,
-        "max_runtime": None,
+        "max_runtime": datetime.timedelta(days=7),
```
KaspariK: Now that this branch/change has expanded to actually fixing max_runtime, lets hold off on doing the default in the same change. We can fix things in this PR and then have a small PR to set a default afterwards.

tron/core/job_scheduler.py
```diff
@@ -39,6 +39,16 @@ def restore_state(self, job_state_data, config_action_runner):
             master_action_runner=config_action_runner,
         )
 
+        # This recreates the max_runtime timer for restored runs as Tron does not persist this timer after restarting.
+        for job_run in job_runs:
+            if self.job.max_runtime and job_run.start_time and not job_run.end_time:
```
KaspariK: JobRun.start_time is derived from its ActionRuns, so it doesn't actually represent when the JobRun’s max_runtime timer was set. This creates two problems for us:
1. If every action is waiting on cross-job triggers, the timer is active but no ActionRun has started, so job_run.start_time is None.
2. When an action retries, its start_time is replaced with the new attempt’s start time, so the JobRun start time can move later.

After a restart, the first case can lose the timer entirely, while the second can incorrectly extend the deadline.

Instead, we probably want to make the max_runtime deadline an optional JobRun field. E.g.:
```
self.max_runtime_deadline: float | None = max_runtime_deadline
self.max_runtime_timer = None
```
max_runtime_deadline would be a Unix timestamp for the stop time. This would then need to be set through:
- JobRun.\_\_init\_\_
- state_data
- to_json()
- from_json()
- from_state()

Existing persisted JobRuns won't have the field, so restore should account for this (`.get()`).

tron/core/job_scheduler.py
```diff
@@ -170,7 +181,7 @@ def run_job(self, job_run, run_queued=False):
     def schedule_termination(self, job_run):
```
KaspariK: Once we are persisting the deadline (my comment on line 44), could we make schedule_termination() the single place that creates and persists deadlines and set/replaces the in-memory timers?

What I mean by this is something like:
```
def schedule_termination(self, job_run, reset_deadline=False):
    ...
```

This would work like:
1. If the job has no max_runtime, return without creating a timer.
2. Get the current Unix timestamp.
3. If reset_deadline=True, set the deadline to now + max_runtime. This is for an API retry that reactivates a completed JobRun and needs a fresh max_runtime window..
4. If the job has a configured max_runtime, but the JobRun does not yet have a persisted max_runtime_deadline (like an active jobrun restored from state written before this field existed) create an initial deadline:
    - If an ActionRun has a start_time, use it as an approximation of when the original timer began.
    - If no action has started because the JobRun is waiting on triggers, use the current time and grant it a fresh max_runtime window.
  5. Whenever the deadline is created or replaced, we want to emit `JobRun.NOTIFY_STATE_CHANGED` so the deadline is persisted.
6. Calculate the remaining duration as: `remaining = max(0, job_run.max_runtime_deadline - now)`
7. Cancel any existing timer before assigning the replacement:
    ```
    job_run.max_runtime_timer = reactor.callLater(
        remaining,
        job_run.stop,
    )
    ```
Restoration should then call schedule_termination(job_run) for JobRuns that have started but not completed. It should not require job_run.start_time, because trigger-waiting runs may not have one. A useful check would be that the run has no end_time and is neither scheduled nor queued.
  
We might also want a JobRun.cancel_max_runtime_timer() helper so we can use it from finalize(), cleanup(), and before replacing a timer. That keeps the reactor-timer lifecycle in one place.

tron/core/jobrun.py
```diff
@@ -364,12 +364,19 @@ def finalize(self) -> None:
             prom_metrics.tron_job_runs_completed_counter.labels(outcome="success").inc()
             log.info(f"{self} succeeded")
 
+        # Here we check if the JobRun has a max_runtime timer, if so, we cancel it via CallLater.cancel()
+        # Twisted docs mention that attempting to cancel an already cancelled timer will raise an AlreadyCancelledError, which is why we check if timer is active first.
+        if hasattr(self, "max_runtime_timer") and self.max_runtime_timer.active():
```
KaspariK: One edge case here is that an ActionRun belonging to a completed JobRun can still be retried through the API. ActionRun.retry() does this without going through JobScheduler, so the reactivated JobRun has no max_runtime timer.

A retry that reactivates a completed JobRun should receive a fresh max_runtime, while a retry within an already-active JobRun should keep its existing deadline (because max_runtime limits the whole JobRun, not each ActionRun attempt). When detecting whether the retry is reactivating a completed JobRun, note that action_runs.is_done excludes the cleanup action. You'll want to require that any cleanup action is also finished before granting a fresh deadline.

The fresh deadline should be persisted before ActionRun.retry() because retry can synchronously persist the active transition. If retry fails or raises, roll it back only if the JobRun remains complete. If the retry already became active, keep the fresh deadline.

KaspariK (commented):
Great work on this. I think you’re very close!

High-level we want to:
- Have max_runtime as a persisted JobRun field, because ActionRun start times are missing or mutable
- Centralize the management of the reactor timers
- Ensure API retries receive their correct timeouts

KaspariK (approved):

PR #1107, opened 2026-07-18
Create a tool for inspecting individual tron dynamodb items

CharlesPeacock:
This is a script made to check individual Tron DynamoDB items.

During [DAR-2843](https://jira.yelpcorp.com/browse/DAR-2843), diagnosing for corrupted state required ad-hoc scripting, this reusable tool aims to help in future investigations.

It can also be useful for anyone working on dynamodb_state_store.py now that these states are no longer stored in plain text.

It handles multi-partition reconstruction, decompression and deserialization so that you no longer need to do it manually.


tools/read_dynamodb_item.py
```diff
@@ -0,0 +1,202 @@
...
+        sys.exit(1)
+
+    if args.metadata:
+        print_metadata(partitions, key)
+        return
+
+    raw_json = reassemble_json(partitions)
+
+    if args.raw:
+        parsed = json.loads(raw_json)
```
KaspariK: `--raw` should not parse the decompressed payload. Using json.loads() will raise (or normalize, if there is somehow a duplicate key or something) malformed json before we inspect it. I think we'd want to write the raw_json directly

tools/read_dynamodb_item.py
```diff
@@ -0,0 +1,202 @@
...
+This is a tool that reads individual Tron DynamoDB items for debugging purposes
+
+It already handles multi-partition reassembly, gzip decompression, and deserialization so you don't have to do it ad-hoc during incidents.
+
+Example:
+
+Checking a job's state:
+
+AWS_PROFILE=devc python3 tools/read_dynamodb_item.py --table infrastage-tron-state --region us-west-1 --type job_state --name compute-i
+nfra-test-service.test_partitions
```
KaspariK: You probably didn't mean to add this line break

tools/read_dynamodb_item.py
```diff
@@ -0,0 +1,202 @@
...
+
+
+def build_key(state_type: str, name: str) -> str:
+    return f"{state_type} {name}"
+
+
+def get_client(table_name: str, region: str):
+    retry_config = Config(retries={"max_attempts": 5, "mode": "standard"})
+    client = boto3.client("dynamodb", region_name=region, config=retry_config)
+    return client, table_name
```
KaspariK: This returns table_name unchanged. Instead, we can accept only region and return the client.

tools/read_dynamodb_item.py
```diff
@@ -0,0 +1,202 @@
...
+    parser.add_argument(
+        "--raw",
+        action="store_true",
+        help="Dump decompressed JSON without deserializing through Tron's from_json",
+    )
+    parser.add_argument(
+        "--metadata",
+        action="store_true",
+        help="Show only partition metadata (count, sizes) without the payload",
+    )
```
KaspariK: This is a total nit, but `--raw` and `--metadata` can currently be provided together, with `--metadata` silently winning. If we wanted, we could use a mutually exclusive group to clarify this. Really not a big deal imo

KaspariK (approved):

PR #1108, opened 2026-07-22
U/krall/croniter

EvanKrall:
This _should_ be backwards-compatible, as we're already validating cron expressions in yelpsoa-configs using croniter. This PR provides a few benefits:

1. Gets rid of our home-grown cron parser, which has been the source of issues in the past.
2. Croniter supports things like "2nd monday" (`mon#2` or `1#2`), so once we ship this, we can migrate anything the groc syntax to croniter, then delete even more code.
3. Croniter supports Jenkins-style hashed expressions, where it can use a hash key (I'm using the job name here) to choose an arbitrary minute within the hour or hour within the day for e.g. daily/hourly jobs. This might help us improve node utilization by having fewer jobs all starting at the same time. (We currently support jitter, but we never use it - and a consistent time of day seems nicer for users than random-per-run offsets anyway.)


sidtuladhar (dismissed):

tron/scheduler.py
```diff
@@ -45,18 +47,11 @@ def scheduler_from_config(config, time_zone):
...
-            hours=config.hours,
-            monthdays=config.monthdays,
-            months=config.months,
-            weekdays=config.weekdays,
-            ordinals=config.ordinals,
-            seconds=[0],
-            name="cron",
-            original=config.original,
             jitter=config.jitter,
+            hash_id=job_name,
```
KaspariK, line 54: If we end up using the Jenkins-style hashed expressions in any capacity I think equivalent Airflow support should be a prerequisite (or...postrequisite?). 

They don't support the `H` syntax, so either we'd need to implement something like this that returns an Airflow CronTriggerTimetable or Airflow folks would need to implement something vaguely similar (coincidentally, there is a recent PR up to add JitteredCronTimetable, though that wouldn't be 1:1 in it's current approach).

tron/scheduler.py
```diff
@@ -85,6 +80,52 @@ def get_jitter_str(time_delta):
...
+    def next_run_time(self, start_time):
+        if not start_time:
+            start_time = timeutils.current_time(tz=self.time_zone)
+        elif self.time_zone:
+            if start_time.tzinfo is None or start_time.tzinfo.utcoffset(start_time) is None:
+                start_time = trontimespec.naive_as_timezone(start_time, self.time_zone)
+            else:
+                start_time = start_time.astimezone(self.time_zone)
+
+        it = croniter(self.cron_expression, start_time, hash_id=self.hash_id)
```
KaspariK, line 110: I think replacing `self.time_spec.get_match(start_time)` with a direct call to croniter changes our documented DST behaviour (for tz aware jobs that include a run in the transition hours). 

The GeneralScheduler calculates the next match as naive local time and then localizes it in trontimespec. Right now it catches ambiguous times and selects the first occurrence with `is_dst=True` ([sg](https://sourcegraph.yelpcorp.com/r/Yelp/Tron/-/blob/tron/utils/trontimespec.py?L71-74)). 

Passing a tz aware time directly to croniter lets it hit both instances, so `30 1 * * *` will schedule both 1:30 PDT and 1:30 PST. I think we want croniter to calculate against naive local time, then apply Tron’s existing timezone policy.

KaspariK: Oh, we have test_fall_back, but that just looks at generalscheduler

EvanKrall:
in light of #1072 being closed, I might want to add a test for Feb 30

tron/config/schedule_parse.py
```diff
@@ -289,18 +291,19 @@ def parse_groc_expression(config, config_context):
...
-        if crontab_kwargs["monthdays"] is not None and crontab_kwargs["weekdays"] is not None:
-            raise ValueError("cannot supply both monthdays and weekdays")
-        return ConfigCronScheduler(
-            original=config.value,
-            jitter=config.jitter,
-            **crontab_kwargs,
-        )
-    except ValueError as e:
+    expression = re.sub(r"\s*,\s*", ",", config.value.strip())
+    if not croniter.is_valid(expression, hash_id="validation_placeholder"):
```
KaspariK, line 295: Should this use the actual hash_id instead of a placeholder? If we have someone doing `0 0 H 2 *` we could end up with 30/31 since Croniter hashes H across the full 1–31 range. A less likely but funnier risk would be `0 0 31 H *`

tests/scheduler_test.py
```diff
@@ -48,6 +50,15 @@ def test_daily_scheduler(self):
...
 
+class TestCronSchedulerImpossibleDate(TestCase):
+    def test_impossible_date_raises_instead_of_infinite_loop(self):
+        """Feb 30 can never match. Croniter must raise rather than loop forever."""
+        sched = scheduler.CronScheduler(cron_expression="0 0 30 2 *")
+        start_time = datetime.datetime(2024, 1, 1)
+        with pytest.raises(CroniterBadDateError):
+            sched.next_run_time(start_time)
+
+
```
KaspariK, line 61: Thoughts on repurposing some of these GeneralScheduler tests for CronScheduler? 
I think some of these should be carried forward. E.g.
- `test_fall_back` should look at CronScheduler("30 1 * * *")
- `test_spring_forward` should look at CronScheduler("30 2 * * *") to ensure we shift nonexistent times forward
- `test_handles_unsetting_the_time_zone` and `test_handles_changing_the_time_zone` both test behaviour we should maintain

KaspariK: Jeez, `test_handles_unsetting_the_time_zone` is just incorrect in its current form. I like the spirit of it, but imo it shouldn't start with UTC and it shouldn't assert hour == 0.

KaspariK: JEEZ, `test_handles_changing_the_time_zone` is also funky. I don't think this should look at datetime.now and I very much disagree with asserting on just the hour.

tron/config/schedule_parse.py
EvanKrall: Interesting point.

PR #1109, opened 2026-07-23
First run of a new job is never persisted to dynamodb

CharlesPeacock:
This PR aims to fix a bug where Tron did not to attach a watcher to the first run of a new job. Interestingly enough, this bug has existed for several years now; since 85% of Tron jobs run every 20 minutes or so, this went unnoticed for a while since reaching the run_limit for a given job removes the old runs from the tail.

This will help heavier, less frequent job runs in ensuring data integrity and consistent observability across Tron. A unit test has been added to assert old behavior is no longer present.

All tests pass, keeping as draft PR for the time being until I verify more job runs on my local env.

### DIFF
--- a/tests/core/job_collection_test.py
+++ b/tests/core/job_collection_test.py
@@ -31,6 +31,16 @@ class TestJobCollection(TestCase):
             job_scheduler.schedule.assert_called_with()
             job_scheduler.get_job.assert_called_with()
 
+    def test_map_to_job_and_schedule(self):
+        autospec_method(self.collection.jobs.filter_by_name)
+        autospec_method(self.collection.add)
+        factory = mock.create_autospec(JobSchedulerFactory)
+        job_configs = {"a": mock.Mock(), "b": mock.Mock()}
+        generator = self.collection.update_from_config(job_configs, factory, True)
+        next(generator)
+        job_schedulers = [call[1][0] for call in self.collection.add.mock_calls[::2]]
+        job_schedulers[0].schedule.assert_not_called()
+
     def test_update_from_config_reconfigure_one_namespace(self):
         autospec_method(self.collection.jobs.filter_by_name)
         autospec_method(self.collection.add)
--- a/tron/core/job_collection.py
+++ b/tron/core/job_collection.py
@@ -30,9 +30,9 @@ class JobCollection:
 
         def map_to_job_and_schedule(job_schedulers):
             for job_scheduler in job_schedulers:
+                yield job_scheduler.get_job()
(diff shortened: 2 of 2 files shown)


CharlesPeacock:
I attempted to replicate the error after the fix, it now works as intended:



Note: I was initially having trouble with SIGHUP until I discovered that Tron caches configs in memory until `write_config()` is called via the API, which triggers the reconfigure, I now realize this is the process triggered by the `yelpsoa-configs` in the production environment.

**Setup**
I first set up the environment by creating a dynamodb table in dev env called `tmp-tron-state` in `us-west-1`

Tron running locally and k8s pointing to infrastage.
(This is mostly the setup contained in Eman's Elsabban [How to run Tron locally](https://yelp.atlassian.net/wiki/spaces/COMPUTE/pages/77923283/How+to+run+Tron+locally) guide)

### **Testing if new jobs via reconfigure persist run 0:**

Started Tron with only `testjob0` in `MASTER.yaml`
Waited for `Tron built the schedulers in logs`
Added `persist_bug_test job` via 
`tronfig --server http://localhost:8089 -n MASTER dev/config/MASTER.yaml`

Logs confirmed desired behavior:

```
Watching new run JobRun:MASTER.persist_bug_test.0
Buffering state save for: job_run_state MASTER.persist_bug_test.0
State saved using ...DynamoDBStateStore...
```

I then verified this in dynamodb:

```
aws dynamodb get-item --table-name tmp-tron-state --region us-west-1 \
  --key '{"key": {"S": "job_run_state MASTER.persist_bug_test.0"}, "index": {"N": "0"}}'
```
Item was present inside dynamodb in full serialized state

Without this fix, `yield` came in after `schedule()` in `map_to_job_and_schedule`, meaning the state watcher was not attached when run 0 was created

### Verifying existing jobs reconfigure correctly**

I changed the `testjob0` command string, which I pushed via **tronfig** API

The logs showed that `Job:MASTER.testjob0 reconfigured`, which is what we want

### Checking no duplicates were created

After these tests I queried `job_state MASTER.persist_bug_test` from dynamodb, after deserialization:
`run_nums: [2, 1, 0]`
This confirms that the `yield` before `schedule()` ordering is not creating double scheduling for jobs.

### Making sure restarting restores run 0

I killed tron and restarted it, the logs showed the following information:

```Restoring 2 jobs
Watching new run JobRun:MASTER.persist_bug_test.0
JobScheduler(Job:MASTER.persist_bug_test) restored
```

This confirms run0 is being retrieved successfully

KaspariK (approved):

PR #1111, opened 2026-07-28
Improve Tron restart times by adding parallelization to job_collection.restore_state()

CharlesPeacock:
Follow-up to the series of Tron restart improvements.

As of recently, Tron has been seeing restart times of over 6 minutes in pnw-prod. Optimizing these restart times allows for less downtime and more opportunities to experiment with new changes.

This change adds a `ThreadPoolExecutor` to parallelize `job_collection.restore_state()`. By setting `max_workers` to 5 we can see a significant improvement in state restoration times:

Original performance (tested with copy of tron-pnw-devc dynamodb table):

```
Run: linear run (Tue Jul 28 10:32:01 AM PDT 2026)
Total startup: 93.59s
Schedulers built: 3.786149263381958s
DynamoDB retrieval: 10:27:25 -> 10:27:55 (30.000s)
Apply state: 10:27:55 -> 10:28:53 (58.000s)
Unprocessed key retries: 5
Log: /tmp/output_log_linear_attempt1.log
Restore state: 57.425s <----- 
```

Performance after parallelization (same environment as original test, 5 workers):

```
Run: parallel run (Tue Jul 28 10:19:16 AM PDT 2026)
Total startup: 66.10s
Schedulers built: 3.781592607498169s
DynamoDB retrieval: 10:14:21 -> 10:14:52 (31.000s)
Apply state: 10:14:52 -> 10:15:22 (30.000s)
Unprocessed key retries: 7
Log: /tmp/output_log_parallel_attempt2.log
Restore state: 29.658s <-----
```

Tests passed

### DIFF
--- a/tron/core/job_collection.py
+++ b/tron/core/job_collection.py
@@ -1,3 +1,4 @@
+import concurrent.futures
 import logging
 
 from tron.core.job import Job
@@ -75,9 +76,17 @@ class JobCollection:
         state for each run. As we load the state, we will also schedule the next
         runs for each job
         """
-        for name, state in job_state_data.items():
-            self.jobs[name].restore_state(state, config_action_runner)
-        log.info(f"Loaded state for {len(job_state_data)} jobs")
+
+        with concurrent.futures.ThreadPoolExecutor(max_workers=5) as executor:
+            results = {
+                executor.submit(self.jobs[name].restore_state, state, config_action_runner): name
+                for name, state in job_state_data.items()
+            }
+            for result in concurrent.futures.as_completed(results):
+                try:
+                    result.result()
+                except Exception:
+                    log.exception(f"Unable to restore state for {results[result]} - exiting")
 
     def get_by_name(self, name):
         return self.jobs.get(name)


tron/core/job_collection.py
```diff
@@ -75,9 +76,17 @@ def restore_state(self, job_state_data, config_action_runner):
...
+        with concurrent.futures.ThreadPoolExecutor(max_workers=5) as executor:
+            results = {
+                executor.submit(self.jobs[name].restore_state, state, config_action_runner): name
+                for name, state in job_state_data.items()
+            }
+            for result in concurrent.futures.as_completed(results):
+                try:
+                    result.result()
+                except Exception:
+                    log.exception(f"Unable to restore state for {results[result]} - exiting")
```
KaspariK, line 89: Adding try/except Exception around the future.result without re-raising introduces an issue where we suppress restoration failures. If a job’s deserialization raises, its worker stops, but restore_state() eventually returns normally. Tron will eventually see this as a complete restoration and try to queue a run for that unrestored job, which can cause duplicate runs. Example:
1. DynamoDB records run 123 of job foo as active.
2. The Kubernetes pod for run 123 continues to run.
3. Tron does not restore run 123, or Tron gives the run an UNKNOWN state.
4. The exception handler suppresses the restore error.
5. Startup continues and calls run_queue_schedule().
6. Tron does not treat an UNKNOWN run as active.
7. Tron schedules a new run. If no runs were restored, the new run number is 0.
8. The new run can start while the pod for run 123 is still running.

tron/core/job_collection.py
```diff
@@ -75,9 +76,17 @@ def restore_state(self, job_state_data, config_action_runner):
...
         runs for each job
         """
-        for name, state in job_state_data.items():
-            self.jobs[name].restore_state(state, config_action_runner)
-        log.info(f"Loaded state for {len(job_state_data)} jobs")
+
+        with concurrent.futures.ThreadPoolExecutor(max_workers=5) as executor:
+            results = {
+                executor.submit(self.jobs[name].restore_state, state, config_action_runner): name
+                for name, state in job_state_data.items()
```
KaspariK, line 83: PersistentStateManager (StateSaveBuffer) uses one shared dictionary. Parallel restore causes multiple workers to save through that shared PersistentStateManager and we can end up with:
- Worker A copies the old value for job foo
- Worker B writes a newer value for foo
- Worker A clears the shared buffer

The newer value will be removed without being saved.

I think the smallest fix is to synchronize the buffer operation in PersistentStateManager. Job reconstruction and recovery can still be parallel but access to shared persistence state stays serialized.

tron/core/job_collection.py
CharlesPeacock: Thank you for the feedback. It taught me more about how Tron works under the hood! Thanks again for the explanation in detail :)

CharlesPeacock: Adding a unit test to check this behavior could be a good idea...

PR #1112, opened 2026-07-29
Remove pickle writes to DynamoDB

CharlesPeacock:
This change removes pickle writes to dynamodb. Since we no longer read from pickles, this was the next logical step on [In a Pickle No More: A Plan to Escape Our Briny Dill-emma](https://yelp.atlassian.net/wiki/spaces/COMPUTE/pages/77923163/In+a+Pickle+No+More+A+Plan+to+Escape+Our+Briny+Dill-emma). 

Next steps include applying these changes to the `_delete_item()` and `_get_num_of_partitions()` functions, right now this should not be done until we run `compress_json.py` to keep ourselves from leaving orphaned entries in dynamodb.

After seeing the behavior of these changes, modifying dynamodb OBJECT_SIZE from 150kb to 350-380kb would be a good next step.

Tested against a copy of the tron-pnw-devc dynamodb table. The save loop wrote items without `val/num_partitions` and no serialization errors occurred. I also changed two tests to reflect changes. make test passed.

Example of dynamodb item write:

```
{
  "key": {
    "S": "job_run_state ee_metrics_event_batches.puppet_role_ubuntu_release_jammy.1295"
  },
  "index": {
    "N": "0"
  },
  "json_val": {
    "B": "H.....sK2B0AAA==" #Trimmed to save space
  },
  "num_json_val_partitions": {
    "N": "1"
  }
}
```

### DIFF
--- a/tests/serialize/runstate/dynamodb_state_store_test.py
+++ b/tests/serialize/runstate/dynamodb_state_store_test.py
@@ -228,9 +228,9 @@ class TestDynamoDBStateStore:
         assert store.save_errors == 0
 
         for key in keys:
+            # Now that we are not writing pickles, we only perform assertions on num_json_val_partitions. I keep num_partitions for visibility of changes.
             num_partitions, num_json_val_partitions = store._get_num_of_partitions(key)
             assert num_json_val_partitions > 1
-            assert num_partitions > 1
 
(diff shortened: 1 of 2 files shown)


KaspariK (approved):
Nice work!

PR #1113, opened 2026-08-05
Avoid read-only api endpoints from blocking reactor

CharlesPeacock:
Fix for [TRON-2678](https://jira.yelpcorp.com/browse/TRON-2678)

This PR targets two Tron read-only endpoints: `GET /api/metrics` and `GET /api/events`.
These two endpoints currently run `exclusive`, meaning that they drain all 10 semaphore slots and block the reactor until they finish executing. Since these endpoints are called every minute or so, this is constantly detrimental to Tron's performance.

This PR changes these two endpoints from `exclusive` to `bounded` while adding a mutex lock on `all_metrics` inside of `metrics.py` to protect it from concurrent inserts during iterations.

A shallow copy was added to `EventsController.info()` to avoid racing with concurrent publishes and discards.

I tested the changes on my local environment with a copy of the `tron-pnw-devc` DynamoDB table.
I created a script that made requests to the Tron API in order to confirm stability, it consisted of stress tests, concurrency tests, lock contention and continuous read/writes. ~~I did not find issues while testing.~~

All tests pass.

EDIT: After making a last change to the PR, I discovered an existing race condition affecting `/api/jobs` as I was mysteriously getting 500 errors during concurrency tests. Back in 2018,` /api/jobs` became `@bounded` from its previous `@exclusive` state, but reads execute on worker threads while the shared deque `JobRunCollection.runs() `can be mutated by the reactor, resulting in replicable `RuntimeError: deque mutated during iteration.` exceptions.
I have made a new PR to address this issue and 500 errors are no longer an issue.


tron/metrics.py
```diff
@@ -86,8 +96,10 @@ def view_histogram(histogram):
 
 
 def view_all_metrics():
+    with all_metrics_lock:
+        all_metrics_copy = all_metrics.copy()
```
KaspariK, line 100: Nice change moving this to bounded!

One thread safety issue is that dict.copy is shallow. all_metrics_copy is a new dictionary, so adding a metric to all_metrics won’t change the structure of all_metrics_copy. That makes it safe to iterate over while another thread adds entries. However, the metric values themselves are shared. E.g., all_metrics_copy[key] and all_metrics[key] refer to the same metric object. The reactor can call all_metrics[key].mark() while the worker reads that object through all_metrics_copy[key].

Could we keep all_metrics_lock held while building all_data instead? The lock will be released before respond() serializes the result and the endpoint will still run in the worker. Hopefully that keeps the majority of the benefit you've introduced. I.e.:
```
with all_metrics_lock:
    all_data = {metric_type: [] for metric_type in metrics_to_viewers}

    for (metric_type, name, dims), metric in all_metrics.items():
        ...
return all_data
```

PR #1116, opened 2026-08-21
Tweak max_workers and max_pool_connections - restart improvements #3

CharlesPeacock:
EDIT: This PR now adds the possibility of setting max_workers and max_pool_connections via MASTER.yaml.
This should allow us to easily manage and experiment with these settings across different environments.

All tests pass and retrocompatibility is maintained.

follow up to TRON-2651.

This PR adds tweaks to `max_workers` and `max_pool_connections`. I measured startup time before and after the changes and measured a marginal improvement of approximately 5%.

After analyzing our `tron-pnw-prod` DynamoDB RCU graphs, I settled on 5 outer and 10 inner workers. Testing in dev showed no issues or state loss.

Our theoretical peak concurrency increased from 15 (3 × 5) to 50 (5 × 10). `max_pool_connections` increased from the default of 10 to 30. This is not a hard concurrency limit, it allows connections for up to 60% of the theoretical peak to be retained and reused, reducing connection churn.


KaspariK (dismissed):

tron/config/config_parse.py
```diff
@@ -883,15 +883,29 @@ class ValidateStatePersistence(Validator):
...
             raise ConfigError("%s buffer_size must be >= 1." % path)
 
+        if max_pool_connections < 10:
+            path = config_context.path
+            raise ConfigError("%s max_pool_connections must be 10 or higher" % path)
+
+        if batch_get_workers < 3 or restore_workers < 3:
+            path = config_context.path
+            raise ConfigError("%s restore_workers and batch_get_workers must be 3 or higher" % path)
+
```
KaspariK, line 908: I'm curious about this. Did you hit certain things at these values? Or are these set somewhat arbitrarily?

CharlesPeacock: For max_pool_connections, boto3 has a default of 10, which is what we have always been using.
I decided to reject a config lower than that as it could cause unexpected behavior when running Tron in Prod.

For batch_get_workers and restore_workers, I decided to stay with 3 as that is the lowest stable number we have used in the past, though you could say it is somewhat arbitrary.

KaspariK: Cool, makes sense as a supported "floor" rather than a real limitation. I’m okay retaining those

KaspariK (approved):

tron/config/config_parse.py
KaspariK: I don't think we'll ever miss not having maximums either. If someone wants to set 1000000 workers they can try :p

PR #1117, opened 2026-08-21
U/kkasp/tron 2686 tronweb actionrun durations

KaspariK:
Just when I thought I was out, CoffeeScript pulls me back in. This adds an Attempts table to the ActionRun page showing each attempt’s:
- Start and end time
- Duration
- Exit code and human-readable result
- Current running state

While I was here, I opted to make Tron-specific exit codes display their descriptions, e.g. `-11 Kubernetes task failed due to the autoscaler scaling down a node`. This was probably on my mt. rushmore of tronweb peeves.

Funny bugs that I fixed along the way:
- Retrying a completed ActionRun updated its start time but kept the previous end time. This produced a negative duration until the run finished.
- A running retry retained the previous attempt’s exit status, causing the ActionRun summary to show a stale result while the new attempt was active.

**What does it look like?**: [fluffy](https://fluffy.yelpcorp.com/i/bdGTGBk1HC7VTGJ0P6Q5B93JFRZQBbZC.html)


tron/api/adapter.py
```diff
@@ -147,6 +151,35 @@ def get_raw_command(self):
     def get_command(self):
         return self._obj.rendered_command
 
+    def get_exit_reason(self):
+        return exitcode.get_exit_reason(self._obj.exit_status)
+
+    def get_attempts(self):
```
jfongatyelp, line 169: Since we don't seem to use `attempts` except in the actionrun view, should we `@toggle_flag` like we do with other action-run-detail-only methods? like get_requirements i guess?

jfongatyelp (dismissed):
> What does it look like?: [fluffy](https://fluffy.yelpcorp.com/i/bdGTGBk1HC7VTGJ0P6Q5B93JFRZQBbZC.html) 
 
This one shows the new exit reasons (<3 , til that was in tronview but not tronweb), but i am kinda curious what the new attempts table looks like too!

KaspariK:
> > What does it look like?: [fluffy](https://fluffy.yelpcorp.com/i/bdGTGBk1HC7VTGJ0P6Q5B93JFRZQBbZC.html)
> 
> This one shows the new exit reasons (<3 , til that was in tronview but not tronweb), but i am kinda curious what the new attempts table looks like too!

@jfongatyelp you might need to scroll down on the fluffy page, but in case that's not working this is the table while [running](https://fluffy.yelpcorp.com/i/3cQ1VZj1xJxcrDD0NWg6hc0CjVt4Sl5N.png) and [after completion](https://fluffy.yelpcorp.com/i/CRQvwtZCD7GwZcZkm7z2wnDnbB69KHXg.png).

tron/api/adapter.py
KaspariK: ++, added an `include_attempts` toggle

jfongatyelp:
> > > What does it look like?: [fluffy](https://fluffy.yelpcorp.com/i/bdGTGBk1HC7VTGJ0P6Q5B93JFRZQBbZC.html)
> > 
> > 
> > This one shows the new exit reasons (<3 , til that was in tronview but not tronweb), but i am kinda curious what the new attempts table looks like too!
> 
> @jfongatyelp you might need to scroll down on the fluffy page, but in case that's not working this is the table while [running](https://fluffy.yelpcorp.com/i/3cQ1VZj1xJxcrDD0NWg6hc0CjVt4Sl5N.png) and [after completion](https://fluffy.yelpcorp.com/i/CRQvwtZCD7GwZcZkm7z2wnDnbB69KHXg.png).

oops, sorry bout that, had clicked on the first image and didnt realize it brought me out a page of multiple!

tron/api/resource.py
```diff
@@ -169,6 +169,7 @@ def render_GET(self, request):
             include_stdout=requestargs.get_bool(request, "include_stdout"),
             include_stderr=requestargs.get_bool(request, "include_stderr"),
             include_meta=requestargs.get_bool(request, "include_meta"),
+            include_attempts=True,
```
jfongatyelp, line 172: since we're never getting this from the request and the default is true, is it doing anything?

KaspariK: It defaults to False in the Adapter (on L140). 

A JobRun GET that includes its ActionRuns (i.e. `GET /jobs/<job>/<run>?include_action_runs=1`) uses a path that constructs each ActionRunAdapter without include_attempts, so it should use the default.

My understanding is that only `GET /jobs/<job>/<run>/<action>` goes through ActionRunResource.render_GET().

KaspariK: I should probably have followed the existing pattern and used a `&include_attempts=1` or something tho

tron/api/adapter.py
```diff
@@ -133,20 +137,52 @@ def __init__(
...
+    def get_exit_reason(self):
+        return exitcode.get_exit_reason(self._obj.exit_status)
+
+    @toggle_flag("include_attempts")
+    def get_attempts(self):
+        attempts = []
+        for attempt_number, attempt in enumerate(self._obj.attempts, start=1):
+            start_time = attempt.start_time if isinstance(attempt.start_time, datetime.datetime) else None
+            end_time = attempt.end_time if isinstance(attempt.end_time, datetime.datetime) else None
+            is_running = start_time is not None and attempt.end_time is None and attempt.exit_status is None
```
KaspariK, line 174: omg we don't set end time on cancellations what in tarnation

jfongatyelp (dismissed):

KaspariK:
I'll hold off on this one until #1118

jfongatyelp (approved):

PR #1119, opened 2026-09-08
Enabling max_runtime in MASTER.yaml

CharlesPeacock:
This change enables setting the max_runtime feature via MASTER.yaml. Initial testing in dev will show if further tweaks are necessary. Prod rollout will be done after notifying engineering and ensuring the feature is stable.

---
UPDATE:
In a similar fashion to time_zone, max_runtime can now be set up via MASTER.yaml configs, I added guards on invalid per-job max_runtime values that fallback to the default we set (7 days, to be set-up in yelpsoa-configs). I added and updated tests to verify behavior.

More context can be found here:
- max_runtime feature PR: https://github.com/Yelp/Tron/pull/1106
- Slack discussions: https://yelp.slack.com/archives/C02NACJMW77/p1787222649286709


tron/config/config_parse.py
```diff
@@ -771,7 +771,7 @@ class ValidateJob(Validator):
         "enabled": True,
         "queueing": True,
         "allow_overlap": False,
-        "max_runtime": None,
+        "max_runtime": datetime.timedelta(days=7),
```
KaspariK: imo this would be better off set in schema.py. I say that because there is a world where we want a different default in different envs, or want to adjust it without deploying Tron. Job level max runtime could still override it. This should jsut require the config schema/parser and threading the value through mcp to job_scheduler (kind of like time_zone)

tron/config/config_utils.py
```diff
@@ -159,6 +159,8 @@ def valid_time(value, config_context):
 
 
 def valid_time_delta(value, config_context):
+    if value is None:
+        return None
```
KaspariK: We have to be a little careful here because this is a shared validator. None would not be valid on expected_runtime, for example.

CharlesPeacock: Thank you, I added a helper that copies valid_time_delta() with the None conditional, only used for max_runtime validation.

KaspariK (approved):

PR #1120, opened 2026-09-23
Deliver eternal rest to most deprecated mesos code and tests

CharlesPeacock:
Removing old mesos code as `pymesos` is no longer supported in Python versions 3.10+. This originated from `pymesos` requiring the now incompatible `http-parser` library, making it necessary to remove from Tron.

Parts removed:

- Mesos Runtime, tests and documentation
- Mesos only dependencies (`http-parser`, `pymesos`, `addict`, `task-processing[mesos-executor]`)
- Mesos implementation in Tron's configuration, execution, persistence and recovery

Tested in infrastage for 15 hours. No visible issues arised.

Changes applied at the 18:00 spike:
image

DynamoDB showed no errors and regular read/write performance:

image

Tests pass. 
For CIB: [Slack Thread](https://yelp.slack.com/archives/C011T3ECGHX/p1790115712335379)


tron/config/config_parse.py
```diff
@@ -60,6 +58,18 @@
...
+# to skip the deprecated pymesos dependency.
+CONSTRAINT_OPERATORS = (
+    "EQUALS",
+    "==",
+    "NOTEQUALS",
+    "!=",
+    "LIKE",
+    "UNLIKE",
+)
+
```
nemacysts: fwiw: nothing should be using these anymore

nemacysts: (constraints were a mesos thing)

requirements-minimal.txt
```diff
@@ -1,11 +1,9 @@
-addict  # not sure why check-requirements is not picking this up from task_processing[mesos_executor]
 argcomplete
 boto3
 bsddb3
 cryptography
 dataclasses
```
nemacysts: huh, this should be built-in already

nemacysts (commented):
@CharlesPeacock in case you haven't done this already - probably worth chatting with @KaspariK and verifying that we can still load all the data from dynamodb in the other clusters


we should be able to now that we just read/write json - but i think this will be the first time we use this new capability in anger :p

tron/core/actionrun.py
```diff
@@ -196,12 +188,6 @@ def to_json(state_data: dict) -> str:
                     "end_time": state_data["end_time"].isoformat() if state_data["end_time"] else None,
                     "rendered_command": state_data["rendered_command"],
                     "exit_status": state_data["exit_status"],
-                    # NOTE: mesos_task_id can be deleted once we delete all Mesos
```
KaspariK: Mentioned in Slack, but we should keep writing `"mesos_task_id": null` and remove it in a later change. Removing it now will make things non-revertable (new runs are written in a way old code can't handle)

